PR #1, opened 2023-11-10
.github: enable ARM builds

ondrejbudai:


teg (approved):
Thank you!

mvo5 (approved):
This looks very nice! Given that it already has two +1s, can we just merge it :) ? Or is something missing?

achilleas-k:
We have RHTAP builds now so I think it's not needed?
Or do we want it as a secondary for some reason?

lmilbaum:
I've just configured RHTAP pipeline to push the image to quay.io/centos-bootc namespace.
In case you find it valuable, the following GHA workflow push a multi-arch image - https://github.com/CentOS/centos-bootc-infra/blob/main/.github/workflows/build-and-push-image.yml

teg:
This takes 2h to build, should we drop it in favour of RHTAP?

PR #5, opened 2023-11-13
WIP: Merge osbuildbootc into here, use osbuild/images PR

cgwalters:
- Take the code from https://github.com/achilleas-k/images/tree/bifrost-image/cmd/osbuild-deploy-container
  and merge it into this repository, using the code from https://github.com/osbuild/images/pull/243
  as a `replace`
- Also merge in osbuildbootc

OK this was all pretty straightforward until I hit:

```
$ make
(cd cmd && go build -mod vendor -o ../bin/osbuildbootc)
(top=$(pwd); cd cmd/osbuild-deploy-container && go build -mod vendor -o ${top}/bin/osbuild-deploy-container)
package github.com/cgwalters/osbuildbootc/cmd/osbuild-deploy-container
        image.go:7:2: use of internal package github.com/osbuild/images/internal/common not allowed
package github.com/cgwalters/osbuildbootc/cmd/osbuild-deploy-container
        main.go:12:2: use of internal package github.com/osbuild/images/internal/dnfjson not allowed
package github.com/cgwalters/osbuildbootc/cmd/osbuild-deploy-container
        image.go:8:2: use of internal package github.com/osbuild/images/internal/users not allowed
package github.com/cgwalters/osbuildbootc/cmd/osbuild-deploy-container
        image.go:9:2: use of internal package github.com/osbuild/images/internal/workload not allowed
make: *** [Makefile:20: bin] Error 1
```

So I think we're going to need to figure that stuff out right?


achilleas-k:
> package github.com/cgwalters/osbuildbootc/cmd/osbuild-deploy-container
>         image.go:7:2: use of internal package github.com/osbuild/images/internal/common not allowed
> package github.com/cgwalters/osbuildbootc/cmd/osbuild-deploy-container
>         main.go:12:2: use of internal package github.com/osbuild/images/internal/dnfjson not allowed
> package github.com/cgwalters/osbuildbootc/cmd/osbuild-deploy-container
>         image.go:8:2: use of internal package github.com/osbuild/images/internal/users not allowed
> package github.com/cgwalters/osbuildbootc/cmd/osbuild-deploy-container
>         image.go:9:2: use of internal package github.com/osbuild/images/internal/workload not allowed

- `dnfjson` should be public and we're going to move it into `pkg` as part of [this PR](https://github.com/osbuild/images/pull/246) so it can also be used by osbuild-composer.
- The things from `common` are trivial: one is just a `ToPtr()` function for making pointers from literals and the other is basically the equivalent of `uname -m` for getting the machine architecture.
- `workload` should also be public.  It's meant for configuring workload-specific customizations and would be useful to have in general (but maybe not for this.  It's not really doing anything in this code (it's empty), I just needed a workload, any workload, to make the pipelines happy.
- `users` I'll have to think about.  We can drop it for now if we don't care about creating users while creating the image and think about it.  I think it will end up being part of the public API though.

ondrejbudai:
Do we still need this PR? The Go code is already here. I'm not sure about the VM-related code - it adds more user-facing surface and complexity. Do we actually have any users who need this?

PR #8, opened 2023-11-23
Make the container work on an SELinux-enforcing systems

### LINKED ISSUE #6: `setfiles: Could not set context` on SELinux-enforcing systems
Tried with `quay.io/centos-boot/fedora-tier-1:eln` and `quay.io/centos-boot/centos-tier-1:stream9`.

```
sudo podman run --rm -it --privileged -v $(pwd)/images:/output ghcr.io/osbuild/osbuild-deploy-container -imageref quay.io/centos-boot/centos-tier-1:stream9
[...]
org.osbuild.selinux: c73ddc1b46d5d88c144b1b185cf2559477ea8bcb72f87365ce5fbc02d4625ef3 {
  "file_contexts": "etc/selinux/targeted/contexts/files/file_contexts",
  "labels": {
    "/usr/bin/cp": "system_u:object_r:install_exec_t:s0"
  }
}
/usr/lib/tmpfiles.d/journal-nocow.conf:26: Failed to resolve specifier: uninitialized /etc/ detected, skipping.
All rules containing unresolvable specifiers will be skipped.
setfiles: Could not set context for /run/osbuild/tree/usr/lib/systemd/system-generators/systemd-fstab-generator:  Invalid argument
setfiles: Could not set context for /run/osbuild/tree/usr/lib/systemd/system-generators/systemd-rc-local-generator:  Invalid argument
setfiles: Could not set context for /run/osbuild/tree/usr/lib/systemd/system-generators/systemd-sysv-generator:  Invalid argument
  File "/run/osbuild/bin/org.osbuild.selinux", line 75, in <module>
    r = main(args["tree"], args["options"])
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/run/osbuild/bin/org.osbuild.selinux", line 62, in main
    subprocess.run(["setfiles", "-F", "-r", f"{tree}", f"{file_contexts}", f"{tree}"], check=True)
subprocess.CalledProcessError: Command '['setfiles', '-F', '-r', '/run/osbuild/tree', '/run/osbuild/tree/etc/selinux/targeted/contexts/files/file_contexts', '/run/osbuild/tree']' returned non-zero exit status 255.

⏱  Duration: 4s

Failed
running osbuild failed: exit status 1
```

ondrejbudai:
When building an image, we need to make sure that the target system is
correctly labeled. This becomes challenging if the target system contains
labels that are unknown to the host because the process setting the label
needs to have CAP_MAC_ADMIN if the host is SELinux-enforcing.

CAP_MAC_ADMIN isn't a common capability on a SELinux-enforcing system.
Even unconfined_t doesn't have it (same for spc_t - label used by
--privileged containers). Thus, we need to ensure that we transition to
a domain that actually has it.

This commit relabels osbuild as install_t, a domain that has CAP_MAP_ADMIN.
A bit of mount-dancing is needed in order to achieve that, see prepare.sh.

I decided to make prepare.sh a separate script. This is useful for debugging:

host # podman run -it \
  --privileged \
  --entrypoint bash \
  localhost/osbuild-deploy-container
container # ./prepare.sh

This way, you get the same environment as if you run the container the
default way.

See https://github.com/osbuild/osbuild-deploy-container/issues/6#issuecomment-1822809036
and links in this comment for further information.

### Open questions:
- Does this work on:
  - [x] EL8?
  - [x] EL9?
  - [x] Fedora 38?
  - [x] Fedora 39?
  - [x] non-SELinux system (Ubuntu/Arch)?
- Can we indeed ignore cleaning up the new mounts?
  - [x] It looks like it
- Can we reasonably fix the denial that happens in the `ostree.container.deploy` stage?
  - [x] We think so, but as this doesn't have any effect on the resulting artifact, we can handle this in a follow-up.

Fixes #6

### DIFF
--- a/prepare.sh
+++ b/prepare.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+set -euo pipefail
+
+# Create a new tmpfs. This solves two issues for us:
+# - / is mounted as nosuid, this prevents SELinux to transition to `install_t` because domain transitions are
+#   disallowed if they give more caps to the process and the target executable is on `nosuid` filesystem
+# - / can be mounted as OverlayFS that doesn't support overlaying SELinux labels. Thus, we need to ensure that
+#  the relabeling happens on a mountpoint that's definitely not an OverlayFS.
+TMP=/run/suidtmp
+mkdir -p "${TMP}"
+
+# The container is mounted as MS_SHARED, this mount as well. Thus, we don't need to care about cleanup, when the
+# container dies, it will take this mount with itself.
+mount -t tmpfs tmpfs "${TMP}"
+
+# Copy osbuild to the new mountpoint.
+cp /usr/bin/osbuild "${TMP}/osbuild"
+
+# Label it as `install_exec_t`. We need this in order to get `install_t` that has `CAP_MAC_ADMIN` for creating SELinux
+# labels unknown to the host.
+#
+# Note that the transition to `install_t` must happen at this point. Osbuild stages run in `bwrap` that creates
+# a nosuid, no_new_privs environment. In such an environment, we cannot transition from `unconfined_t` to `install_t`,
+# because we would get more privileges.
+chcon system_u:object_r:install_exec_t:s0 "${TMP}/osbuild"
+
+# "Copy" back the relabeled osbuild to its right place. We obviously cannot copy it, so let's bind-mount it instead.
+# Once again, we don't care about clean-up, this is MS_SHARED.
+mount -o bind "${TMP}/osbuild" /usr/bin/osbuild
--- a/Containerfile
+++ b/Containerfile
@@ -6,7 +6,7 @@ RUN ./build.sh
 FROM registry.fedoraproject.org/fedora:39
 RUN dnf install -y osbuild osbuild-ostree && dnf clean all
 COPY --from=builder images/osbuild-deploy-container /usr/bin/osbuild-deploy-container
-COPY entrypoint.sh /
+COPY prepare.sh entrypoint.sh /
 COPY --from=builder images/dnf-json .
 
 ENTRYPOINT ["/entrypoint.sh"]
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ A simpler container for deploying bootable container images.
 
 ```
 mkdir output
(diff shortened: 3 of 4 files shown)


achilleas-k:
Testing on the distros you listed using osbuild `main` to include the loop device fix.

achilleas-k:
Manually tested the build process with this PR and it's working on:
- Fedora 39
- Fedora 38
- CentOS Stream 9
- CentOS Stream 8
- Ubuntu 24.04 (Noble Numbat)

teg:
If this is meant to work on an M1, let me know and I would happily sanity check!

ondrejbudai:
This is now ready to be merged.

@teg AFAIK, the container should work on ARM. Once this is merged, you can just build the container yourself (clone the repository and `sudo podman build`) and run a build using instructions from `README.md`. I don't think it's worth blocking on this now. Proper ARM builds are definitely on the roadmap.

achilleas-k (approved):

prepare.sh
```diff
@@ -0,0 +1,30 @@
...
+# Copy osbuild to the new mountpoint.
+cp /usr/bin/osbuild "${TMP}/osbuild"
+
+# Label it as `install_exec_t`. We need this in order to get `install_t` that has `CAP_MAC_ADMIN` for creating SELinux
+# labels unknown to the host.
+#
+# Note that the transition to `install_t` must happen at this point. Osbuild stages run in `bwrap` that creates
+# a nosuid, no_new_privs environment. In such an environment, we cannot transition from `unconfined_t` to `install_t`,
+# because we would get more privileges.
+chcon system_u:object_r:install_exec_t:s0 "${TMP}/osbuild"
```
rhatdan, line 26: It would have been better to do

chcon -t install_exec_t "${TMP}/osbuild"

But this is fine.

teg:
> @teg AFAIK, the container should work on ARM. Once this is merged, you can just build the container yourself (clone the repository and `sudo podman build`) and run a build using instructions from `README.md`. I don't think it's worth blocking on this now. Proper ARM builds are definitely on the roadmap.

Sanity tested on my M2. The container builds and runs, but we run into the (expected?) loop issue.

PR #10, opened 2023-11-24
github: run build workflow on the merge queue

achilleas-k:
I just enabled the merge queue for this repository.  Let's run the container build workflow on the queue.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -8,6 +8,8 @@ on:
   workflow_dispatch:
   # for merge queue
   merge_group:
+  push:
+    branches: [main]
 
 env:
   REGISTRY: ghcr.io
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -8,6 +8,8 @@ on:
   workflow_dispatch:
   schedule:
```
ondrejbudai, line 9: Let's remove the `schedule` for now? I took this workflow from a different project and didn't care much about it, but it might be the right time to remove it since we are cleaning up the project. Note that if we do that, we also need to remove some conditions from `jobs`.

achilleas-k:
Removed the scheduled builds and limited the push actions to only run when an update lands in `main`, or when triggered manually.

ondrejbudai (approved):

PR #11, opened 2023-11-24
Add source for osbuild-deploy-container command

achilleas-k:
Partially replaces #5.

This PR adds the source code for the `osbuild-deploy-container` command, which up until now was only available [in a branch on my fork of the images repository](https://github.com/achilleas-k/images/tree/bifrost-image).

It depends on some open PRs in the osbuild/images repository for now.

Depends on:
- [x] osbuild/images#272
- [x] osbuild/images#274


ondrejbudai:
Since we have no intentions of pushing this to RHEL for now, what if we don't vendor? We can always start if we need to.

achilleas-k:
Noticed a couple of other issues here:
- We need to get dnf-json from images or osbuild (we should actually start using the new packaged version from osbuild).
- The Containerfile doesn't copy the sources into the builder.

ondrejbudai:
> Sure, we can drop the vendoring. I imagine that might make builds a bit slower (and annoying when rebuilding locally) since after cloning the repo and starting the container build, it will have to fetch all the dependencies inside the container.

That's true. :thinking: Not sure if this trade-off is worth it.

achilleas-k:
This is ready now.

teg (approved):
🥳

PR #16, opened 2023-11-27
prepare.sh: label osbuild,setfiles "correctly" in the container

mvo5:
When running osbuild inside a container the selinux setup is slightly different from a non container setup:

Inside the container all files under "/" in the overlayfs have the label `system_u:object_r:container_files_t`. This includes `/usr/bin/osbuild` and `/usr/sbin/setfiles`. Because the container is using an overlayfs the files cannot directly relabled under "/" with the correct selinux labels. This is why `prepare.sh` creates a tmpfs and relabels there and bind mounts back.

This commit tweaks this mechanism now to create labels that are closer to a real selinux system. I.e. it will:
1. label /usr/bin/osbuild as `osbuild_exec_t`
2. label /usr/sbin/setfiles as `setfiles_exec_t`

With that the normal transition rules work, i.e. osbuild_exec_t can transition to mount_t and when setfiles is called it will automatically use `setfiles_exec_t`.

With this PR I do not see any selinux denials when running the container anymore. I want to do a regression test for this as well to ensure that we don't regress and that this can be easily tested on multiple versions of fedora/rhel but that will have to be a followup as it's a bit of work.

Fwiw, AFAICT the root mount is not actually mounted "nosuid" (at last on my fc38 test system), I think the denials are from the fact that setfiles was not labeled correctly before.

[edit: setting to draft until some knowledgeable people had the chance to look at this first :)]
[edit2: an alternative would be to label {,u}mount inside the buildroot only, https://github.com/osbuild/images/compare/main...mvo5:bifrost/build-root-install_t-labeling?expand=1 - but this version here seem cleaner to me]

### DIFF
--- a/prepare.sh
+++ b/prepare.sh
@@ -3,8 +3,7 @@
 set -euo pipefail
 
 # Create a new tmpfs. This solves two issues for us:
-# - / is mounted as nosuid, this prevents SELinux to transition to `install_t` because domain transitions are
-#   disallowed if they give more caps to the process and the target executable is on `nosuid` filesystem
+# - / can be mounted as overlayfs with all files being `system_u:object_r:container_files_t`
 # - / can be mounted as OverlayFS that doesn't support overlaying SELinux labels. Thus, we need to ensure that
 #  the relabeling happens on a mountpoint that's definitely not an OverlayFS.
 TMP=/run/suidtmp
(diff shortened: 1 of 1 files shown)


ondrejbudai:
Does it work on systems without `osbuild-selinux` installed on the host? AFAIK, `osbuild_exec_t` is unknown to such hosts. I think we cannot assume that `osbuild-selinux` is everywhere.

teg:
> Does it work on systems without `osbuild-selinux` installed on the host? AFAIK, `osbuild_exec_t` is unknown to such hosts. I think we cannot assume that `osbuild-selinux` is everywhere.

On macos, podman runs the containers in a VM, where osbuild-selinux presumably is not available (though we could make it available by speaking to our fcos friends). Does it do that on Linux too, or are the containers running straight on the host? If the latter that might give us the wrong impression...

ondrejbudai:
We run directly on the host on Linux. Surely we can tell people to use podman-machine (and preinstall osbuild-selinux there), but forcing everyone to use a container AND a VM all the time doesn't feel convenient. Also, there are environments without virtualization (hello, AWS).

Thus, I would prefer if we can fix these issues without relying on our custom selinux policy.

mvo5:
Thanks! And indeed, this only works if the host has `osbuild-selinux` installed. Sorry that I missed this.

 I will close this one and go back to https://github.com/achilleas-k/images/pull/1 and clean this up and open it instead. That one works without the need for the osbuild-selinux package on the host.

PR #17, opened 2023-11-28
README.md: A few tweaks

cgwalters:
- Clarify functionality
- Link to centos-bootc
- Use new image reference
- Target c9s because ELN is more aggressive and has bugs

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,13 @@
 # osbuild-deploy-container
 
-A simpler container for deploying bootable container images.
+A tool that runs as a container that generates disk images from bootable container images, designed
+to pair with the [centos-bootc project](https://github.com/centos/centos-bootc).
 
 ## Example
 
 ```
 mkdir output
-sudo podman run --rm -it --privileged --security-opt label=type:unconfined_t -v $(pwd)/output:/output ghcr.io/osbuild/osbuild-deploy-container -imageref quay.io/centos-boot/fedora-tier-1:eln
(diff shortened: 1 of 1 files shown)


cgwalters:
Ah yes, switching to c9s is going to be a problem because of the hardcoded uefi vendor stuff:

```
  "uefi": {
    "vendor": "fedora",
    "install": true,
    "unified": true
  },
```

Which gets into the bootupd handling, which in turn gets into using bootc install.

achilleas-k:
> Ah yes, switching to c9s is going to be a problem because of the hardcoded uefi vendor stuff

Let's pull all that out into a config then.  Build-root repos are already externally defined but embedded at compile time.  We can do the same so that the binary is self contained or keep a config with some strings in the repo.

vrothberg (approved):
LGTM

cgwalters:
> Sounds like the kind of stuff that should be declared in the container metadata? at least if it declares the distro, we can have a mapping (as I believe we do) from distro to things like UEFI vendor.

Yes, in theory we could do that.  But this also leads to https://github.com/osbuild/bootc-image-builder/issues/18

where bootupd basically handles this dynamically.  Now in theory of course we could change the osbuild bootloader code to do the same; I actually don't understand why the vendor is hardcoded instead of being detected today.

ondrejbudai:
What about splitting this PR? I'm happy to make explicit in the README.md that this is paired with centos-bootc.

However, we are not ready to make the centos container the default experience, we still have some issues with them, see https://github.com/osbuild/bootc-image-builder/issues/20. Note that @mvo5 is working on a short-term fix for this, before we can fully switch to `bootc`.

cgwalters:
This is obsolete.

PR #19, opened 2023-11-28
README: add aarch64 instructions

teg:
This should not be merged quite yet. Firstly, there is no aarch64 version of osbulid-deploy-container in our container registry, so it needs to be built locally. Secondly, the bootable container we pull in is currently `:latest`, but should be `:eln` (which does not yet exist).


teg:
With these instructions this builds and boots on my M2.

PR #21, opened 2023-12-01
README: add `--pull=newer`

vrothberg:
To make sure Podman is updating the image if a newer one is on the registry.  The image doesn't specify a tag, so Podman should defailt to `:latest` which should imply `--pull=newer`.

I will take a look whether there's a Podman bug.


ondrejbudai (approved):

vrothberg:
Correction on my end, `:latest` does not default to pulling down the image.  Only Kubernetes does that but not Docker or Podman.

rhatdan:
LGTM

rhatdan:
Lets get this merged.

vrothberg:
Updated, we also need to update the tier-1 image to the new reference.  Can do that later.

PR #24, opened 2023-12-04
tests: add basic integration testing

mvo5:
This commit adds basic integration testing for the project. It is pytest based and can run both locally or via `tmt` [0] which will spin up a clean VM and run the tests inside.

See https://github.com/osbuild/osbuild-deploy-container/compare/main...mvo5:tmt-smoke?expand=1#diff-5de36acd90308dc62abf7855a686ee7052ffb6e762c756fd735fb0c9fbd9595dR1 (tests/README.md) for a very basic description.

Opening as draft as there are some things still todo:
[ ] integrate with tmt/testing farm inside the GH runners
[ ] boot the generated VM and test it, code for this already exists https://github.com/osbuild/osbuild-deploy-container/compare/main...mvo5:integration-test?expand=1 - maybe best as a followup?
[ ] selinux denails checks (need https://github.com/osbuild/images/pull/287)

[0] https://github.com/teemtee/tmt


achilleas-k:
I clicked the rebase button.  The linter was failing because the branch was from before the repo and package rename so it was failing.

test/test_smoke.py
```diff
@@ -0,0 +1,67 @@
...
+    }
+    config_json_path = output_path / "config.json"
+    config_json_path.write_text(json.dumps(CFG), encoding="utf-8")
+    return config_json_path
+
+
+@pytest.mark.skipif(os.getuid() != 0, reason="needs root")
+@pytest.mark.skipif(not testutil.has_executable("podman"), reason="need podman")
+def test_smoke(output_path, config_json):
+    # build local container
```
cgwalters, line 42: The interesting thing here of course is that I think it'd also make sense to support a flow that doesn't provision an intermediate VM and rebuild the container image in it, but just executes on an already built container. 

Among other things, that's the clear pattern we want to use for things like reverse dependency testing.

I think tmt in theory has some support for this.

test/test_smoke.py
```diff
@@ -0,0 +1,67 @@
...
+    ])
+    cursor = testutil.journal_cursor()
+    # and run container to deploy an image into output/disk.qcow2
+    subprocess.check_call([
+        "podman", "run", "--rm",
+        "--privileged",
+        "--security-opt", "label=type:unconfined_t",
+        "-v", f"{output_path}:/output",
+        "osbuild-deploy-container-test",
+        "quay.io/centos-boot/fedora-tier-1:eln",
```
cgwalters, line 56: This is the old image reference, see https://centos.github.io/centos-bootc/ for the latest

test/test_smoke.py
teg: @ondrejbudai would be great to capture these expected use cases, probably in the Feature, so we know what to test against - when you are back.

test/test_smoke.py
mvo5: Thank you! I updated the reference!

test/README.md
```diff
@@ -0,0 +1,12 @@
+Integration tests for osbuild-deploy-container
+----------------------------------------------
+
+This directory contans integration tests for osbuild-deploy-container.
```
achilleas-k, line 4: Should be renamed now.

achilleas-k (changes requested):

test/README.md
mvo5: Nice catch, thank you! Fixed now :)

ochosi:
@mvo5 Just fyi, I changed to PR title because the Jira reference needs to be a prefix for the automation to pick it up and link it.

mvo5:
> @mvo5 Just fyi, I changed to PR title because the Jira reference needs to be a prefix for the automation to pick it up and link it.

Thanks @ochosi ! Maybe it's just me (or me just not being used to this) but I wonder if we could change the automation to look at the end? I find it aesthetically more pleasing to keep the existing form of `<subsystem>: what changed` and just make that  `<subsystem>: what changed (JIRA-xxxx)`. Similar patterns are used e.g. in debian changelogs when bugs are closed (there its technically free form but the convention is that it's at the end). No super strong opinion though just an wondering because now it's early enough to tweak what we do :)

test/test_smoke.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+@pytest.mark.skipif(os.getuid() != 0, reason="needs root")
+@pytest.mark.skipif(not testutil.has_executable("podman"), reason="need podman")
+def test_smoke(output_path, config_json):
+    # build local container
+    subprocess.check_call([
+        "podman", "build",
+        "-f", "Containerfile",
+        "-t", "osbuild-deploy-container-test",
```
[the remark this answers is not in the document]
mvo5: And fixed too

ochosi:
> > @mvo5 Just fyi, I changed to PR title because the Jira reference needs to be a prefix for the automation to pick it up and link it.
> 
> Thanks @ochosi ! Maybe it's just me (or me just not being used to this) but I wonder if we could change the automation to look at the end? I find it aesthetically more pleasing to keep the existing form of `<subsystem>: what changed` and just make that `<subsystem>: what changed (JIRA-xxxx)`. Similar patterns are used e.g. in debian changelogs when bugs are closed (there its technically free form but the convention is that it's at the end). No super strong opinion though just an wondering because now it's early enough to tweak what we do :)

I wholeheartedly agree, but it seems this is hardcoded in Jira's DVCS plugin, so there's little we can do about it.

PR #33, opened 2023-12-06
HMS-3162: tests: run basic integration test on macOS too

mvo5:
This is mostly here to show that the GH macos  runners can do a basic darwin (x86 though) test [0] (and that it works with the right podman setup) and as a starting point for discussion with the podman team (see below).

It's (very) quirky right now and I will probably need some help from the podman team around the issues that arise from sharing a folder on macOS between (macOS) host and the container. The second commit is the relevant one, the PR is build on top of https://github.com/osbuild/bootc-image-builder/pull/24

[0] A arm64 test can be added via AWS or via the `macos-13-xlarge` runners


mvo5:
Closin in favor of #75

PR #36, opened 2023-12-06
HMS-3162: tests: add initial testing farm integration

mvo5:
This adds initial testing farm integration via a github action. To run jobs on testing farm a token is required that is stored as a repository secret. For security reasons repository secrets are not visible accross forks [0].

There are multiple ways to work around this limiation, this commit goes with the suggestion from [1], i.e.: the workflow is run within the `pull_request_target` trigger which has access to secrets.

This means the (potentially untrusted) branch is only checked out if the person triggering the workflow has already write access to the repository (we could make this restriction strong but it seems a reasonable permisson level). In practise the workflow will fail for outside contributions but a re-trigger from anyone in the term should be enough to get it tested inside the testing farm.

[0] https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
[1] https://michaelheap.com/access-secrets-from-forks/

### DIFF
--- a/.github/workflows/testingfarm.yml
+++ b/.github/workflows/testingfarm.yml
@@ -1,0 +1,52 @@
+---
+name: Testing farm tests
+
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened]
+
+# To use testing farm we need the TF_API_KEY secret available inside the
+# forked repo which requires the pull_request_target trigger. To protect
+# the secrets we need to make sure only people with repo write access
+# can trigger this workflow. This means that ouside contributors will
+# get an initial failure when the workflow is run. But once someone from
+# the team re-triggers it it will work.
+#
+# Note that "pull_requqest_target" events are always triggered even
+# when the "Fork pull request workflows from outside collaborators"
+# setting is restricted to "Require approval for all outside collaborators"
+# (see https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks)
+#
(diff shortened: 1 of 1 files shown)


.github/workflows/testingfarm.yml
```diff
@@ -0,0 +1,42 @@
+---
+name: Testing farm tests
+
+on:
+  pull_request_target:
```
croissanne, line 5: Any reason to use `pull_request_target` over `pull_request` ? Afaik the only difference is that `pull_request_target` gets a token which can write to the repository, but I'm not sure if that's needed here.

mvo5: Thanks - maybe I should spell this out better. My understanding is that the addition difference is that pull_request_target gets access to the `secrets.TF_API_KEY` across forks but that the regular `pull_request` does not (as a security measure). But maybe I'm missing something here?

ochosi: Yes, afaik that's correct. It's a way to work around the security restrictions of GH Actions and forks.

There's some precedence to outside contributors not being able to trigger CI. We have a "friends of osbuild" group that we add contributors to that we trust and they can trigger CI. Everyone else has to wait for a team member to push the button.

thrix:
LGTM from Testing Farm side, it is sad it cannot be tested on MR :)

As discussed on Slack, if you would need /dev/kvm

1. use the RH ranch token
2. use this in the `hardware` field of the `provision` plugin in the plan

```
hardware:
  virtualization:
    is-supported: true
```

PR #41, opened 2023-12-07
HMS-3141: tests,bib: fix/check that there are no selinux denials

mvo5:
This test ensure that we build images without selinux denials in the logs.


mvo5:
Once https://github.com/osbuild/images/pull/287 is merged and we set `ContainerBuildable` here in the `bib` tree this should turn green :)

PR #44, opened 2023-12-08
Update README.md

teg:
Add instructions for running qemu on macOS M1 and update second Podman invocation.


supakeen:
Let's mention somewhere that this needs qemu installed through macports (and link to their downloads page: https://www.qemu.org/download/#macos).

The same thing works for me on an M2.

teg:
Yeah, I used brew to get this to work, but adding the instructions can't hurt. At some point I ended up with the x86 version which caused my a world of pain.

supakeen:
Let's merge so it can be iterated on.

supakeen (approved):

PR #48, opened 2023-12-11
test: fix race in wait_ssh_ready() and add test

mvo5:
The wait_ssh_ready() had a subtle bug - when ssh was starting up and not yet producing the expected output (e.g. just empty output) the loop that sleeps would not wait but just loop very quickly which could lead to a confuging:
```
ConnectionRefusedError: cannot connect to port 39973 after 600s
```
even though clearly 600s had not passed.

This commit fixes this (subtle) bug and adds a test that will generate an unexpected output and ensures there that the expected number of sleeps happend.


supakeen (dismissed):
Nice :bug:.

ondrejbudai (dismissed):

test/test_smoke.py
```diff
@@ -91,9 +91,14 @@ def test_smoke(output_path, config_json):
...
-    with VM(generated_img) as test_vm:
-        exit_status, _ = test_vm.run("true", user="test", password="password")
-        assert exit_status == 0
-        exit_status, output = test_vm.run("echo hello", user="test", password="password")
-        assert exit_status == 0
-        assert "hello" in output
+    # building an ELN image needs x86_64-v3 to work, we use avx2 as a proxy
+    # to detect if we have x86-64-v3 (not perfect but should be good enough)
+    if " avx2 " not in pathlib.Path("/proc/cpuinfo").read_text():
+        print("WARNING: no x86_64-v3 cpu detected, skipping VM boot test")
```
mvo5, line 97: Fwiw, this is ugly, I will look into making this a proper "pytest.mark.skipif" somehow

mvo5: Either way is fine with me, I think it's okay to land and unblock PRs but I also really want to push a PR before my EOD that untangles this mess a bit more

mvo5: But might be best as a followup just to have an easier review/discussion

supakeen (approved):

PR #49, opened 2023-12-11
test: refactor monolithic `test_smoke` into multiple tests

mvo5:
The existing test_smoke was becoming a bit too big and especially the fact that part have to be skipped depending on the environment is really not nice.

This commit refactors it to be more module. For this it uses the pytest session scope to ensure that the build container and the test image are only build once and then shared accross the various individual tests.

### DIFF
--- a/test/test_smoke.py
+++ b/test/test_smoke.py
@@ -1,3 +1,4 @@
+import collections
 import json
 import os
 import pathlib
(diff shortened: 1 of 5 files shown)


test/test_smoke.py
```diff
@@ -11,39 +12,97 @@
...
+    subprocess.check_call([
+        "podman", "build",
+        "-f", "Containerfile",
+        "-t", "bootc-image-builder-test",
+    ])
+
 
+@pytest.fixture(name="build_image", scope="session")
+def build_image_fixture(tmpdir_factory, build_container):
+    ret = collections.namedtuple(
```
bcl, line 38: This confused me :) by the time I read the end of the function I'd forgotten about the namedtuple so I went searching for a function named `ret()`, maybe move it to the end or name it something a bit more descriptive.

bcl (dismissed):
Looks good!

mvo5:
This includes now https://github.com/osbuild/bootc-image-builder/pull/50 which apparently was required to make the RHTAP build work (it was failing before without this).

supakeen (approved):

PR #54, opened 2023-12-13
HMS-3276: Add support for building AMIs

achilleas-k:
Add support for building an AMI with a `--type` flag.  The only difference between the qcow2 and the AMI is the output file format, which for AMI is raw.

The default image type is qcow2.

Tests now build and boot both image types.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /.idea
 /output
 /bin
+__pycache__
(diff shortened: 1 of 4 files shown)


ondrejbudai (approved):
Looks good. Shall we do an actual booting in AWS in the test-suite as a follow-up?

achilleas-k:
> Looks good. Shall we do an actual booting in AWS in the test-suite as a follow-up?

Yeah.  Maybe we can reuse the images boot tests.

mvo5 (approved):
I like this!

mvo5 (approved):
I still like it :)

PR #57, opened 2023-12-14
Fix artifact ownership problems

### LINKED ISSUE #14: `cp: failed to preserve ownership for` in Podman Desktop on Mac
@teg reported that the build failed for him when running the build on M1 Mac:

```
⏱  Duration: 32s
cp: failed to preserve ownership for '/output/qcow2/./disk.qcow2': Operation not permitted
cp: failed to preserve ownership for '/output/qcow2/.': Operation not permitted
  File "/usr/bin/osbuild", line 33, in <module>
    sys.exit(load_entry_point('osbuild==101', 'console_scripts', 'osbuild')())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/osbuild/main_cli.py", line 179, in osbuild_cli
    export(pid, output_directory, object_store, manifest)
  File "/usr/lib/python3.12/site-packages/osbuild/main_cli.py", line 56, in export
    obj.export(dest)
  File "/usr/lib/python3.12/site-packages/osbuild/objectstore.py", line 249, in export
    subprocess.run(
subprocess.CalledProcessError: Command '['cp', '--reflink=auto', '-a', '/store/stage/uuid-c6ab67ad31ed40abae8922411cadb303/data/tree/.', '/output/qcow2']' returned non-zero exit status 1.
running osbuild failed: exit status 1
teg@MacBook-Pro osbuild-deploy-container % 
```

achilleas-k:
osbuild uses `cp -a` to export artifacts from a build and files are (usually) owned by root.  The `cp` command tries to set the ownership to root on the mounted volume from the host but if it's running in a podman machine, this will fail.

Added a new check, `canChownInPath()`, which creates a temporary file in the output directory and tries to `chown` it to the uid and gid of the caller.  This lets us know if osbuild will be able to preserve the permissions of the output file when copying the results, which should be created with the uid and gid of the caller of osbuild, to the output directory.

This is used to set export option when calling osbuild, `OSBUILD_EXPORT_FORCE_NO_PRESERVE_OWNER`.  The option prevents osbuild from trying to set the ownership of the files in the destination and failing.

Requires osbuild/osbuild#1511 or osbuild/osbuild#1512.

Output directory on macOS
```
$ ls -lR output
total 384
drwxr-xr-x  3 achilleas  staff     96 Dec 14 15:46 image
-rw-r--r--  1 achilleas  staff  95668 Dec 14 15:48 manifest-ami.json
-rw-r--r--  1 achilleas  staff  97855 Dec 14 15:57 manifest-qcow2.json
drwxr-xr-x  3 achilleas  staff     96 Dec 14 15:40 qcow2

output/image:
total 6496120
-rw-r--r--  1 achilleas  staff  10737418240 Dec 14 15:48 disk.raw

output/qcow2:
total 1278064
-rw-r--r--  1 achilleas  staff  640516096 Dec 14 15:58 disk.qcow2
```

Fixes #14

### DIFF
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -43,6 +43,11 @@ type BuildConfig struct {
 	Blueprint *blueprint.Blueprint `json:"blueprint,omitempty"`
 }
 
+var (
+	osGetuid = os.Getuid
+	osGetgid = os.Getgid
+)
+
 // canChownInPath checks if the ownership of files can be set in a given path.
 func canChownInPath(path string) (bool, error) {
 	info, err := os.Stat(path)
@@ -63,7 +68,7 @@ func canChownInPath(path string) (bool, error) {
 			fmt.Fprintf(os.Stderr, "error deleting %s: %s\n", checkFile.Name(), err.Error())
 		}
 	}()
-	return checkFile.Chown(os.Getuid(), os.Getgid()) == nil, nil
+	return checkFile.Chown(osGetuid(), osGetgid()) == nil, nil
 }
 
 // Parse embedded repositories and return repo configs for the given
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -43,6 +43,29 @@ type BuildConfig struct {
 	Blueprint *blueprint.Blueprint `json:"blueprint,omitempty"`
 }
 
+// canChownInPath checks if the ownership of files can be set in a given path.
+func canChownInPath(path string) (bool, error) {
+	info, err := os.Stat(path)
+	if err != nil {
+		return false, err
+	}
+	if !info.IsDir() {
+		return false, fmt.Errorf("%s is not a directory", path)
+	}
+
+	checkFile, err := os.CreateTemp(path, ".writecheck")
+	if err != nil {
+		return false, err
+	}
+	defer func() {
+		if err := os.Remove(checkFile.Name()); err != nil {
+			// print the error message for info but don't error out
+			fmt.Fprintf(os.Stderr, "error deleting %s: %s\n", checkFile.Name(), err.Error())
+		}
+	}()
+	return checkFile.Chown(os.Getuid(), os.Getgid()) == nil, nil
+}
+
 // Parse embedded repositories and return repo configs for the given
 // architecture.
 func loadRepos(archName string) []rpmmd.RepoConfig {
@@ -152,6 +175,9 @@ func build(cmd *cobra.Command, args []string) {
 		upload = true
 	}
 
+	canChown, err := canChownInPath(outputDir)
+	check(err)
+
 	config := BuildConfig{}
 	if configFile != "" {
 		config = loadConfig(configFile)
@@ -186,7 +212,12 @@ func build(cmd *cobra.Command, args []string) {
 
 	fmt.Printf("Building %s\n", manifest_fname)
 
-	_, err = osbuild.RunOSBuild(mf, osbuildStore, outputDir, exports, nil, nil, false, os.Stderr)
+	var osbuildEnv []string
+	if !canChown {
+		// set export options for osbuild
+		osbuildEnv = []string{"OSBUILD_EXPORT_FORCE_NO_PRESERVE_OWNER=1"}
+	}
+	_, err = osbuild.RunOSBuild(mf, osbuildStore, outputDir, exports, nil, osbuildEnv, false, os.Stderr)
 	check(err)
 
 	fmt.Println("Build complete!")
--- a/bib/cmd/bootc-image-builder/export_test.go
+++ b/bib/cmd/bootc-image-builder/export_test.go
@@ -1,0 +1,11 @@
+package main
+
+var CanChownInPath = canChownInPath
+
+func MockOsGetuid(new func() int) (restore func()) {
+	saved := osGetuid
+	osGetuid = new
+	return func() {
+		osGetuid = saved
+	}
+}
--- a/bib/cmd/bootc-image-builder/main_test.go
+++ b/bib/cmd/bootc-image-builder/main_test.go
@@ -1,0 +1,45 @@
+package main_test
+
+import (
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	main "github.com/osbuild/bootc-image-builder/bib/cmd/bootc-image-builder"
+)
+
+func TestCanChownInPathHappy(t *testing.T) {
+	tmpdir := t.TempDir()
+	canChown, err := main.CanChownInPath(tmpdir)
+	require.Nil(t, err)
+	assert.Equal(t, canChown, true)
+
+	// no tmpfile leftover
+	content, err := os.ReadDir(tmpdir)
+	require.Nil(t, err)
+	assert.Equal(t, len(content), 0)
+}
+
+func TestCanChownInPathNotExists(t *testing.T) {
+	canChown, err := main.CanChownInPath("/does/not/exists")
+	assert.Equal(t, canChown, false)
+	assert.ErrorContains(t, err, ": no such file or directory")
+}
+
+func TestCanChownInPathCannotChange(t *testing.T) {
+	if os.Getuid() == 0 {
+		t.Skip("cannot run as root (fchown never errors here)")
+	}
+
+	restore := main.MockOsGetuid(func() int {
+		return -2
+	})
+	defer restore()
+
+	tmpdir := t.TempDir()
+	canChown, err := main.CanChownInPath(tmpdir)
+	require.Nil(t, err)
+	assert.Equal(t, canChown, false)
+}
--- a/bib/go.mod
+++ b/bib/go.mod
@@ -9,6 +9,7 @@ require (
 	github.com/sirupsen/logrus v1.9.3
 	github.com/spf13/cobra v1.8.0
 	github.com/spf13/pflag v1.0.5
+	github.com/stretchr/testify v1.8.4
 )
 
 require (
@@ -22,6 +23,7 @@ require (
 	github.com/containers/ocicrypt v1.1.9 // indirect
 	github.com/containers/storage v1.51.0 // indirect
 	github.com/cyberphone/json-canonicalization v0.0.0-20231011164504-785e29786b46 // indirect
+	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/distribution/reference v0.5.0 // indirect
 	github.com/docker/distribution v2.8.3+incompatible // indirect
 	github.com/docker/docker v24.0.7+incompatible // indirect
@@ -67,6 +69,7 @@ require (
 	github.com/opencontainers/runc v1.1.10 // indirect
 	github.com/opencontainers/runtime-spec v1.1.0 // indirect
 	github.com/pkg/errors v0.9.1 // indirect
+	github.com/pmezard/go-difflib v1.0.0 // indirect
 	github.com/proglottis/gpgme v0.1.3 // indirect
 	github.com/rivo/uniseg v0.4.4 // indirect
 	github.com/secure-systems-lab/go-securesystemslib v0.7.0 // indirect
--- a/bib/go.sum
+++ b/bib/go.sum
@@ -270,6 +270,7 @@ github.com/stretchr/testify v1.7.1/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/
 github.com/stretchr/testify v1.8.0/go.mod h1:yNjHg4UonilssWZ8iaSj1OCr/vHnekPRkoO+kdMU+MU=
 github.com/stretchr/testify v1.8.1/go.mod h1:w2LPCIKwWwSfY2zedu0+kehJoqGctiVI29o6fzry7u4=
 github.com/stretchr/testify v1.8.4 h1:CcVxjf3Q8PM0mHUKJCdn+eZZtm5yQwehR5yeSVQQcUk=
+github.com/stretchr/testify v1.8.4/go.mod h1:sz/lmYIOXD/1dqDmKjjqLyZ2RngseejIcXlSw2iwfAo=
 github.com/syndtr/gocapability v0.0.0-20200815063812-42c35b437635 h1:kdXcSzyDtseVEc4yCz2qF8ZrQvIDBJLl4S1c3GCXmoI=
 github.com/syndtr/gocapability v0.0.0-20200815063812-42c35b437635/go.mod h1:hkRG7XYTFWNJGYcbNJQlaLq0fg1yr4J4t/NcTQtrfww=
 github.com/tidwall/pretty v1.0.0/go.mod h1:XNkn88O1ChpSDQmQeStsy+sBenx6DDtFZJxhVysOjyk=


entrypoint.sh
```diff
@@ -3,4 +3,8 @@
 set -euo pipefail
 
 ./prepare.sh
-/usr/bin/bootc-image-builder --store /store --rpmmd /rpmmd --output /output  "$@"
+tmpout="$(mktemp -p / -d)"
+/usr/bin/bootc-image-builder --store /store --rpmmd /rpmmd --output "${tmpout}" "$@"
+echo "Moving artifacts to output directory."
+cp -rf "${tmpout}/." /output
```
rhatdan, line 9: Doesn't this still end up with the artifact owned by root? Since root is doing the cp?

rhatdan: It should at least ensure that the file is world readable.

cgwalters, line 9: I don't quite understand why we have this bash script instead of doing things in Go?

Also in the end it'd be a good bit more efficient to create images in their final destination to avoid copying.

achilleas-k: We have the script so we can call `/prepare.sh` which we need for other reasons.  I could do it in Go, that's not an issue.

But yes, this essentially means we're copying the artifact twice.  It's not terrible, but it'd be simpler to export directly to our destination.  The problem there is what I mentioned in the original comment: `cp -a` will fail to preserve the ownership if it can't set it on the destination, which is what's happening here and osbuild using `cp -a` to export artifacts is important for all sorts of scenarios (copying entire system root trees for example).

mvo5: Yeah, we can easily test this by doing a test branch with this one here and https://github.com/osbuild/bootc-image-builder/pull/33 and removing the workarounds (i.e. https://github.com/osbuild/bootc-image-builder/pull/33/files#diff-1db27d93186e46d3b441ece35801b244db8ee144ff1405ca27a163bfe878957fR103) - I can do this later today or in my morning if desired

ondrejbudai (changes requested):
Interesting that this works. I would love to know what's happening in `9p` that `cp` works, but `cp -a` isn't.

The more important bit: I think I agree with Colin, this should be in Go, if there's no good reason not to put it there.

entrypoint.sh
rhatdan: The problem is on a Mac when you mount the homedir into the MAC, it can only store files owned by the host user. Imaging the User UID=1000. When the user runs the VM, it is running as UID=1000.  When the user picks a directory for /output that is in their homedir, then bootc-image-builder attempts to create a UID==0 file in the users homedir, which ends up with the VM process (qemu) running as UID=1000 attempting to create a file UID=0 on the host system. The MAC Os does not allow this.  

On Windows this does not cause issues because the UIDs/GIDs mean nothing on Windows.
On Linux the qemu process is running within the User Namespace, so that UID=0 within the processes User namespace ends up being UID=1000 in the users homedir.

So this is only an issue on Macs.

rhatdan:
cp -a attempts to maintain the ownership bits of the source file.

Let's look at some examples running as root.

First I create a file /foo as root which will be owned as root, then I copy the file to the users homedir, and the new file is owned as root. (This is what is blowing up on the MAC. cp creates files that do not exist based on the ownership UID of the process running `cp`, so file is owned by root.
```
# rm -rf /home/dwalsh/foo
# touch /foo
# cp /foo ~dwalsh/
# ls -l ~dwalsh/foo
-rw-r--r--. 1 root root 0 Dec 15 07:20 /home/dwalsh/foo
```

Now I show an example of what happens if the file pre-existed in the homedir and the root process replaced the file with the file owned by root.  In this case the file ends up owned by the user.  Since `cp` preserves the ownership of the destination file if it exists. This means even though the process running `cp` is root, the replacement file is still owned by dwalsh.
```
# chown dwalsh ~dwalsh/foo
# cp /foo ~dwalsh/
# ls -l ~dwalsh/foo
-rw-r--r--. 1 dwalsh root 0 Dec 15 07:21 /home/dwalsh/foo
```

When you run `cp -a` you are telling cp to maintain the ownership of the source file when replacing the destination file.
```
# cp -a /foo ~dwalsh/
# ls -l ~dwalsh/foo
-rw-r--r--. 1 root root 0 Dec 15 07:20 /home/dwalsh/foo
```

I am not sure your new containerimage works correctly if the target file does not pre-exist, or at least I don't understand what is going on, since the process inside of the container running cp is running as root.

achilleas-k:
> cp -a attempts to maintain the ownership bits of the source file.

But there's also some interesting interactions with the way volumes are mounted in podman machine.  If `cp` can't set the ownership in the destination, and `-a` isn't specified, it wont, and I think the 9p `access=client` option that podman machine uses is what's controlling this.  You can see this if you do stuff onto the mounted volume from inside the machine:
```
achilleas@mmini ~ % podman machine ssh 'sudo touch /root/rootfile && sudo cp /root/rootfile /Users/achilleas/rootfile'
achilleas@mmini ~ % podman machine ssh 'sudo ls -l /root/rootfile /Users/achilleas/rootfile'
-rw-r--r--. 1 core games 0 Dec 14 16:42 /Users/achilleas/rootfile
-rw-r--r--. 1 root root  0 Dec 14 16:42 /root/rootfile
achilleas@mmini ~ % ls -l /Users/achilleas/rootfile
-rw-r--r--  1 achilleas  staff  0 Dec 14 16:42 /Users/achilleas/rootfile
```

Or even if you try to set perms:
```
achilleas@mmini ~ % podman machine ssh
Connecting to vm podman-machine-default. To close connection, use `~.` or `exit`
Fedora CoreOS 39.20231204.2.1
Tracker: https://github.com/coreos/fedora-coreos-tracker
Discuss: https://discussion.fedoraproject.org/tag/coreos

Last login: Fri Dec 15 14:43:44 2023 from 192.168.127.1
root@localhost:~# pwd
/root
root@localhost:~# touch rootfile
root@localhost:~# ls -l rootfile
-rw-r--r--. 1 root root 0 Dec 15 14:44 rootfile
root@localhost:~# cp -a rootfile /Users/achilleas/
cp: failed to preserve ownership for '/Users/achilleas/rootfile': Operation not permitted
root@localhost:~# ls -l /Users/achilleas/rootfile
-rw-------. 1 core games 0 Dec 15 14:44 /Users/achilleas/rootfile
root@localhost:~# chown root:root /Users/achilleas/rootfile
chown: changing ownership of '/Users/achilleas/rootfile': Operation not permitted
root@localhost:~# rm /Users/achilleas/rootfile
root@localhost:~# cp rootfile /Users/achilleas/
root@localhost:~# ls -l /Users/achilleas/rootfile
-rw-r--r--. 1 core games 0 Dec 15  2023 /Users/achilleas/rootfile
root@localhost:~#
logout
achilleas@mmini ~ % ls -l /Users/achilleas/rootfile
-rw-r--r--  1 achilleas  staff  0 Dec 15 14:57 /Users/achilleas/rootfile
```

Now this is on the VM, not inside the container that's running in the VM, but the same rules should apply.

> The problem is on a Mac when you mount the homedir into the MAC, it can only store files owned by the host user. Imaging the User UID=1000. When the user runs the VM, it is running as UID=1000. When the user picks a directory for /output that is in their homedir, then bootc-image-builder attempts to create a UID==0 file in the users homedir, which ends up with the VM process (qemu) running as UID=1000 attempting to create a file UID=0 on the host system. The MAC Os does not allow this.
> 
> On Windows this does not cause issues because the UIDs/GIDs mean nothing on Windows.
> On Linux the qemu process is running within the User Namespace, so that UID=0 within the processes User namespace ends up being UID=1000 in the users homedir.
> 
> So this is only an issue on Macs.

This is close to the argument I want to make for having it in the entrypoint.  This is only an issue in containers and only when running on macOS (or possibly also on Linux with podman machine?).  I have it working as part of the bootc-image-builder binary, but that means that if you ever run the command outside a container (which should be fine, why not support this case?), then you're doing an unnecessary extra copy that was added to get around a specific container-related case even when you don't need it.  And to me it always feels strange when tools have container-specific behaviour built-in that still happens when they're not containerised.

achilleas-k:
Pushed the go code.

achilleas-k:
Tests are failing because of lack of space, which makes sense since the AMI is 10 GiB.  The previous tests with `cp` in the entrypoint succeeded, probably because it preserves file sparseness and the `io.Copy()` I used here doesn't.
We could keep the copy operation in the Go code but shell out to `cp` instead of doing it natively.

bib/cmd/bootc-image-builder/util.go
```diff
@@ -0,0 +1,68 @@
...
+
+import (
+	"fmt"
+	"io"
+	"io/fs"
+	"os"
+	"path/filepath"
+)
+
+func copyFile(src, dest string) error {
```
mvo5, line 11: I was starting to review this and wanted to mention that we want/need tests for this and then I stepped back and wondered if we cannot use one of the existing go modules for this? Maybe https://pkg.go.dev/github.com/otiai10/copy (I really don't mind which one fwiw). Wdyt? If we want to roll our own I think we need a bunch of basic unit tests and I'm happy to come back and review copy{Dir,File} :)

mvo5 (commented):
Thank you! This looks nice overall and it's great that we fix the permission errors. I have some question/suggestions inline.

bib/cmd/bootc-image-builder/main.go
```diff
@@ -174,11 +174,23 @@ func build(cmd *cobra.Command, args []string) {
 	check(saveManifest(mf, manifestPath))
 
 	fmt.Printf("Building %s\n", manifest_fname)
+	check(runOSBuild(mf, osbuildStore, outputDir, exports))
+	fmt.Printf("Build complete. Results saved in %s\n", outputDir)
+	return nil
```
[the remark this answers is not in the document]
achilleas-k: The `build` command is the value of the `RunE` field in the `cobra.Command` struct so it needs to return an error.  We can use the `Run` field though.
But what we should actually do, I think, is pass errors up when they're related to bad command line arguments, since returning an error from the `RunE` function of the command will print the usage, which is a good idea in such cases.

bib/cmd/bootc-image-builder/main.go
```diff
@@ -174,11 +174,23 @@ func build(cmd *cobra.Command, args []string) {
...
+}
+func runOSBuild(mf manifest.OSBuildManifest, osbuildStore, outputDir string, exports []string) error {
+	tempOutput, err := os.MkdirTemp("/", "tmpout")
+	if err != nil {
+		return err
+	}
 
-	_, err = osbuild.RunOSBuild(mf, osbuildStore, outputDir, exports, nil, nil, false, os.Stderr)
-	check(err)
+	defer os.Remove(tempOutput)
```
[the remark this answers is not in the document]
achilleas-k: It should be `RemoveAll()`.  It's actually silently failing now.

bib/cmd/bootc-image-builder/util.go
achilleas-k: So here's where I'm at with this:
- Roll our own: Good for keeping things small.  We know what we're exporting and can focus on just that.  For example we know we're building single regular files and don't need to cover cases like copying symlinks or preserving hard links or even needing to make decisions like that in the first place.  Though like you said, it'd be good to have unit tests, even though the use case is covered by the integration test.
- Include a separate package: Overkill?  It'd be nice if there was one that handled sparse files.  Feels a bit unnecessary to be pulling in packages that can handle all sorts of tree-copying situations for a single file in a subdirectory that is entirely predictable.
- Shell out to `cp`: Preserves sparseness.  It's a "dependency" of sorts but ... I mean it's `cp`.
- Run `cp` in the entrypoint (previous version of this PR): I'm still not convinced this is a bad idea even though it seems everyone else disagrees.  This is a containerised edge case we're solving and the entrypoint is there to do "the container stuff".  Just looking at the whole conversation and all the work we're doing to solve this is starting to feel a bit silly.

achilleas-k:
Updated with new approach: Check if we can set the ownership of files in the output directory and if we can't, use a new osbuild feature to disable the ownership preservation when exporting.

Converting the PR to draft until the osbuild PR is merged and released.

bib/cmd/bootc-image-builder/main.go
```diff
@@ -43,6 +43,29 @@ type BuildConfig struct {
 	Blueprint *blueprint.Blueprint `json:"blueprint,omitempty"`
 }
 
+// canChownInPath checks if the ownership of files can be set in a given path.
+func canChownInPath(path string) (bool, error) {
+	if info, err := os.Stat(path); err == nil {
```
mvo5, line 48: I wonder if we could use the standard go convention here, i.e.
```go
	info, err := os.Stat(path)
	if err != nil {
		return false, err
	}
	if !info.IsDir() {
		return false, fmt.Errorf("%s is not a directory", path)
	}
```
or am I missing something?

bib/cmd/bootc-image-builder/main.go
```diff
@@ -43,6 +43,29 @@ type BuildConfig struct {
...
+		}
+	} else {
+		return false, err
+	}
+	checkFile, err := os.CreateTemp(path, ".writecheck")
+	if err != nil {
+		return false, err
+	}
+	defer func() {
+		err := os.Remove(checkFile.Name())
```
mvo5, line 60: (super nitpick) - I like  the one line pattern for things like this, i.e. `if err := os.Remove(...); err != nil {` as it's slightly more compact and err is in the if scope only. But it's probably just personal preference :)

mvo5 (approved):
Thanks for this PR! It looks very nice and I like the new division of labor between bib and osbuild here! I have one small suggestion and a nitpit inline. On a higher level, I would love to see more tests, I attach a small draft of what I have in mind,  something like canChownInPath() I would love to see tested. Ideally the integration of it in main as well but that is harder and we can probably leave it to a macOS based workflow test :)

My ideas around some basic testing:
[0001-bib-add-basic-testing-for-canChownInPath.patch.txt](https://github.com/osbuild/bootc-image-builder/files/13723972/0001-bib-add-basic-testing-for-canChownInPath.patch.txt)

mvo5:
Fwiw  https://github.com/osbuild/osbuild/pull/1512 is now merged so this should be unblocked now

bib/cmd/bootc-image-builder/main.go
achilleas-k: I prefer it too.  Not sure how I missed it there.

bib/cmd/bootc-image-builder/main.go
achilleas-k: Not missing anything.  I think I was just having a weird brain day.

achilleas-k:
> My ideas around some basic testing:
> [0001-bib-add-basic-testing-for-canChownInPath.patch.txt](https://github.com/osbuild/bootc-image-builder/files/13723972/0001-bib-add-basic-testing-for-canChownInPath.patch.txt)

Added.  Thanks for this.

bib/cmd/bootc-image-builder/main.go
```diff
@@ -174,11 +174,23 @@ func build(cmd *cobra.Command, args []string) {
 	check(saveManifest(mf, manifestPath))
 
 	fmt.Printf("Building %s\n", manifest_fname)
+	check(runOSBuild(mf, osbuildStore, outputDir, exports))
```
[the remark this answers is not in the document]
achilleas-k: Yeah, I get what you mean.  It's something I do for convenience with small dev tools or throwaway `main.go` files but I think you're right, this is not a little dev tool that few people look at any more, so it's probably better to use standard patterns.

mvo5 (approved):
\o/ Thank you!

bib/cmd/bootc-image-builder/main.go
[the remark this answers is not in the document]
mvo5: Totally fine to do in followups, sorry, was more a general comment :)

PR #58, opened 2023-12-14
iso: be able to build the ISO

supakeen:
Needs: https://github.com/osbuild/images/pull/322

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,7 @@ The following image types are currently available via the `--type` argument:
 |-----------------------|---------------------------------------------------------------------------------------|
 | `ami`                 | [Amazon Machine Image](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/AMIs.html) |
 | `qcow2` **(default)** | [QEMU](https://www.qemu.org/)                                                         |
+| `iso`                 | An unattended Anaconda installer that installs to the first disk found.               |
 
 ## ☁️ Cloud uploaders
 
--- a/bib/cmd/bootc-image-builder/image.go
+++ b/bib/cmd/bootc-image-builder/image.go
@@ -161,11 +161,10 @@ func pipelinesForISO(c *ManifestConfig, rng *rand.Rand) (image.ImageKind, error)
 		fail("pipeline: no base image defined")
 	}
 	ref := "ostree/1/1/0"
-	tlsVerify := true
 	containerSource := container.SourceSpec{
 		Source:    c.Imgref,
 		Name:      c.Imgref,
-		TLSVerify: &tlsVerify,
+		TLSVerify: &c.TLSVerify,
 	}
 
 	img := image.NewAnacondaContainerInstaller(containerSource, ref)
--- a/bib/cmd/bootc-image-builder/image.go
+++ b/bib/cmd/bootc-image-builder/image.go
@@ -55,6 +55,8 @@ func Manifest(c *ManifestConfig) (*manifest.Manifest, error) {
 		fallthrough
 	case "ami":
 		img, err = pipelinesForDiskImage(c, rng)
+	case "iso":
+		img, err = pipelinesForISO(c, rng)
 	default:
 		fail(fmt.Sprintf("Manifest(): unsupported image type %q", c.ImgType))
 	}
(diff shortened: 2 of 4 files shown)


bib/cmd/bootc-image-builder/image.go
```diff
@@ -154,6 +156,195 @@ func pipelinesForDiskImage(c *ManifestConfig, rng *rand.Rand) (image.ImageKind,
...
+			"anaconda-dracut",
+			"atheros-firmware",
+			"brcmfmac-firmware",
+			"curl",
+			"dracut-config-generic",
+			"dracut-network",
+			"hostname",
+			"iwlwifi-dvm-firmware",
+			"iwlwifi-mvm-firmware",
+			"kernel",
```
cgwalters, line 183: I don't understand why we have hardcoded "RPM package lists in Go code" here...what do we have in this that isn't part of https://src.fedoraproject.org/rpms/anaconda/blob/rawhide/f/anaconda.spec#_230 ?

There's another really big topic here which is in the *general* case...I think we need to be using the dependencies from the container.  Especially the kernel package - we want to support people adding e.g. 3rd party hardware drivers into their container build, and the ISO generated from that should use the same kernel, so that they can enable the NIC or storage.

I'm aware there's a lot of history here and we're reusing a existing logic, and that makes sense in the short term.

But the problem I'm talking about here I think is going to drive us towards an architecture where generating the ISO looks more like:

```dockerfile
FROM <provided base image>
RUN dnf -y install anaconda-install-img
```
(Plus then some further logic to convert *that* filesystem tree into an ISO)

bib/cmd/bootc-image-builder/image.go
```diff
@@ -154,6 +156,195 @@ func pipelinesForDiskImage(c *ManifestConfig, rng *rand.Rand) (image.ImageKind,
...
+	img.Groups = users.GroupsFromBP(customizations.GetGroups())
+
+	switch c.Architecture {
+	case arch.ARCH_X86_64:
+		img.Platform = &platform.X86{
+			BasePlatform: platform.BasePlatform{
+				ImageFormat: platform.FORMAT_ISO,
+			},
+			BIOS:       true,
+			UEFIVendor: "fedora",
```
cgwalters, line 331: (I still think this hardcoding of UEFI vendor in Go code is problematic...it gets fixed when using bootupd for images that use that; this relates to the above)

cgwalters (commented):
Thanks for working on this!

bib/cmd/bootc-image-builder/image.go
```diff
@@ -154,6 +156,195 @@ func pipelinesForDiskImage(c *ManifestConfig, rng *rand.Rand) (image.ImageKind,
...
+	}
+	ref := "ostree/1/1/0"
+	tlsVerify := true
+	containerSource := container.SourceSpec{
+		Source:    c.Imgref,
+		Name:      c.Imgref,
+		TLSVerify: &tlsVerify,
+	}
+
+	img := image.NewAnacondaContainerInstaller(containerSource, ref)
```
ondrejbudai, line 171: I hope to do a proper review later, but something I need to get off my head right now:

Can we switch to lz4 for now?
```
img.SquashfsCompression = "lz4"
```

I think that we should focus on speed first, and deal with artifact size later.

bib/cmd/bootc-image-builder/image.go
achilleas-k: The kernel and firmware and kernel args being the same for the ISO as for the base image is an understandable requirement (the machine that is going to run the image will have to be able to run the ISO as well), but do we really want to build the ISO from the base image directly, with all the things that the user might have put in their custom images?

It seems like there's so much that can go wrong there.  Packages (a user package that might conflict with an Anaconda dependency, for example), users and groups, services, are all things that can interfere with the installer environment.

Or is this not as likely and big a concern as I'm thinking?

bib/cmd/bootc-image-builder/image.go
achilleas-k: > I still think this hardcoding of UEFI vendor in Go code is problematic

It is, but until we have a robust way of detecting the distro of the container, we need to define it somewhere (or ask the user for it).
And since bootupd will solve this, should we be thinking of other solutions?

ondrejbudai: Wait, will bootupd solve this also for this ISO? I don't expect bootupd to be used there.

bib/cmd/bootc-image-builder/image.go
ondrejbudai: Well, I think that the base image doesn't need to be the payload container - but it can be a customer's "golden base image" that e.g. includes drivers for their environment.

I actually think that distributing the installer tree as a container image isn't a bad idea. I'm imagining something like this:

```
bootc-image-builder --type iso --iso-source ghcr.io/example/installer ghcr.io/example/payload
```

`ghcr.io/example/installer` would be defined in a similar way as Colin suggested. Alternatively, a distribution might just provide a base image if building one by users is too hard

bib/cmd/bootc-image-builder/image.go
```diff
@@ -154,6 +156,195 @@ func pipelinesForDiskImage(c *ManifestConfig, rng *rand.Rand) (image.ImageKind,
...
+func pipelinesForISO(c *ManifestConfig, rng *rand.Rand) (image.ImageKind, error) {
+	if c.Imgref == "" {
+		fail("pipeline: no base image defined")
+	}
+	ref := "ostree/1/1/0"
+	tlsVerify := true
+	containerSource := container.SourceSpec{
+		Source:    c.Imgref,
+		Name:      c.Imgref,
+		TLSVerify: &tlsVerify,
```
ondrejbudai, line 168:
```suggestion
		TLSVerify: common.ToPtr(c.TLSVerify),
```

bib/cmd/bootc-image-builder/image.go
```diff
@@ -154,6 +156,195 @@ func pipelinesForDiskImage(c *ManifestConfig, rng *rand.Rand) (image.ImageKind,
...
+			"rpcbind",
+			"selinux-policy-targeted",
+			"systemd",
+			"tar",
+			"xfsprogs",
+			"xz",
+		},
+	}
+
+	img.ExtraBasePackages = img.ExtraBasePackages.Append(rpmmd.PackageSet{
```
ondrejbudai, line 201: possibly stupid question: Why are the packages split into two groups? (happy to accept it, but I think that adding a comment with the reason would be helpful to future readers)

ondrejbudai (changes requested):
Looks very nice! Two things that I would like to see in this PR:

- document this in `README.md`
- add the new type to the `--type` description in the argument parser

cgwalters:
> Wait, will bootupd solve this also for this ISO? I don't expect bootupd to be used there.

I just mean used as far as image construction.  Agree it doesn’t make too much sense to have a concept of updating the bootloader of an ISO.  

Basically I’m just saying we fixed this papercut for containers and hence when we use containers for the ISO we get this fixed for free.

bib/cmd/bootc-image-builder/image.go
achilleas-k: Don't need `common.ToPtr()`, just `&c.TLSVerify` no?

bib/cmd/bootc-image-builder/image.go
achilleas-k: Uh, yeah, sorry, got a bit confused there.  This is the bootloader for the ISO itself and I imagine we'll keep doing grub install for that and will need to set the vendor (based on the distro).

bib/cmd/bootc-image-builder/image.go
cgwalters: > But using the user's payload would likely cause more issues than it would solve.

I would definitely say there are not-small ramifications; one obvious one to me is that the user would need to be aware that any custom systems units would run by default in the installer environment too by default, so they'd need to conditionalize...unless we did some work to have the Anaconda environment explicitly boot into a non-default system target.

But what issues specifically were you thinking of?

I know for sure though that the use case of a customized kernel is a real one for some important OCP customers for example, so if we *don't* use the customer base image for the ISO we'd need to in the general case expose some nontrivial subset of the customization functionality for the install environment too.  Maybe just kernel but there may in some cases be userspace tooling and things like "check in after install but before reboot" systemd units.

achilleas-k: > one obvious one to me is that the user would need to be aware that any custom systems units would run by default in the installer environment too by default, so they'd need to conditionalize...unless we did some work to have the Anaconda environment explicitly boot into a non-default system target.
> 
> But what issues specifically were you thinking of?

That's definitely one of them, potentially the most important one.  There's also packages.  A user package that might conflict with an Anaconda dependency, for example, or .  I just think that if we want our installer to inherit something from the payload container, we can figure out a way to take just the things we need (firmware, kernel modules, kernel args) rather than start from the container and chip away at it.

bib/cmd/bootc-image-builder/image.go
supakeen: Agreed.

bib/cmd/bootc-image-builder/image.go
supakeen: Because they were split in two in `images`, I've consolidated them.

mvo5:
I wonder if:
```python
diff --git a/test/test_smoke.py b/test/test_smoke.py
index 03e9fd8..0a22471 100644
--- a/test/test_smoke.py
+++ b/test/test_smoke.py
@@ -40,7 +40,7 @@ def build_container_fixture():
 
 
 # image types to test
-SUPPORTED_IMAGE_TYPES = ["qcow2", "ami"]
+SUPPORTED_IMAGE_TYPES = ["qcow2", "ami", "iso"]
 
 
 class ImageBuildResult(NamedTuple):
@@ -76,6 +76,7 @@ def image_type_fixture(tmpdir_factory, build_container, request):
     artifact = {
         "qcow2": pathlib.Path(output_path) / "qcow2/disk.qcow2",
         "ami": pathlib.Path(output_path) / "image/disk.raw",
+        "iso": pathlib.Path(output_path) / "image/install.iso",
     }
     assert len(artifact) == len(SUPPORTED_IMAGE_TYPES), \
         "please keep artifact mapping and supported images in sync"
```
works? might be enough to get a "free" simple test

supakeen:
@mvo5 thanks, let's try!

bib/cmd/bootc-image-builder/image.go
supakeen: Ok, I had this in the PR but had to drop it again as our dracut that mounts the squashfs is without lz4 support. I'll take a second look :)

supakeen: Right, I've now set it to `zstd` which is both faster and supported by dracut for the squashfs.

supakeen:
@mvo5 I've backed out the smoke test commit until we resolve the issues with the ISO rebooting and how to handle that in the test case(s).

README.md
```diff
@@ -258,6 +258,14 @@ sudo podman run \
...
 ```
 
+### Types
+
+Several image types can be output by `bootc-image-builder`:
+
+- `qcow2`, a disk image.
+- `iso`, an unattended Anaconda installer that will reformat and install to the first found disk.
+- `ami`, an Amazon EC2 image that can be directly uploaded.
+
```
achilleas-k, line 268: There's already an Image Types section above (L123).

bib/cmd/bootc-image-builder/image.go
```diff
@@ -168,6 +168,8 @@ func pipelinesForISO(c *ManifestConfig, rng *rand.Rand) (image.ImageKind, error)
 	}
 
 	img := image.NewAnacondaContainerInstaller(containerSource, ref)
+	img.SquashfsCompression = "zstd"
```
achilleas-k, line 171: This commit sets the compression to `zstd` gut also adds the `lz4` package to the ISO, which is a bit weird (leftover from trying out `lz4` I imagine).

achilleas-k (changes requested):

mvo5:
> @mvo5 I've backed out the smoke test commit until we resolve the issues with the ISO rebooting and how to handle that in the test case(s).

Yeah, this will probably need to be a followup. My initial (naive) idea was that we can just re-use the existing tests. But they rely on the fact that the blueprint adds the user to the image. This is not true for the installer image, there is no way to ssh to the installer image (only to the install*ed* image).

So the approach will be:
- use QMP in test/vm.py to get qemu events
- extend the QEMU to support a "cdrom" in it's constructor
- extend QEMU to open a qmp socket and add hook in QEMU class with stop_on="event=RESET"
- add start(no_wait_ssh)
- run QEMU("test-disk.img", cdrom="installer.iso", snapshot=False), wait for reboot, stop vm
- run QEMU("test-disk.img") and check that ssh works there

bib/cmd/bootc-image-builder/image.go
supakeen: Good catch, removed the `lz4` package from that commit.

bib/cmd/bootc-image-builder/image.go
achilleas-k: Since we're building the ISO from hardcoded repos anyway, let's keep this hardcoded for now as well.  Maybe add a comment that it will change once we get multi-distro support.

bib/cmd/bootc-image-builder/image.go
```diff
@@ -154,6 +156,195 @@ func pipelinesForDiskImage(c *ManifestConfig, rng *rand.Rand) (image.ImageKind,
 	return img, nil
 }
 
+func pipelinesForISO(c *ManifestConfig, rng *rand.Rand) (image.ImageKind, error) {
+	if c.Imgref == "" {
+		fail("pipeline: no base image defined")
+	}
+	ref := "ostree/1/1/0"
```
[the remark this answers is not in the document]
achilleas-k: Let's make it an empty string and add a comment that it's not needed and that the argument will be removed.

achilleas-k (approved):

PR #59, opened 2023-12-14
Upload to AWS and register AMI

achilleas-k:
Add new command for uploading a disk image to AWS and registering it as an AMI.
The command expects credentials from the environment and the keys / variable names match those that the AWS CLI uses.

I haven't made any changes to the Containerfile yet.  I'd like to get people's ideas on this.  Do we add it to the same container with a subcommand that calls this binary?  Do we make "upload aws" a subcommand of `bootc-image-builder`?  Either works and is easy to set up, but I'm not sure what we think is simpler for users.


lmilbaum:
I'm newbie to this project so maybe my questions are stupid.
Should there be a functional test added that runs in CI?
Is it expected to support the profile authentication at some point?
From an architectural point of view, would it make sense to separate the image building functionality from the cloud providers specific functionality? In my point of view, supporting the other major cloud vendors should be added in a similar way if this functionality is embedded in this project.

ondrejbudai:
> I'm newbie to this project so maybe my questions are stupid. Should there be a functional test added that runs in CI?

Yes!

> Is it expected to support the profile authentication at some point?

Yes!

> From an architectural point of view, would it make sense to separate the image building functionality from the cloud providers specific functionality?

I don't think so. There are three reasons:

1) If you build an image for AWS, there isn't much point in having it locally, because there's no local AWS. If you want to use the image locally, just build e.g. a qcow2 for QEMU. In other words, an AWS image is meant to be used in AWS, other uses are questionable.

2) Cloud images have metadata. E.g. AWS images need the information whether the image has UEFI support. Even though you can just inspect the bootable image just before you upload it, it's much more convenient to figure out the bootloader setup during the image build itself.

3) If we make the image builds blazingly fast, I don't think there's any need for separating these steps.

> In my point of view, supporting the other major cloud vendors should be added in a similar way if this functionality is embedded in this project.

Yup, AWS has just the highest priority.

ondrejbudai:
Do we want to tie building and uploading together? I think it makes a ton of sense, but I'm happy to be convinced otherwise.

achilleas-k:
> Do we want to tie building and uploading together? I think it makes a ton of sense, but I'm happy to be convinced otherwise.

I think there's good reasons for supporting both:
- Single step convenience: A single command that builds an image, pushes it to AWS and prints out your AMI ID, ready to go is really nice.
- Multi-step control: I want to build an image and maybe upload it to several accounts, or push it to test and then have the disk image to inspect locally.

In the latter case though we have the issue you mentioned with metadata.  In the single step case we know that what we're uploading is what we built, but in the second case it's not straightforward.

achilleas-k:
Converted to draft.  Pushed a commit that enables building and uploading with a single call, but I need to clean it up.
Need to clean it up but wanted to commit and push before stopping for the day.

achilleas-k:
Cleaned up, rebased, and marked READY.

The tool now supports building and uploading an AMI in a single step.  There's still a second `cmd` for uploading separately that shares the upload code.  Right now it's not built into the container, but we can integrate it as a separate (sub)command of the entrypoint.

ondrejbudai (changes requested):
image

The code looks good, but I really want this project to have a state-of-the-art documentation, so I'm requesting one (in README.md). :)

(Also, tests would be nice, but I think I value docs more than tests in this case. I probably don't want to block merging this on tests.)

achilleas-k:
> (Also, tests would be nice, but I think I value docs more than tests in this case. I probably don't want to block merging this on tests.)

Testing might need a little bit of work to set everything up with CI creds and cleanup.  We shouldn't let it slide though.  I'm okay with blocking here if you really want them in alongside the feature.

README.md
```diff
@@ -105,6 +105,18 @@ The following image types are currently available via the `--type` argument:
 | `ami`                 | [Amazon Machine Image](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/AMIs.html) |
 | `qcow2` **(default)** | [QEMU](https://www.qemu.org/)                                                         |
 
+### ☁️ AMI flags
```
ondrejbudai, line 108: I think we also need to document that you need to pass the secrets via environment flags. I wonder if it's worth having a new section for uploaders...

achilleas-k: Ah you're right.  Credentials were on my mind while I was writing it and it somehow slipped out by the end.

achilleas-k: Done.  Take a look :)

ondrejbudai (approved):
Splendid!

PR #61, opened 2023-12-15
Containerfile: Enable basic caching of go deps

ondrejbudai:
If we firstly copy go.mod and go.sum in and call go mod download, we get a cached layer with the go dependencies. This is useful for developers, because you only need to redownload the dependencies if go.mod or go.sum actually change.


achilleas-k (approved):
YES!

PR #63, opened 2023-12-16
Add doc and shell completion generation commands

bcl:
Use 'doc' to generate manpages, and 'completion' to generate shell completion scripts.

This changes <imgref> to IMGREF in the usage string because manpage won't output <imgref>

### DIFF
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -17,6 +17,7 @@ import (
 	"github.com/osbuild/images/pkg/rpmmd"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+	"github.com/spf13/cobra/doc"
 )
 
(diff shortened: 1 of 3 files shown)


ondrejbudai:
Thanks! Not sure if you know this, but we intend to ship this only as a container image, so having a man page might be a challenge, since containers generally lack the man tooling.

This makes me wonder how to properly do man pages (or offline documentation shipped with the tool itself in general) for container: One of my ideas was to bundle `README.md` inside the image and teach the container to show it using a pager (like `less(1)`). Basically run `podman run bootc-image-builder man` and it would give you a paginated `README.md`. Alternatively, we can tie this somehow to what `cobra` generates.

I'm happy to hear your thoughts.

bcl:
It might be useful to have them available even without man so that people exploring using (or extending) the container can generate the manpages and read them with cat :) Automated doc generation is always better than trying to keep a readme in sync.

PR #67, opened 2023-12-19
cmd/bootc-image-builder: support to use aws credential file and profiles

schuellerf:
Support to use aws credential file and profiles
implements HMS-3305

please comment especially on the part with 
```
		// SMELL we might want to implement a separate function
		//  like awscloud.newFromLocalProfile(awsProfile string, region string)
```
		
i guess the function would look similar to this:
```
// Create a new session from the locally stored AWS credentials file
// but given profile. NewDefault() above will expect "default" as profile or use environment variable AWS_PROFILE
func newFromLocalProfile(profile string, region string) (*AWS, error) {
	// Create a Session with a custom region and specified profile
	sess, err := session.NewSessionWithOptions(session.Options{
		Config: aws.Config{
			Region: aws.String(region),
		},
		Profile: profile,
	})
	if err != nil {
		return nil, err
	}

	return &AWS{
		uploader: s3manager.NewUploader(sess),
		ec2:      ec2.New(sess),
		s3:       s3.New(sess),
	}, nil
}
```

but would need to go to the images repository...

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ The following example builds a [Fedora ELN](https://docs.fedoraproject.org/en-US
 the command on.
 
 ```
-mkdir output
+mkdir -p output
 sudo podman run \
     --rm \
     -it \
@@ -107,7 +107,7 @@ The following image types are currently available via the `--type` argument:
 
 ## ☁️ Cloud uploaders
 
-### AMI
+### Amazon Machine Images (AMIs)
 
 AMIs can be automatically uploaded to AWS by specifying the following flags:
 
@@ -117,9 +117,39 @@ AMIs can be automatically uploaded to AWS by specifying the following flags:
 | --aws-bucket   | Target S3 bucket name for intermediate storage when creating AMI |
 | --aws-region   | Target region for AWS uploads                                    |
 
-*Note: These flags must all be specified together. If none are specified, the AMI is exported to the output directory.*
+*Notes:*
+- *These flags must all be specified together. If none are specified, the AMI is exported to the output directory.*
+- *The bucket must already exist in the selected region, bootc-image-builder will not create it if it is missing.*
+- *The output volume is not needed in this case. The image is uploaded to AWS and not exported.*
 
-AWS credentials must be specified through two environment variables:
+#### AWS credentials file
+If you already have a credentials file (usually in `$HOME/.aws/credentials`) you need to forward the
+directory to the container
+
+For example:
+ ```
+ $ sudo podman run \
+  --rm \
+  -it \
+  --privileged \
+  --pull=newer \
+  --security-opt label=type:unconfined_t \
+  -v $HOME/.aws:/root/.aws:ro \
+  quay.io/centos-bootc/bootc-image-builder:latest \
+  --type ami \
+  --aws-profile default \
+  --aws-ami-name fedora-bootc-ami \
+  --aws-bucket fedora-bootc-bucket \
+  --aws-region us-east-1 \
+  quay.io/centos-bootc/fedora-bootc:eln
+```
+
+Notes:
+ - *the argument `--aws-profile` is optional and is "default" if not specified*
+ - *instead of the `--aws-profile` you can also set the environment variable `AWS_PROFILE` like described below.* 
+
+#### AWS credentials via environment
+AWS credentials can be specified through two environment variables:
 | Variable name         | Description                                                                                                         |
 |-----------------------|---------------------------------------------------------------------------------------------------------------------|
 | AWS_ACCESS_KEY_ID     | AWS access key associated with an IAM account.                                                                      |
@@ -140,16 +170,12 @@ $ sudo podman run \
   --env-file=aws.secrets \
   --security-opt label=type:unconfined_t \
   quay.io/centos-bootc/bootc-image-builder:latest \
+  --type ami \
   --aws-ami-name fedora-bootc-ami \
   --aws-bucket fedora-bootc-bucket \
   --aws-region us-east-1 \
   quay.io/centos-bootc/fedora-bootc:eln
 ```
-
-*Notes:*
-- *The bucket must already exist in the selected region, bootc-image-builder will not create it if it is missing.*
-- *The output volume is not needed in this case. The image is uploaded to AWS and not exported.*
-
 
 ## 💽 Volumes
 
@@ -229,6 +255,15 @@ Example:
   ]
 }
 ```
+
+## Building
+
+To build the container locally you can run
+```shell
+sudo podman build --tag bootc-image-builder .
+```
+NOTE: running already the `podman build` as root avoids problems later as we need to run the building
+of the image as root anyway
 
 ## 📊 Project
 
--- a/bib/internal/uploader/aws.go
+++ b/bib/internal/uploader/aws.go
@@ -12,19 +12,51 @@ import (
 	"github.com/osbuild/images/pkg/cloud/awscloud"
 )
 
+func checkAWSCredentialFileExists() (bool, error) {
+	// SMELL make this path configurable?
+	credentialFile := "$HOME/.aws/credentials"
+
+	expandedFilename := os.ExpandEnv(credentialFile)
+	_, err := os.Stat(expandedFilename)
+	if err != nil {
+		return false, fmt.Errorf("AWS credential file not found: %s", expandedFilename)
+	}
+	return true, nil
+}
+
 // NewAWSClient returns an awscloud.AWS configured with a session for a given
 // region. It reads the credentials from environment variables:
 // AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY.
-func NewAWSClient(region string) (*awscloud.AWS, error) {
+// or tries to use "$HOME/.aws/credentials" otherwise
+func NewAWSClient(region string, awsProfile string) (*awscloud.AWS, error) {
+	foundCredentialFile := false
+
 	keyID := os.Getenv("AWS_ACCESS_KEY_ID")
 	if keyID == "" {
-		return nil, fmt.Errorf("AWS_ACCESS_KEY_ID environment variable is required")
+		var err error
+		foundCredentialFile, err = checkAWSCredentialFileExists()
+
+		if err != nil {
+			return nil, fmt.Errorf("AWS_ACCESS_KEY_ID environment variable or a valid '$HOME/.aws/credentials' is required")
+		}
 	}
-	secretKey := os.Getenv("AWS_SECRET_ACCESS_KEY")
-	if secretKey == "" {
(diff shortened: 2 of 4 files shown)


bib/internal/uploader/aws.go
```diff
@@ -12,19 +12,51 @@ import (
...
 )
 
+func checkAWSCredentialFileExists() (bool, error) {
+	// SMELL make this path configurable?
+	credentialFile := "$HOME/.aws/credentials"
+
+	expandedFilename := os.ExpandEnv(credentialFile)
+	_, err := os.Stat(expandedFilename)
+	if err != nil {
+		return false, fmt.Errorf("AWS credential file not found: %s", expandedFilename)
```
mvo5, line 22: Super nitpick: When writing the tests for this I noticed that the error is a bit misleading, this can result in `AWS credential file not found: stat /tmp/TestCheckAWSCredentialFileExistsNotExisting694989933/001/.aws/credentials: permission denied` so maybe something more like `cannot use aws file` as the prefix?

mvo5: I would love if we would test as much as possible by default. This function is a nice one, I attached a test with an example of what I have in mind (it's a bit rough, a table driven test or multiple ones are better :)

[0001-uploader-create-simple-unit-test.patch.txt](https://github.com/osbuild/bootc-image-builder/files/13724250/0001-uploader-create-simple-unit-test.patch.txt)

bib/internal/uploader/aws.go
```diff
@@ -12,19 +12,51 @@ import (
...
+	if foundCredentialFile {
+		// SMELL we might want to implement a separate function
+		//  like awscloud.newFromLocalProfile(awsProfile string, region string)
+		//  then we can remove this Setenv!
+		if  awsProfile != "default" {
+			os.Setenv("AWS_PROFILE", awsProfile)
+		}
+
+		return awscloud.NewDefault(region)
+	} else {
```
mvo5, line 53: (nitpick) Effective go [0] recommends to not use `else` if the `if` ends with a return. I like this pattern (a lot) because IMHO it reduce the complexity of the functions by having to worry about one less branch in my head :) So this would just become:
```go
...
	if foundCredentialFile {
		// SMELL we might want to implement a separate function
		//  like awscloud.newFromLocalProfile(awsProfile string, region string)
		//  then we can remove this Setenv!
		if awsProfile != "default" {
			os.Setenv("AWS_PROFILE", awsProfile)
		}

		return awscloud.NewDefault(region)
	}
	secretKey := os.Getenv("AWS_SECRET_ACCESS_KEY")
...
```



[0] https://go.dev/doc/effective_go#if

bib/internal/uploader/aws.go
```diff
@@ -12,19 +12,51 @@ import (
...
+	foundCredentialFile := false
+
 	keyID := os.Getenv("AWS_ACCESS_KEY_ID")
 	if keyID == "" {
-		return nil, fmt.Errorf("AWS_ACCESS_KEY_ID environment variable is required")
+		var err error
+		foundCredentialFile, err = checkAWSCredentialFileExists()
+
+		if err != nil {
+			return nil, fmt.Errorf("AWS_ACCESS_KEY_ID environment variable or a valid '$HOME/.aws/credentials' is required")
```
mvo5, line 40: By not wrapping the original error important information for the user gets lost, i.e. if it's a permission denied error that would be part of `err` and probably useful information for the user :)

bib/internal/uploader/aws.go
```diff
@@ -12,19 +12,51 @@ import (
...
+	}
+	return true, nil
+}
+
 // NewAWSClient returns an awscloud.AWS configured with a session for a given
 // region. It reads the credentials from environment variables:
 // AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY.
-func NewAWSClient(region string) (*awscloud.AWS, error) {
+// or tries to use "$HOME/.aws/credentials" otherwise
+func NewAWSClient(region string, awsProfile string) (*awscloud.AWS, error) {
```
mvo5, line 31: I find the current function a bit entangled and wonder if this could be made a bit easier to read (right now it's a bit hard for me to see what takes precendence for example). My gut feeling is that it might be easier if we would extract two helpers from the function and write this as, maybe something like this:
```go
func newAWSClientFromEnv(region string, awsProfile string) (*awscloud.AWS, error) {
	keyID := os.Getenv("AWS_ACCESS_KEY_ID")
	if keyID == "" {
		return nil, fmt.Errorf("AWS_ACCESS_KEY_ID must be set")
	}
	secretKey := os.Getenv("AWS_SECRET_ACCESS_KEY")
	if secretKey == "" {
		return nil, fmt.Errorf("AWS_SECRET_ACCESS_KEY must be set")
	}
	return awscloud.New(region, keyID, secretKey, "")
}

func newAWSClientFromCredsFile(region string, awsProfile string) (*awscloud.AWS, error) {
	foundCredentialFile, err := checkAWSCredentialFileExists()
	if err != nil {
		return nil, err
	}
	if !foundCredentialFile {
		return nil, fmt.Errorf("cannot find credentials file %v", credentialFile)
	}
	if awsProfile != "default" {
		os.Setenv("AWS_PROFILE", awsProfile)
	}
	return awscloud.NewDefault(region)
}

// NewAWSClient returns an awscloud.AWS configured with a session for a given
// region. It reads the credentials from environment variables:
// AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY.
// or tries to use "$HOME/.aws/credentials" otherwise
func NewAWSClient(region string, awsProfile string) (*awscloud.AWS, error) {
	client, errEnv := newAWSClientFromEnv(region, awsProfile)
	if errEnv == nil {
		return client, nil
	}
	client, errCredsFile := newAWSClientFromCredsFile(region, awsProfile)
	if errCredsFile == nil {
		return client, nil
	}

	return nil, fmt.Errorf("cannot usw AWS, neither env nor file method worked\n- env: %v\n- file: %v", errEnv, errCredsFile)
}
```
this way it's more linear and it's clear which of the two takes precedence. This assumes of course that I understand the existing flow which I might not have :) (if so, sorry! it's still early here).

And I would absolutely love to see a test for this too :)

mvo5 (commented):
Thanks for this PR! I especially like the care about updating the README.

A bit of a general comment - I noticed that you use your "main" branch for the PR. In general we prefer to have "feature" branches with names that reflect the feature. E.g. the "add-aws-credentials" or "feature/add-aws-credentials".

I put a lot of comments/ideas inhere - please don't hate me for putting those into this PR. I hope they are sensible and help to make things better. Feels free to ignore or ask about things you disagree with. I hope it's  useful!

bib/internal/uploader/aws.go
```diff
@@ -12,19 +12,51 @@ import (
 	"github.com/osbuild/images/pkg/cloud/awscloud"
 )
 
+func checkAWSCredentialFileExists() (bool, error) {
+	// SMELL make this path configurable?
+	credentialFile := "$HOME/.aws/credentials"
+
+	expandedFilename := os.ExpandEnv(credentialFile)
+	_, err := os.Stat(expandedFilename)
+	if err != nil {
```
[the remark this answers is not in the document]
schuellerf: looks very nice, but then I'll change the error to something like "AWS credential file stat failed: %w"

bib/internal/uploader/aws.go
ondrejbudai: I wonder if we cannot just use `awscloud.NewDefault()`. According to [the docs](https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html#specifying-credentials), it checks the env variables first, then it checks the credentials file, and then it even checks some more places (if IAM roles also work out of the box, that would be sweat).

README.md
```diff
@@ -230,6 +256,15 @@ Example:
 }
 ```
 
+## Building
```
ondrejbudai, line 259: We try to keep commits small and doing just one thing. One PR can consist of as many commits as required. In this case, adding a new section isn't directly connected to the main purpose of this commit (adding support for .aws/credentials), so I suggest extracting it into a separate commit.

README.md
```diff
@@ -24,7 +24,7 @@ The following example builds a [Fedora ELN](https://docs.fedoraproject.org/en-US
 the command on.
 
 ```
-mkdir output
+mkdir -p output
```
ondrejbudai, line 27: Don't hesitate to make this a separate commit as well. Or just group all minor changes to the README under one commit.

README.md
```diff
@@ -117,9 +117,39 @@ AMIs can be automatically uploaded to AWS by specifying the following flags:
 | --aws-bucket   | Target S3 bucket name for intermediate storage when creating AMI |
 | --aws-region   | Target region for AWS uploads                                    |
 
-*Note: These flags must all be specified together. If none are specified, the AMI is exported to the output directory.*
+*Notes:*
+- *These flags must all be specified together. If none are specified, the AMI is exported to the output directory.*
+- *The bucket must already exist in the selected region, bootc-image-builder will not create it if it is missing.*
+- *The output volume is not needed in this case. The image is uploaded to AWS and not exported.*
+
```
ondrejbudai, line 124: Is it worth adding a tiny summary of the auth methods here? Something like:

The AMI uploader supports two authentication methods:

1) AWS credentials file
2) AWS credentials via environment

With hyper-links? Or is that too much?

ondrejbudai (commented):
FTR, I've tried this locally and it worked wonderfully! :heart: 

Some more comments inline, but @mvo5 already did a great job reviewing. :)

bib/internal/uploader/aws.go
schuellerf: You are right - the main question for me is if we should provide commandline arguments and translate them to os.Setenv for convenience - as the approach with the env file does not seem to be very convenient (for all usecases)
I can rewrite this and check the compatibility for sure

README.md
schuellerf: I just left those comments as the majority of them were there - you are right keeping it short does make sense here

bib/internal/uploader/aws.go
ondrejbudai: For `AWS_ACCESS_KEY_ID` and the other one: They basically have to be `--env-file`, you don't want to pass secrets as `argv`.

For the credentials file: I guess the options are:
```
podman run \
  -v $HOME/.aws:/root/.aws:ro \
  --aws-profile default
```

vs.

```
podman run \
  -v $HOME/.aws:/root/.aws:ro \
  --env AWS_PROFILE=default
  quay.io/centos-bootc/bootc-image-builder:latest
```

(you don't have to use `--env-file`, you can just `--env` for one env variable)

I think I might actually prefer the latter one, because the options are closely together. Not a strong opinion, though. @mvo5, wdyt?

schuellerf:
I'll recreate this pullrequest (also to test the syntax for jira issues) (see #72)

PR #70, opened 2023-12-20
A few minor patches

cgwalters:
image: Hoist up hardcoded constant, add TODO

To make it more obvious we should fix this.

---

Remove extraneous newlines from errors

My go linter correctly warns about these; the error
printer will include a newline as needed.

---

image: Add a TODO for kargs

This will go away when we use bootc.

---


bcl (approved):
Looks good to me

PR #72, opened 2023-12-21
HMS-3305: add support for authentication profiles

schuellerf:
Just using awscloud.NewDefault() should cover all usecases of AWS authentication.

### DIFF
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -144,7 +144,7 @@ func build(cmd *cobra.Command, args []string) {
 	upload := false
 	if region, _ := cmd.Flags().GetString("aws-region"); region != "" {
 		if imgType != "ami" {
-			fail("aws flags set for non-ami image type")
+			fail(fmt.Sprintf("aws flags set for non-ami image type (type is set to %s)", imgType))
 		}
 		// initialise the client to check if the env vars exist before building the image
 		_, err := awscloud.NewDefault(region)
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -7,9 +7,9 @@ import (
 	"os"
 	"path/filepath"
 
-	"github.com/osbuild/bootc-image-builder/bib/internal/uploader"
 	"github.com/osbuild/images/pkg/arch"
 	"github.com/osbuild/images/pkg/blueprint"
+	"github.com/osbuild/images/pkg/cloud/awscloud"
 	"github.com/osbuild/images/pkg/container"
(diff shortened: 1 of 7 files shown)


bib/internal/uploader/aws.go
```diff
@@ -12,19 +11,11 @@ import (
...
-	keyID := os.Getenv("AWS_ACCESS_KEY_ID")
-	if keyID == "" {
-		return nil, fmt.Errorf("AWS_ACCESS_KEY_ID environment variable is required")
-	}
-	secretKey := os.Getenv("AWS_SECRET_ACCESS_KEY")
-	if secretKey == "" {
-		return nil, fmt.Errorf("AWS_SECRET_ACCESS_KEY environment variable is required")
-	}
-	return awscloud.New(region, keyID, secretKey, "")
+	return awscloud.NewDefault(region)
```
achilleas-k, line 18: I guess the function is useless now, we should just use `awscloud.NewDefault()` directly instead.

bib/cmd/bootc-image-builder/main.go
```diff
@@ -144,7 +144,7 @@ func build(cmd *cobra.Command, args []string) {
 	upload := false
 	if region, _ := cmd.Flags().GetString("aws-region"); region != "" {
 		if imgType != "ami" {
-			fail("aws flags set for non-ami image type")
+			fail(fmt.Sprintf("aws flags set for non-ami image type (type is set to %s)", imgType))
```
achilleas-k, line 147: Can you make this change in a separate commit so it's not mixed in with the other change?

Also the commit message subject line is a bit too long (should be < 72 chars).

I think you can add the Jira ticket ID in the message body and it will still work.

achilleas-k (dismissed):
Nice!  Thanks for this.  I have a nitpick or two but also we should probably drop that function that just calls a different one now.

ondrejbudai (approved):
Splendid! This is amazing, I can finally just use my credentials under `.aws`. :) I put one item to think about inline, but it's really just "food" for a potential follow-up.

One last style-related nitpick: Our preferred commit title format is: `<module>: <verb> <rest>`. The `verb` is usually something like, add, bump, remove, clean, make. This list is definitely non-exhaustive, and the rule isn't strict, but it's a nice convention to follow. We should probably put this in our developer guides. :thinking: 

For fun and giggles, I asked AI to rewrite yours, and the results are mixed. :hand_over_mouth: 

```
    bib: Cover all use cases using awscloud.NewDefault(). HMS-3305
    bib/cmd/bootc-image-builder: Provide more detailed error output. HMS-3305
    README: Utilize env vars for AWS. HMS-3305
    README: Update general documentation on building and testing.
```

Nothing to block on here, approving!

PR #73, opened 2023-12-21
update readme to include the `key` and add `--config` arg to first example

sallyom:
/cc @ondrejbudai @achilleas-k

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,19 +26,20 @@ the command on.
 The `fedora-bootc:eln` base image does not include a default user. This example injects a [user configuration file](#-build-config)
 by adding a volume-mount for the local file as well as the `--config` flag to the bootc-image-builder container.
 
-```
-# create output/config.json as described above
+The following command will create a QCOW2 disk image. First, create `./config.json` as described above to configure user access.
 
+```
 sudo podman run \
     --rm \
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -230,6 +254,29 @@ Example:
 }
 ```
 
+### Accessing the system
+
+With a virtual machine launched with the above [virt-install](#virt-install) example, access the system with
+
+```
+ssh -i /path/to/private/ssh-key foo@ip-address
```
achilleas-k, line 262: We have two example configs, one with username `foo` for the general build config example and one with `alice` for the user specific example.  I doubt anyone will be confused by this but maybe we should have just one username.

achilleas-k (approved):
Thanks so much!  Small comment below about the usernames in the examples but I think we can fix it later.

README.md
sallyom: let's go with alice! i like that better :) i pushed the update...

teg (dismissed):

achilleas-k (approved):
Thanks so much!!

achilleas-k:
Rebased and resolved the conflicts with `main`.

achilleas-k (approved):

PR #75, opened 2023-12-22
tests: run integration test on macos GH runners as well

mvo5:
This is a rebase/update of https://github.com/osbuild/bootc-image-builder/pull/33 without hacks.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -79,8 +79,9 @@ jobs:
 
   integration-macos:
     name: "Integration macos"
-    # needed to get latest cpu
-    runs-on: macos-13
+    # disabled GH runner as it takes ~50min to run this test
+    #runs-on: macos-13  # needed to get latest cpu
+    runs-on: self-hosted
     steps:
     - uses: actions/checkout@v4
       with:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -76,3 +76,25 @@ jobs:
         # XDG_RUNTIME_DIR is set.
         # TODO: figure out what exactly podman needs
         sudo -E XDG_RUNTIME_DIR= pytest-3 -s -vv
+
+  integration-macos:
+    name: "Integration macos"
+    # needed to get latest cpu
+    runs-on: macos-13
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
+    - name: Setup up python
+      uses: actions/setup-python@v5
+      with:
+        cache: 'pip'
+    - run: pip install -r test/requirements.txt
+    - name: Setup up podman
+      run: |
+        brew install podman netcat
+        podman machine init --rootful
+        podman machine start
+    - name: Run tests
+      run: |
+        pytest -rs -s -vv
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -79,7 +79,8 @@ jobs:
 
   integration-macos:
     name: "Integration macos"
-    # disabled GH runner as it takes ~50min to run this test
+    # disabled GH runner as it takes ~50min to run this test, self-hosted
+    # is much faster (~15min)
     #runs-on: macos-13  # needed to get latest cpu
     runs-on: self-hosted
     steps:
@@ -90,12 +91,16 @@ jobs:
       uses: actions/setup-python@v5
       with:
         cache: 'pip'
-    - run: pip install -r test/requirements.txt
+    - run: python3 -m pip install -r test/requirements.txt
     - name: Setup up podman
       run: |
         brew install podman netcat
-        podman machine init --rootful
-        podman machine start
+        if ! podman machine inspect; then
+            podman machine init --rootful
+        fi
+        if [ "$(podman machine inspect --format='{{.State}}')" != "running" ]; then
+            podman machine start
+        fi
     - name: Run tests
       run: |
         pytest -rs -s -vv
--- a/test/requirements.txt
+++ b/test/requirements.txt
@@ -1,3 +1,3 @@
-pytest
-paramiko
-flake8
+pytest==7.4.3
+flake8==6.1.0
+paramiko==2.12.0
--- a/test/test_smoke.py
+++ b/test/test_smoke.py
@@ -1,6 +1,7 @@
 import json
 import os
 import pathlib
+import platform
 import re
 import subprocess
 from typing import NamedTuple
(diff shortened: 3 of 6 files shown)


achilleas-k:
Instead of relying on the journal for logs and losing the checks on macOS, we could run the `podman run` command without `--rm` and then fetch the logs with `podman logs <container>`.
Worth it?

mvo5:
> Instead of relying on the journal for logs and losing the checks on macOS, we could run the `podman run` command without `--rm` and then fetch the logs with `podman logs <container>`. Worth it?

Love the suggestion - added now!

.github/workflows/tests.yml
```diff
@@ -76,3 +76,25 @@ jobs:
...
         sudo -E XDG_RUNTIME_DIR= pytest-3 -s -vv
+
+  integration-macos:
+    name: "Integration macos"
+    # needed to get latest cpu
+    runs-on: macos-13
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
```
lmilbaum, line 87: Is the checkout action uses this ref by default? Is that addition necessary?

.github/workflows/tests.yml
```diff
@@ -76,3 +76,25 @@ jobs:
         # XDG_RUNTIME_DIR is set.
         # TODO: figure out what exactly podman needs
         sudo -E XDG_RUNTIME_DIR= pytest-3 -s -vv
+
+  integration-macos:
+    name: "Integration macos"
+    # needed to get latest cpu
+    runs-on: macos-13
+    steps:
+    - uses: actions/checkout@v4
```
lmilbaum, line 85: I'd suggest using the full action version (vX.Y.Z) because there might be rare cases where action versions are not backward compatible.

.github/workflows/tests.yml
```diff
@@ -76,3 +76,25 @@ jobs:
...
+    - name: Setup up python
+      uses: actions/setup-python@v4
+    - name: Setup up podman
+      run: |
+        brew install podman netcat
+        podman machine init --rootful
+        podman machine start
+    - name: Install test dependencies
+      run: |
+        sudo pip install pytest flake8 paramiko
```
lmilbaum, line 97: Would you mind using a requirements.txt file to list the packages with their corresponding pinned versions?

.github/workflows/tests.yml
mvo5: Thank you, that is a nice catch. TBH I just cargo-culted this from the rest of the file. I will clean it up everywhere :)

mvo5: Actually, I looked at this more closely and there is a subtle difference. The default behavior is AFAICT that a merge commit of main and the PR is generated and checked out. With our current setup we only get the PR HEAD [1].

I have no strong opinion either way but would prefer to consult with the team first (as I'm relatively new and lack a lot of the history) what the rational for this decision is and if it needs changing I will do it in a separate PR. Hope that is okay with you :)

[1] https://github.com/actions/checkout#checkout-pull-request-head-commit-instead-of-merge-commit

.github/workflows/tests.yml
mvo5: Thanks for this suggestion. I would like to get input from @ondrejbudai or @achilleas-k because AFAICT we are not doing this in any of our other repos. I'm happy to check this but want it to be consistent (in this repo at least). It will be a separate PR that changes it for all checkout actions in the workflow. Hope that is okay?

.github/workflows/tests.yml
mvo5: Thanks, good point. Let me fix this.

.github/workflows/tests.yml
ondrejbudai: No strong opinion here, I generally like pinning deps, but I have never seen an issue with `actions/checkout` (and our use-case is trivial), so I don't care if this particular one is pinned.

.github/workflows/tests.yml
ondrejbudai: I think that one advantage is of using a PR head is that you clearly know what commit is being tested. But it's matter of taste, tbh.

I'm more interested if this works correctly when run on a merge queue, it looks a bit fishy to me tbh.

ondrejbudai:
This test takes 50 minutes to run on the Github-hosted worker. Should we try deploying a self-hosted runner on our AWS machine?

.github/workflows/tests.yml
achilleas-k: That's a good point, because on the merge queue, the `github.event.pull_request` should be empty, so I guess `ref` is unset then.
Comparing https://github.com/osbuild/bootc-image-builder/actions/runs/7384902395/job/20088639130#step:2:2 (checkout action for shellcheck on a PR) with https://github.com/osbuild/bootc-image-builder/actions/runs/7386384027/job/20092854868#step:2:2 (checkout action for shellcheck on the merge queue) seems to support this.

mvo5: If I read the room correctly here it sounds like we do want to move to testing the merge commit instead (at least on systems that use a merge queue)? Fwiw, I tend to agree but have no strong opinion either way. I do care that we do it consistently though :) Unless someone here disagrees I will prepare PRs to make this change consistently.

mvo5:
> This test takes 50 minutes to run on the Github-hosted worker. Should we try deploying a self-hosted runner on our AWS machine?

Great point  - I setup our AWS machine as a self hosted runner and it finisehd in 14min. so definitely worth it. Self hosting on macOS in AWS is a bit of a PITA though because of https://github.com/actions/runner/issues/947 and https://github.com/actions/runner/issues/1056 - I will see what I can do about that.

.github/workflows/tests.yml
mvo5: I will do this in a followup if that's okay? Would like to not block on it for this PR.

mvo5:
> > This test takes 50 minutes to run on the Github-hosted worker. Should we try deploying a self-hosted runner on our AWS machine?
> 
> Great point - I setup our AWS machine as a self hosted runner and it finisehd in 14min. so definitely worth it. Self hosting on macOS in AWS is a bit of a PITA though because of [actions/runner#947](https://github.com/actions/runner/issues/947) and [actions/runner#1056](https://github.com/actions/runner/issues/1056) - I will see what I can do about that.

This is ready now, sorry that it took a bit, the self hosted follows the setup from https://github.com/actions/runner/issues/1056#issuecomment-1237426462 and is now running a macOS background service.

achilleas-k (approved):

ondrejbudai (approved):
:apple: Go!

PR #82, opened 2024-01-01
test: use indirect `pytest.mark.parametrize` to build image

mvo5:
This commit tweaks the image tests to use indirect fixtures when creating the images. This allows to write something like:
```
sudo pytest-3 -k 'test_image_is_generated[ami]'
```
to select only specific images to test.

Note that this also contains a tiny fix (in it's own commit) for yet another netcat commandline flavor. I can create a separate PR for this if desired, it's not strictly needed for this PR.

### DIFF
--- a/test/test_smoke.py
+++ b/test/test_smoke.py
@@ -1,9 +1,9 @@
-import collections
 import json
 import os
 import pathlib
 import re
 import subprocess
+from typing import NamedTuple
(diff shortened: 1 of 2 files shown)


achilleas-k (approved):
Very cool.  LGTM.

That netcat thing looks like a real pain.

PR #87, opened 2024-01-05
test: add `BIB_TEST_{BUILD,BOOTC}_CONTAINER_TAG` to customize tests

mvo5:
There is a desire to resue our tests to also test when new bootc containers are generated that they still build and boot. To do this we need a way to customize the tests further so that: a) we can omit building our code and instead use a "stable" container
   that contains bootc-image-builder
b) we can specificy what bootc (target) image to test so that the
   bootc building team can point to their staging area

This is currently done via environment variables. It might be worth exploring the pytest commandline option support [0] but that can be done in a followup.

[0] https://docs.pytest.org/en/7.1.x/example/simple.html


achilleas-k (approved):
Thanks.  LGTM!

PR #89, opened 2024-01-05
Two refactoring patches, validation of entrypoint state

cgwalters:
Move build logic into `build` sub-command verb

In the future it will very likely be useful for us to have
other verbs.  For now, this is effectively an implementation
detail though because the entrypoint is changed to always pass `build`.

(Actually right now this fixes the problem that the `completion`
 verb was being masked)

---

Move `prepare.sh` into Go code

It's just ugly to have bits of this in shell script in this
way; it's not complex code and doing it in Go helps keep
the development environment simpler.

---

Move some helpers into modules

For functional clarity.

---

Validate that we're in rootful podman

As this is a footgun that multiple people have run into.

---

Validate that we're running privileged

On general principle.

---


cgwalters:
What I actually was going to fix is detecting `rootless=1` podman containers but I didn't want do that in shell script, so this is prep for having environment mutation/detection in Go, like how we do all similar stuff in bootc in Rust.

cgwalters:
Only lightly tested locally so far...doing some more

cgwalters:
OK now with some logic to validate that we're not running rootless, etc.

bib/internal/podmanutils/podmanutils.go
```diff
@@ -0,0 +1,38 @@
+package podmanutils
```
mvo5, line 1: (super nitick) in the go standard library the convention seems to be to use the singular form (ioutil, httputil). So maybe just `podmanutil`?

bib/internal/podmanutils/podmanutils.go
```diff
@@ -0,0 +1,38 @@
...
+const envPath = "/run/.containerenv"
+
+// rootlessKey is set when we are rootless
+const rootlessKey = "rootless=1"
+
+// IsRootless detects if we are running rootless in podman;
+// other situations (e.g. docker) will successfuly return false.
+func IsRootless() (bool, error) {
+	buf, err := os.ReadFile(envPath)
+	if err != nil {
```
mvo5, line 22: (niptick^3) we could maybe avoid the nested if and write this as
```go
	buf, err := os.ReadFile(envPath)
	if errors.Is(err, fs.ErrNotExist) {
		return false, nil
	}
	if err != nil {
		return false, err
	}
...
```
but maybe I'm totally overthinking it :)

bib/internal/setup/setup.go
```diff
@@ -0,0 +1,98 @@
...
+
+	if err := os.MkdirAll(runTmp, 0o755); err != nil {
+		return err
+	}
+	if !utils.IsMountpoint(runTmp) {
+		if err := exec.Command("mount", "-t", "tmpfs", "tmpfs", runTmp).Run(); err != nil {
+			return fmt.Errorf("failed to mount tmpfs to %s: %w", runTmp, err)
+		}
+	}
+	src, err := os.Open("/usr/bin/osbuild")
```
mvo5, line 46: (nipitkc) I wonder if just `exec.Command("cp",...` might be easier here as it's shorter ?

bib/internal/setup/setup.go
```diff
@@ -0,0 +1,98 @@
...
+	installType := "system_u:object_r:install_exec_t:s0"
+	// Where we dump temporary files; this must be an overlayfs as we cannot
+	// write security contexts on overlayfs.
+	runTmp := "/run/osbuild/"
+
+	if err := os.MkdirAll(runTmp, 0o755); err != nil {
+		return err
+	}
+	if !utils.IsMountpoint(runTmp) {
+		if err := exec.Command("mount", "-t", "tmpfs", "tmpfs", runTmp).Run(); err != nil {
```
mvo5, line 42: I would prefer if we could use `CombindOutput()` here and display that in the error case, something like:
```go
+               if output, err := exec.Command("mount", "-t", "tmpfs", "tmpfs", runTmp).CombinedOutput(); err != nil {
+                       return fmt.Errorf("failed to mount tmpfs to %s: %w (output: %q)", runTmp, err, output)
                }
        }
```
(and the same in the other usage of `.Run()` below

bib/internal/setup/setup.go
```diff
@@ -0,0 +1,98 @@
...
+func Validate() error {
+	isRootless, err := podmanutils.IsRootless()
+	if err != nil {
+		return fmt.Errorf("checking rootless: %w", err)
+	}
+	if isRootless {
+		return fmt.Errorf("this command must be run in rootful (not rootless) podman")
+	}
+
+	// Having /sys be writable is an easy to check proxy for privileges; more effective
```
mvo5, line 87: This is super nice!

bib/internal/utils/utils.go
```diff
@@ -0,0 +1,11 @@
+package utils
+
+import "os/exec"
+
+// IsMountpoint checks if the target path is a mount point
+func IsMountpoint(path string) bool {
+	c := exec.Command("mountpoint", path)
+	c.Stderr = nil
```
mvo5, line 8: (nitpick) is this necessary? Iirc that is the default [0] - or am I missing something?


[0] https://github.com/golang/go/blob/master/src/os/exec/exec.go#L376

mvo5 (approved):
Thank you, this is super nice! I have some suggestions/nitpicks inline but no blockers. Feel free to ignore but maybe something in there is useful.

bib/internal/podmanutils/podmanutils.go
cgwalters: I've usually nested them...not sure if there's a truly standard style here but glancing at e.g. podman including some vendors I mostly see nesting for this.

bib/internal/setup/setup.go
cgwalters: Yeah fair, what I've ended up with in most other projects is wrapper functions for `exec`...can add that too

bib/internal/utils/utils.go
cgwalters: You're right; fixed

bib/internal/setup/setup.go
cgwalters: Added and using an exec wrapper now

cgwalters:
OK resolved all the nits except one

bib/internal/utils/utils.go
```diff
@@ -9,3 +16,17 @@ func IsMountpoint(path string) bool {
...
+// Synchronously invoke a command, writing its stdout to our stdout,
+// and gathering stderr into a buffer which will be returned in err
+// in case of error.
+func RunCmdSync(cmdName string, args ...string) error {
+	logrus.Debugf("Running: %s %s", cmdName, strings.Join(args, " "))
+	cmd := exec.Command(cmdName, args...)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("error running %s %s: %s: %w", cmdName, strings.Join(args, " "), err)
```
achilleas-k, line 26: The docstring mentions gathering stderr into a buffer but the code isn't doing it.  And there's a missing arg in the errorf which I guess should have been the buffered stderr?

achilleas-k (dismissed):

bib/internal/utils/utils.go
cgwalters: Thanks, fixed.  I had this locally but had forgotten to push apparently.
(Bigger picture juggling working on this and some other things, trying to use a Mac as a devenv, but was concurrently working on bugs like https://github.com/containers/podman/issues/21160 and have been concurrently destroying/recreating my podman machine.  And actually before this patch *and* before https://github.com/osbuild/bootc-image-builder/commit/c679e847d9d3b9334b3e4abfc18b8f6b1cfb7685#diff-6fd0e644fac1627088d13a0bdd084daee6f0ce1cca74d7343cdb0ffe0d87969cR91, bib happened to build for Mac)

bib/internal/podmanutils/podmanutils.go
mvo5: Yeah, totally fine, sorry for being too nitpicky

bib/internal/setup/setup.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/osbuild/bootc-image-builder/bib/internal/podmanutil"
+	"github.com/osbuild/bootc-image-builder/bib/internal/utils"
+	"golang.org/x/sys/unix"
+)
+
+// EnsureEnvironment mutates external filesystem state as necessary
+// to run in a container environment.  This function is idempotent.
+func EnsureEnvironment() error {
+	osbuildPath := "/usr/bin/osbuild"
+	if utils.IsMountpoint(osbuildPath) {
```
mvo5, line 17: (nitpick^3) I feel a bit bad for mentioning it and ignoring is fine - `utils` vs `util` (plural/singular) similar to podmanutil here :) Feel free to ignore though if it's too much nitpicking (

mvo5 (dismissed):
This is great, just one tiny comment/nitpick.

bib/internal/setup/setup.go
cgwalters: OK I renamed

achilleas-k:
Clicked the rebase button to get the `apt update` fix.

cgwalters:
@achilleas-k I think to merge you need to change your review state to approval

achilleas-k (approved):

PR #92, opened 2024-01-08
README.md: add tl;dr summary

mvo5:
We got some feedback that the summary in the README is a bit vague. This is an attempt to make it more concrete but also to keep it still short.


cgwalters (dismissed):
optional: while we have the patient open, hyperlink to boot?

ondrejbudai (approved):

PR #96, opened 2024-01-10
main: check credentials by listing regions HMS-3329

schuellerf:
Listing regions *early* in main.go is just a way to check if the credentials somehow work and fail/exit otherwise.

The upload is not guaranteed to work like this but might be a good indicator.


schuellerf:
For a more detailed "upload permission check" I would probably need to change `github.com/osbuild/images/pkg/cloud/awscloud` but my PR here might be sufficient to keep the changes small

schuellerf:
I tried to use the GitHub web interface to "resolve" but I can also rebase properly :-|

ondrejbudai (approved):
Thanks!

Issue #99, opened 2024-01-10
ability to create raw images for vfkit


cdrage:
I'd like to be able to create a raw image for me to use with vfkit.

Right now I have to do the following:

```
#!/bin/sh

set -exu

BOOTC_IMG_PATH=$1
qemu-img convert -f qcow2 -O raw ${BOOTC_IMG_PATH} bootc-overlay.img

#cp -c bootc-overlay.img snapshopt.img

./out/vfkit --cpus 2 --memory 2048 \
    --bootloader efi,variable-store=./efi-variable-store,create \
    --device virtio-blk,path=bootc-overlay.img \
    --device virtio-serial,stdio \
    --device virtio-net,nat,mac=72:20:43:d4:38:62 \
    --device virtio-rng \
    --device virtio-input,keyboard \
    --device virtio-input,pointing \
    --device virtio-gpu,width=1920,height=1080 \
    --gui
 ```
 
 It would be nice to be able to run the raw image directly rather than having to convert the qcow2 after.

cdrage:
Thanks to @cfergeau for the script by the way!

mvo5:
I opened https://github.com/osbuild/bootc-image-builder/pull/100 - let's see.

cdrage:
Closed via https://github.com/osbuild/bootc-image-builder/pull/100

PR #100, opened 2024-01-10
bib: add "raw" image format

mvo5:
A super simple PR to allow to get "raw" output for https://github.com/osbuild/bootc-image-builder/issues/99 

Right now this will be identical to an "ami" but it seems cleaner to me to me to separate the two as the may(?) diverge in the future.

### DIFF
--- a/bib/cmd/bootc-image-builder/image.go
+++ b/bib/cmd/bootc-image-builder/image.go
@@ -51,9 +51,7 @@ func Manifest(c *ManifestConfig) (*manifest.Manifest, error) {
 	var err error
 
 	switch c.ImgType {
-	case "qcow2":
-		fallthrough
-	case "ami":
+	case "ami", "qcow2", "raw":
 		img, err = pipelinesForDiskImage(c, rng)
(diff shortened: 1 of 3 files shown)


cgwalters (dismissed):
Looks sane to me; I would hope in the future changes like this also would change some documentation about the available images types.  But not a blocker.

cdrage:
Thank you for adding this!

achilleas-k (approved):
We could add it to the README explaining that it is (currently) identical to the AMI but doesn't support upload options.  Might want to add a note that the two could diverge in the future, though I'm always worried this might _train_ people to think of them as equivalent and then be hard to _untrain_ them of that idea.

Anyway.  LGTM!

cdrage:
After this is merged, is this automatically released to quay.io / built & pushed?

mvo5:
> After this is merged, is this automatically released to quay.io / built & pushed?

Yes, once merged it should get auto released.

achilleas-k (approved):
APPROVED

PR #101, opened 2024-01-10
ci: Ensure we `apt update` before `apt install`

cgwalters:
Otherwise we can operate on cached metadata and fail.


mvo5 (approved):
Thank you

Issue #18, opened 2023-11-28
using bootc install-to-filesystem


cgwalters:
This relates to https://github.com/osbuild/osbuild-deploy-container/issues/4

- We had some general agreement to support `bootc install-to-filesystem`; this will help long term with things like https://github.com/containers/bootc/issues/128
- `bootc install-to-filesystem` should also grow support for being provided the base container image  externally (e.g. cached in osbuild); we know this is needed for offline ISO installs too.  This ties with the above for the lifecycle bound app/infra containers
- We can't drop the osbuild/ostree stages because not every case will use bootc in the near future
- Agreement that for the ISO/installer case any customization (embedded kickstarts, but also *which* installer) would likely live external to the container (blueprint or equivalent)

cgwalters:
> bootc install-to-filesystem should also grow support for being provided the base container image externally 

Digging in, this is messier than I thought.  Still possible, but @ondrejbudai can you state more precisely the concern you had with having bootc install from the running container?

ISTM that in general going forward we'll want to support running images cached in the infrastructure, which will drive us towards using `containers-storage` most likely, as opposed to e.g. the `dir` transport.  And if we do that, ISTM it's just simpler to keep bootc doing exactly what it's doing today in fetching from the underlying store as opposed to having something else push content in, right?

achilleas-k:
Just to clarify, because there are two ideas here that sound very similar but are probably unrelated:
- The issue you're talking about is with the idea of having `bootc` run from outside the base container image when running `install-to-filesystem`.  So the idea of having it do `bootc install-to-filesystem <container ref> <filesystem path>` (or `bootc install-to-filesystem oci-archive:/path/to/container.tar /path/to/tree` for example), from a host machine would require too much work.
    - This is as opposed to running `podman run -v/path/to/tree:/target <container ref> bootc install-to-filesystem /target`, from the host, which is how it currently works.
- This issue is not related to running `bootc` from a container that is in an "offline" storage format like an archive, right?  So we can still do `podman run -v/path/to/tree:/target oci-archive:/path/to/container.tar bootc install-to-filesystem /target`?  Which will _probably_ work fine in osbuild.  My concern, as we discussed yesterday, is that we're putting a few too many layers of containers/namespaces here which make it hard to predict some details, but might be okay.  I think it's time we actually tried this and see what we get.

[...]

PR #106, opened 2024-01-11
bib: Add root filesystem selection (HMS-3110)

bcl:
Add `--rootfs` to select the root filesystem type (eg. ext4, xfs). It defaults to ext4.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,7 @@ Flags:
 | **--config** | Path to a [build config](#-build-config)                         |       ❌      |
 | --tls-verify | Require HTTPS and verify certificates when contacting registries |    `true`     |
 | **--type**   | [Image type](#-image-types) to build                             |    `qcow2`    |
+| --rootfs     | Root filesystem type to create. eg. ext4, xfs                    |    `ext4`     |
 
 *💡 Tip: Flags in **bold** are the most important ones.*
 
--- a/bib/cmd/bootc-image-builder/image.go
+++ b/bib/cmd/bootc-image-builder/image.go
@@ -31,6 +31,11 @@ type ManifestConfig struct {
 	// TODO: Make this an enum.
 	ImgType string
 
+	// Root filesystem type (eg. ext4, xfs. empty defaults to ext4)
+	//
+	// TODO: Make this an enum driven by images/pkg/disk
+	RootFSType string
+
 	// Build config
 	Config *BuildConfig
 
@@ -141,7 +146,7 @@ func pipelinesForDiskImage(c *ManifestConfig, rng *rand.Rand) (image.ImageKind, 
 
 	img.Workload = &NullWorkload{}
 
-	basept, ok := partitionTables[c.Architecture.String()]
+	basept, ok := partitionTables(c.RootFSType)[c.Architecture.String()]
 	if !ok {
 		fail(fmt.Sprintf("pipelines: no partition tables defined for %s", c.Architecture))
 	}
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -169,6 +169,7 @@ func build(cmd *cobra.Command, args []string) {
 	rpmCacheRoot, _ := cmd.Flags().GetString("rpmmd")
 	configFile, _ := cmd.Flags().GetString("config")
 	imgType, _ := cmd.Flags().GetString("type")
+	rootFSType, _ := cmd.Flags().GetString("rootfs")
 	tlsVerify, _ := cmd.Flags().GetBool("tls-verify")
 
 	if err := os.MkdirAll(outputDir, 0777); err != nil {
@@ -209,6 +210,7 @@ func build(cmd *cobra.Command, args []string) {
 	manifestConfig := &ManifestConfig{
 		Imgref:       imgref,
 		ImgType:      imgType,
+		RootFSType:   rootFSType,
 		Config:       &config,
 		Repos:        repos,
 		Architecture: hostArch,
@@ -267,6 +269,7 @@ func main() {
 	buildCmd.Flags().String("rpmmd", "/var/cache/osbuild/rpmmd", "rpm metadata cache directory")
 	buildCmd.Flags().String("config", "", "build config file")
 	buildCmd.Flags().String("type", "qcow2", "image type to build [qcow2, ami]")
+	buildCmd.Flags().String("rootfs", "ext4", "Root filesystem type [ext4, xfs]")
 	buildCmd.Flags().Bool("tls-verify", true, "require HTTPS and verify certificates when contacting registries")
 	buildCmd.Flags().String("aws-region", "", "target region for AWS uploads (only for type=ami)")
 	buildCmd.Flags().String("aws-bucket", "", "target S3 bucket name for intermediate storage when creating AMI (only for type=ami)")
--- a/bib/cmd/bootc-image-builder/partition_tables.go
+++ b/bib/cmd/bootc-image-builder/partition_tables.go
@@ -11,101 +11,108 @@ const (
 	GibiByte = 1024 * 1024 * 1024 // GiB
 )
 
-var partitionTables = distro.BasePartitionTableMap{
-	arch.ARCH_X86_64.String(): disk.PartitionTable{
-		UUID: "D209C89E-EA5E-4FBD-B161-B461CCE297E0",
-		Type: "gpt",
-		Partitions: []disk.Partition{
-			{
-				Size:     1 * MebiByte,
-				Bootable: true,
-				Type:     disk.BIOSBootPartitionGUID,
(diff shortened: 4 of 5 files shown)


bcl:
I think this is complete, but when I try running things with a locally built container I am getting errors from seemingly unrelated things. When I look at the generated manifest it looks ok, with xfs instead of ext4 for the `/` filesystem type.

```
mount/ostree-ostree/1/1/0 (org.osbuild.ostree.deployment): Deployment root at 'ostree/deploy/default/deploy/719b26bb969e44e8f88a5874c8a519d67ed422a6c0ad06b69991fda3177ff620.0'
[sss_cache] [confdb_init] (0x0010): Unable to open config database [/var/lib/sss/db/config.ldb]          
Could not open available domains                                                                         
[sss_cache] [confdb_init] (0x0010): Unable to open config database [/var/lib/sss/db/config.ldb]          
Could not open available domains                                                                         
  File "/run/osbuild/bin/org.osbuild.users", line 199, in <module>                                       
    r = main(args["tree"], args["options"])                                                              
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                              
  File "/run/osbuild/bin/org.osbuild.users", line 182, in main                                           
    ensure_homedir(tree, name, home)                                                                     
  File "/run/osbuild/bin/org.osbuild.users", line 157, in ensure_homedir                                 
    subprocess.run(["chroot", root, "mkhomedir_helper", name], check=True)                               
subprocess.CalledProcessError: Command '['chroot', '/run/osbuild/tree', 'mkhomedir_helper', 'root']' returned non-zero exit status 6.
```

mvo5:
Thanks, this looks very nice (and exciting!). 

I wonder if you also see these unexpected errors if you add the following:
```python
diff --git a/test/test_smoke.py b/test/test_smoke.py
index 03e9fd8..497629e 100644
--- a/test/test_smoke.py
+++ b/test/test_smoke.py
@@ -40,7 +40,7 @@ def build_container_fixture():
 
 
 # image types to test
-SUPPORTED_IMAGE_TYPES = ["qcow2", "ami"]
+SUPPORTED_IMAGE_TYPES = ["qcow2,xfs", "qcow2,btrfs", "ami,ext4"]
 
 
 class ImageBuildResult(NamedTuple):
@@ -48,6 +48,7 @@ class ImageBuildResult(NamedTuple):
     username: str
     password: str
     journal_output: str
+    rootfs: str
 
 
 @pytest.fixture(name="image_type", scope="session")
@@ -64,7 +65,7 @@ def image_type_fixture(tmpdir_factory, build_container, request):
     )
 
     # image_type is passed via special pytest parameter fixture
-    image_type = request.param
+    image_type, rootfs = request.param.split(",")
 
     username = "test"
     password = "password"
@@ -116,11 +117,12 @@ def image_type_fixture(tmpdir_factory, build_container, request):
         container_to_build_ref,
         "--config", "/output/config.json",
         "--type", image_type,
+        "--rootfs", rootfs,
     ])
     journal_output = testutil.journal_after_cursor(cursor)
     journal_log_path.write_text(journal_output, encoding="utf8")
 
-    return ImageBuildResult(generated_img, username, password, journal_output)
+    return ImageBuildResult(generated_img, username, password, journal_output, rootfs)
 
 
 def test_container_builds(build_container):
```
and then run:
```
$ sudo pytest -s -vv 'test/test_smoke.py::test_image_boots[qcow2,xfs]'
```
this seems to work for me. Adding code that the fs is correct could be added via:
```python
diff --git a/test/test_smoke.py b/test/test_smoke.py
index 03e9fd8..2c48cfe 100644
--- a/test/test_smoke.py
+++ b/test/test_smoke.py
@@ -144,6 +146,9 @@ def test_image_boots(image_type):
         exit_status, output = test_vm.run("echo hello", user=image_type.username, password=image_type.password)
         assert exit_status == 0
         assert "hello" in output
+        exit_status, output = test_vm.run("findmnt -no FSTYPE -T /", user=image_type.username, password=image_type.password)
+        assert exit_status == 0
+        assert output.strip() == image_type.rootfs
 
 
 def log_has_osbuild_selinux_denials(log):
```
(on top of the above). And that also works for "xfs" which is really nice.

What is interesting is that xfs seems to be fine but btrfs fails with:
```
...
Generating manifest-qcow2.json ... No match for argument: btrfs-progs
error install_specs
MarkingErrors: Error occurred when marking packages for installation: Problems in request:
missing packages: btrfs-progs
DNF error occurred: MarkingErrors: Error occurred when marking packages for installation: Problems in request:
missing packages: btrfs-progs
ERROR
```
I add the full diff that I used for completeness 
[0001-test-add-support-to-test-images-with-configurable-ro.patch.txt](https://github.com/osbuild/bootc-image-builder/files/13901752/0001-test-add-support-to-test-images-with-configurable-ro.patch.txt)

cgwalters:
> Add --rootfs to select the root filesystem type (eg. ext4, xfs). It defaults to ext4.

Note this overlaps with https://github.com/containers/bootc/blob/main/docs/install.md#operating-system-install-configuration-required

I'd really like to see us fix https://github.com/osbuild/bootc-image-builder/issues/18 because we would then already have filesystem selection instead of re-implementing it.

cgwalters (dismissed):
Per above I think we should target bootc functionality.

bcl:
> Per above I think we should target bootc functionality.

sure, but this is my effort to knock out some low hanging fruit, and to get more familiar with this codebase. So until there is actual bootc work being done I don't see any reason to block this. It's pretty simple.

bcl:
> I wonder if you also see these unexpected errors if you add the following:

Actually no, that seems to work fine. I'm seeing them with my own run script and config.json, so I may be doing something wrong with them...

cgwalters:
> sure, but this is my effort to knock out some low hanging fruit, and to get more familiar with this codebase. 

But the high level goal of the effort branded "bootc" is to make container images the technical heart of things.  This project is now part of that.  Of course, I think everyone is very well aware that we are reusing a giant amount of prior existing code (for example, Anaconda) because it just makes sense.

We have already set a precedent for external-to-container configuration being provided in this project - and that was agreed because as a short term expedience thing managing credentials and secrets is something one may often want to ship outside of the container image (disk images can be managed externally, i.e. secrets may just be one-time provisioning state, as users provided are).

At the same time, we should also aim to have the focus be on containers as where the operating system is configured.  And in *this specific* case, we already have functionality in bootc to do this.  Adding more things like this sets increasing precedent for managing OS state and source of truth outside of containers, which is misaligned with the overall goal.

rhatdan:
If bootc can do this now, why not use it?

cgwalters:
I'd also emphasize that I do feel to some degree like the `bootc install` is still subject to change - and everyone working on this if you have opinions on how it could do things differently please do speak up!  You should all feel as if you have a say in its future.

cgwalters:
This all said, it should absolutely be supported to override the default OS filesystem as a downstream user of an externally-produced base image.

The classic situation here is wanting to e.g. use xfs/bcachefs/etc as a Fedora Silverblue user or whatever - decoupling the ownership of base image from filesystem choice.

Our current primary target for bootc though is where user-owns-OS: the expected entrypoint to the system is deriving from our base image.  And now today, we have an overlap between:

- configuration via container
- configuration via image

That said, I still do also want to enable "generic" image use even in the bootc sphere.  This is one reason we have https://github.com/CentOS/centos-bootc-layered/tree/main/cloud which basically acts as a very compatible low-cost entrypoint to the system - you can boot it and get a root shell via ssh etc. that works the same as many distribution "generic cloud" images *without* building your own container.

In the same vein as this PR, we already have explicit support for downstream filesystem customization via kickstart.

So I guess, I will just lift my changes-requested.

But at the same time, I hope we can slowly start to reconcile these things (kickstart, "bib config", blueprints, and finally bootc install configs).

README.md
```diff
@@ -117,6 +117,7 @@ Flags:
 | **--config** | Path to a [build config](#-build-config)                         |       ❌      |
 | --tls-verify | Require HTTPS and verify certificates when contacting registries |    `true`     |
 | **--type**   | [Image type](#-image-types) to build                             |    `qcow2`    |
+| --rootfs     | Root filesystem type to create. eg. ext4, xfs                    |    `ext4`     |
```
cgwalters, line 120: Can you at least just document here that this functionality will also be accessible in the future via `bootc install` configs?

bcl: Isn't that getting ahead of ourselves though? I think it's confusing to refer to future things that don't work right now.

bcl:
Thanks for the additional comments. Where I'm getting lost is the mention of all the other possible configuration options (most of which aren't yet available?). eg. how does Anaconda and Kickstart fit into `bib` for example? AFAIK this tool is making ami or qcow2 images for booting on bare metal so there is no kickstart.

I agree that it's better to have one config than a pile of them :) I'm just not sure what that is yet (and would lean towards blueprints which are already part of the `bib` configuration).

PR #109, opened 2024-01-11
test: add container installer test

mvo5:
Add a new integration test that checks that the container installer
is working. The contaner installer just does an unattended install
of a disk. The test will run qemu with the installer.iso and an
empty disk. Once a reboot from the ISO is detected (via QMP) qemu
exists and boots the test disk and checks that it boots and the
test user can login.

### DIFF
--- a/test/vm.py
+++ b/test/vm.py
@@ -1,7 +1,9 @@
 import abc
+import os
 import pathlib
 import subprocess
 import sys
+import time
 import uuid
 from io import StringIO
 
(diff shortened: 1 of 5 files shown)


test/test_smoke.py
```diff
@@ -179,3 +179,20 @@ def test_image_build_without_se_linux_denials(image_type):
     assert image_type.journal_output != ""
     assert not log_has_osbuild_selinux_denials(image_type.journal_output), \
         f"denials in log {image_type.journal_output}"
+
+
+@pytest.mark.parametrize("image_type", ["iso"], indirect=["image_type"])
+def test_iso_installs(image_type):
+    installer_iso_path = image_type.img_path
+    test_disk_path = installer_iso_path.with_name("test-disk.img")
+    with open (test_disk_path, "w") as fp:
```
mvo5, line 188: This should probably move into the VM class as "create_disk" or something

test/vm.py
```diff
@@ -38,23 +48,55 @@ def start(self):
...
+                self.wait_qmp_event(qmp_event)
+                self._log(f"qmp event {qmp_event}")
+            case _:
+                raise ValueError(f"unsupported wait_event {wait_event}")
 
     def _log(self, msg):
         # XXX: use a proper logger
         sys.stdout.write(msg.rstrip("\n") + "\n")
 
+    def wait_qmp_event(self, qmp_event):
```
mvo5, line 83: Add wait_event_timeout here too (not just for the socket)?

mvo5:
The testing farm tests fail with "no space left on device" :/

test/vm.py
```diff
@@ -107,18 +125,52 @@ def start(self):
...
         self._qemu_p = subprocess.Popen(
-            qemu_cmdline, stdout=sys.stdout, stderr=sys.stderr)
+            qemu_cmdline,
+            stdout=sys.stdout,
+            stderr=sys.stderr,
+        )
         # XXX: also check that qemu is working and did not crash
-        self.wait_ssh_ready()
-        self._log(f"vm ready at port {self._ssh_port}")
+        match wait_event.split(":"):
```
supakeen, line 146: Do we need to run these tests downstream as well at some point? The `match`-statement was added to a quite recent Python.

lmilbaum: As a general note, I would answer - yes.

mvo5: Thank you, indeed `match` is only available in python 3.10 and centos9  has 3.9 and centos8 has 3.6 so I remove this.

mvo5: This is done now

bcl:
It looks like `SUPPORTED_IMAGE_TYPES` is no longer the list that drives all the test loops. I was a bit confused by the artifact test change until I realized that you have a separate test for `iso` that isn't in the list. Maybe it needs a new name, or a comment?

mvo5:
> It looks like `SUPPORTED_IMAGE_TYPES` is no longer the list that drives all the test loops. I was a bit confused by the artifact test change until I realized that you have a separate test for `iso` that isn't in the list. Maybe it needs a new name, or a comment?

Thanks, that is a nice catch and a good point. I renamed the vars a bit to (hopefully) make it more clear. Ideas/comments welcome!

bcl (dismissed):
Looks good to me, thanks!

mvo5:
It looks like the test is currently failing because of an anaconda error?
Screenshot from 2024-01-25 10-28-39

achilleas-k (approved):
Splendid.  LGTM!

PR #120, opened 2024-01-15
main: enable containers-storage (HMS-3235)

kingsleyzissou:
Enable the ability to use local containers from containers-storage rather than pulling containers from a remote registry.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -23,4 +23,3 @@ VOLUME /output
 WORKDIR /output
 VOLUME /store
 VOLUME /rpmmd
-
--- a/README.md
+++ b/README.md
@@ -46,6 +46,33 @@ sudo podman run \
     quay.io/centos-bootc/fedora-bootc:eln
 ```
 
+### Using local containers
+
+To use containers from local container's storage rather than a registry, we need to ensure two things:
+- the container exists in local storage
+- mount the local container storage
+
+Since the container is run in `rootful` only root container storage paths are allowed.
+
+```bash
+sudo podman run \
+    --rm \
+    -it \
+    --privileged \
+    --pull=newer \
+    --security-opt label=type:unconfined_t \
+    -v $(pwd)/config.json:/config.json \
+    -v $(pwd)/output:/output \
+    -v /var/lib/containers/storage:/var/lib/containers/storage \
+    quay.io/centos-bootc/bootc-image-builder:latest \
+    --type qcow2 \
+    --config /config.json \
+    --local \
+    localhost/bootc:eln
+```
+
+When using the --local flag, we need to mount the storage path as a volume. With this enabled, it is assumed that the target container is in the container storage.
+
 ### Running the resulting QCOW2 file on Linux (x86_64)
 
 A virtual machine can be launched using `qemu-system-x86_64` or with `virt-install` as shown below.
--- a/bib/cmd/bootc-image-builder/image.go
+++ b/bib/cmd/bootc-image-builder/image.go
@@ -42,6 +42,9 @@ type ManifestConfig struct {
 
 	// TLSVerify specifies whether HTTPS and a valid TLS certificate are required
 	TLSVerify bool
+
+	// Use a local container from the host rather than a repository
+	Local bool
 }
 
 func Manifest(c *ManifestConfig) (*manifest.Manifest, error) {
@@ -65,6 +68,7 @@ func manifestForDiskImage(c *ManifestConfig, rng *rand.Rand) (*manifest.Manifest
 		Source:    c.Imgref,
 		Name:      c.Imgref,
 		TLSVerify: &c.TLSVerify,
+		Local:     c.Local,
 	}
 
 	var customizations *blueprint.Customizations
@@ -142,6 +146,7 @@ func manifestForDiskImage(c *ManifestConfig, rng *rand.Rand) (*manifest.Manifest
 			Source:    c.Imgref,
 			Name:      c.Imgref,
 			TLSVerify: &c.TLSVerify,
+			Local:     c.Local,
 		},
 	}
 	_, err = img.InstantiateManifestFromContainers(&mf, containerSources, runner, rng)
@@ -158,6 +163,7 @@ func manifestForISO(c *ManifestConfig, rng *rand.Rand) (*manifest.Manifest, erro
 		Source:    c.Imgref,
 		Name:      c.Imgref,
 		TLSVerify: &c.TLSVerify,
+		Local:     c.Local,
 	}
 
 	// The ref is not needed and will be removed from the ctor later
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -187,6 +187,7 @@ func manifestFromCobra(cmd *cobra.Command, args []string) ([]byte, error) {
 	rpmCacheRoot, _ := cmd.Flags().GetString("rpmmd")
 	targetArch, _ := cmd.Flags().GetString("target-arch")
 	tlsVerify, _ := cmd.Flags().GetBool("tls-verify")
+	localStorage, _ := cmd.Flags().GetBool("local")
 
 	// translate anaconda-iso to iso to avoid multiple image type checks
 	for idx := range imgTypes {
@@ -240,6 +241,7 @@ func manifestFromCobra(cmd *cobra.Command, args []string) ([]byte, error) {
 		Imgref:       imgref,
 		Repos:        repos,
 		TLSVerify:    tlsVerify,
+		Local:        localStorage,
 	}
 	return makeManifest(manifestConfig, rpmCacheRoot)
 }
@@ -422,6 +424,7 @@ func run() error {
 	manifestCmd.Flags().String("rpmmd", "/rpmmd", "rpm metadata cache directory")
 	manifestCmd.Flags().String("target-arch", "", "build for the given target architecture (experimental)")
 	manifestCmd.Flags().StringArray("type", []string{"qcow2"}, "image types to build [qcow2, ami, iso, raw]")
+	manifestCmd.Flags().Bool("local", false, "use a local container rather than a container from a registry")
 
 	logrus.SetLevel(logrus.ErrorLevel)
 	buildCmd.Flags().AddFlagSet(manifestCmd.Flags())
--- a/test/test_build.py
+++ b/test/test_build.py
@@ -42,6 +42,19 @@ class ImageBuildResult(NamedTuple):
     metadata: dict = {}
 
 
+def parse_request_params(request):
+    # image_type is passed via special pytest parameter fixture
+    testcase_ref = request.param
+    if testcase_ref.count(",") == 2:
+        container_ref, images, target_arch, local = testcase_ref.split(",")
+    elif testcase_ref.count(",") == 1:
+        container_ref, images = testcase_ref.split(",")
+        target_arch = None
+    else:
+        raise ValueError(f"cannot parse {testcase_ref.count}")
+    return container_ref, images, target_arch
+
+
 @pytest.fixture(scope='session')
 def shared_tmpdir(tmpdir_factory):
     tmp_path = pathlib.Path(tmpdir_factory.mktemp("shared"))
@@ -78,15 +91,7 @@ def build_images(shared_tmpdir, build_container, request, force_aws_upload):
 
     :request.parm: has the form "container_url,img_type1+img_type2,arch"
     """
-    # image_type is passed via special pytest parameter fixture
-    testcase_ref = request.param
-    if testcase_ref.count(",") == 2:
-        container_ref, images, target_arch = testcase_ref.split(",")
-    elif testcase_ref.count(",") == 1:
-        container_ref, images = testcase_ref.split(",")
-        target_arch = None
-    else:
-        raise ValueError(f"cannot parse {testcase_ref.count}")
+    container_ref, images, target_arch = parse_request_params(request)
 
     # images might be multiple --type args
     # split and check each one
--- a/test/test_build.py
+++ b/test/test_build.py
@@ -2,9 +2,11 @@ import json
 import os
 import pathlib
 import platform
+import random
 import re
 import shutil
 import subprocess
+import string
 import tempfile
 import uuid
 from contextlib import contextmanager
@@ -45,14 +47,19 @@ class ImageBuildResult(NamedTuple):
 def parse_request_params(request):
     # image_type is passed via special pytest parameter fixture
     testcase_ref = request.param
-    if testcase_ref.count(",") == 2:
+    if testcase_ref.count(",") == 3:
         container_ref, images, target_arch, local = testcase_ref.split(",")
+        local = local is not None
+    elif testcase_ref.count(",") == 2:
+        container_ref, images, target_arch = testcase_ref.split(",")
+        local = False
     elif testcase_ref.count(",") == 1:
         container_ref, images = testcase_ref.split(",")
         target_arch = None
+        local = False
     else:
         raise ValueError(f"cannot parse {testcase_ref.count}")
-    return container_ref, images, target_arch
+    return container_ref, images, target_arch, local
 
 
 @pytest.fixture(scope='session')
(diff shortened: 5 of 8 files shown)


bib/cmd/bootc-image-builder/image.go
```diff
@@ -42,6 +42,9 @@ type ManifestConfig struct {
 
 	// TLSVerify specifies whether HTTPS and a valid TLS certificate are required
 	TLSVerify bool
+
+	// Use a local container from the host rather than a repository
+	Local bool
```
kingsleyzissou, line 47: Maybe a `storage-source` flag would be better

cgwalters, line 47: Was it discussed why not to just *always* fetch the image to containers-storage as a preparatory step, hence squashing the distinction between "remote" and "local"?

Containerfile
```diff
@@ -1,5 +1,5 @@
 FROM registry.fedoraproject.org/fedora:39 AS builder
-RUN dnf install -y git-core golang gpgme-devel libassuan-devel && mkdir -p /build/bib
+RUN dnf install -y git-core golang gpgme-devel libassuan-devel device-mapper-devel btrfs-progs-devel && mkdir -p /build/bib
```
cgwalters, line 2: xref https://github.com/osbuild/bootc-image-builder/pull/126

Containerfile
```diff
@@ -14,11 +14,11 @@ FROM registry.fedoraproject.org/fedora:39
 # - https://github.com/osbuild/osbuild/pull/1468
 COPY ./group_osbuild-osbuild-fedora-39.repo /etc/yum.repos.d/
 RUN dnf install -y osbuild osbuild-ostree osbuild-depsolve-dnf && dnf clean all
+
```
cgwalters, line 17: Any reason for this?

cdrage (commented):
I'm assuming you are mounting the local storage folder too to the bootc-image-builder container?

Containerfile
kingsleyzissou: Nope, just a relic from some local dev. Will update.

bib/cmd/bootc-image-builder/image.go
kingsleyzissou: It was discussed, yes. But I don't see how this is not needed even if we copy a remote container into `containers-storage`

kingsleyzissou:
> I'm assuming you are mounting the local storage folder too to the bootc-image-builder container?

Yeah, exactly

kingsleyzissou:
TODO: update readme (will push later)

kingsleyzissou:
Looks like something in images changed between `v38` and `v41` that's resulting in a different manifest and causing CI to break.

```
diff --git a/images_v38.json b/images_v41.json
index 29acbab..a20d089 100644
--- a/images_v38.json
+++ b/images_v41.json
@@ -383,55 +383,35 @@
             }
           },
           "devices": {
-            "-": {
-              "type": "org.osbuild.loopback",
-              "options": {
-                "filename": "disk.img",
-                "start": 3127296,
-                "size": 17844191
-              }
-            },
-            "boot": {
-              "type": "org.osbuild.loopback",
-              "options": {
-                "filename": "disk.img",
-                "start": 1030144,
-                "size": 2097152
-              }
-            },
-            "boot-efi": {
-              "type": "org.osbuild.loopback",
-              "options": {
-                "filename": "disk.img",
-                "start": 4096,
-                "size": 1026048
-              }
-            },
             "disk": {
               "type": "org.osbuild.loopback",
               "options": {
-                "filename": "disk.img"
+                "filename": "disk.img",
+                "partscan": true
               }
             }
           },
           "mounts": [
             {
-              "name": "-",
+              "name": "part4",
               "type": "org.osbuild.ext4",
-              "source": "-",
-              "target": "/"
+              "source": "disk",
+              "target": "/",
+              "partition": 4
             },
             {
-              "name": "boot",
+              "name": "part3",
               "type": "org.osbuild.ext4",
-              "source": "boot",
-              "target": "/boot"
+              "source": "disk",
+              "target": "/boot",
+              "partition": 3
             },
             {
-              "name": "boot-efi",
+              "name": "part2",
               "type": "org.osbuild.fat",
-              "source": "boot-efi",
-              "target": "/boot/efi"
+              "source": "disk",
+              "target": "/boot/efi",
+              "partition": 2
             }
           ]
         }
```

bib/internal/setup/setup.go
```diff
@@ -53,6 +53,14 @@ func EnsureEnvironment() error {
 	if err := util.RunCmdSync("mount", "--bind", destPath, osbuildPath); err != nil {
 		return err
 	}
+
+	// Ensure we have devfs inside the container to get dynamic loop
+	// loop devices inside the container.
+	devMnt := "/dev/"
+	if err := util.RunCmdSync("mount", "-t", "devtmpfs", "devtmpfs", devMnt); err != nil {
```
cgwalters, line 60: In bootc's similar logic I structured things to be idempotent, because we need to do this for the case of running with `--entrypoint bash` and invoking the CLI multiple times.

Of course, this *will* work today, we'll just end up with a new mount for each invocation.

We'll mostly get that by moving this invocation above the bind mount, which the function uses as an "I already succeeded" marker.

bib/internal/setup/setup.go
```diff
@@ -53,6 +53,14 @@ func EnsureEnvironment() error {
 	if err := util.RunCmdSync("mount", "--bind", destPath, osbuildPath); err != nil {
 		return err
 	}
+
+	// Ensure we have devfs inside the container to get dynamic loop
+	// loop devices inside the container.
+	devMnt := "/dev/"
```
cgwalters, line 59: minor nit: Why the separate variable here?  It doesn't seem useful?

bib/internal/setup/setup.go
cgwalters: Eh, I was wrong; in the success case it will work fine.  But it still seems slightly better to leave the bind mount as an "I succeeded" marker and not do anything after; perhaps add a comment like:

```
// NOTE: Don't add new code here, do it before the bind mount which acts as the final success indicator
```

mvo5: Thank you, that is an excellent suggestion. I pushed to https://github.com/osbuild/bootc-image-builder/pull/238 as it seems to be fine to land without this PR. I hope I understood you suggestion correctly.

bib/internal/setup/setup.go
mvo5: Yeah, silly me, fixed :)

kingsleyzissou:
Will rebase on top of #244

Containerfile
```diff
@@ -1,5 +1,5 @@
 FROM registry.fedoraproject.org/fedora:39 AS builder
-RUN dnf install -y git-core golang gpgme-devel libassuan-devel && mkdir -p /build/bib
+RUN dnf install -y git-core golang gpgme-devel libassuan-devel device-mapper-devel && mkdir -p /build/bib
```
cgwalters, line 2: I think we can safely not care about `devicemapper` FWIW.  But, no objections from me for including it either.

README.md
```diff
@@ -46,6 +46,33 @@ sudo podman run \
...
+    --privileged \
+    --pull=newer \
+    --security-opt label=type:unconfined_t \
+    -v $(pwd)/config.json:/config.json \
+    -v $(pwd)/output:/output \
+    -v /var/lib/containers/storage:/var/lib/containers/storage \
+    quay.io/centos-bootc/bootc-image-builder:latest \
+    --type qcow2 \
+    --config /config.json \
+    -- local \
```
cgwalters, line 70: Typo
```suggestion
    --local \
```

Containerfile
rhatdan: Yes remove it. We don't want it and no reason to add it to the container.  Not sure we want BTRFS either.

test/testutil.py
```diff
@@ -108,3 +109,37 @@ def deregister_ami(ami_id):
         err_msg = err.response["Error"]["Message"]
         print(f"Couldn't deregister image {ami_id}.")
         print(f"Error {err_code}: {err_msg}")
+
+
+def make_fake_tree(basedir: pathlib.Path, fake_content: dict):
```
mvo5, line 114: Those two look like they come from osbuild which is totally fine, but it would be nice to add a comment where they come from. Importing from osbuild would be even nicer but that is probably a bit overkill (or a shared testutil repo which is equally overkill)

test/testcases.py
```diff
@@ -45,9 +45,12 @@ def gen_testcases(what):
             CONTAINERS_TO_TEST["fedora"] + "," + DIRECT_BOOT_IMAGE_TYPES[2],
             CONTAINERS_TO_TEST["centos"] + "," + DIRECT_BOOT_IMAGE_TYPES[2],
             CONTAINERS_TO_TEST["fedora"] + "," + DIRECT_BOOT_IMAGE_TYPES[0],
+            CONTAINERS_TO_TEST["centos"] + "," + DIRECT_BOOT_IMAGE_TYPES[0] + ",,true",
+            CONTAINERS_TO_TEST["fedora"] + "," + DIRECT_BOOT_IMAGE_TYPES[2] + ",,true",
         ]
         # do a cross arch test too
         if platform.machine() == "x86_64":
+            print()
```
mvo5, line 53: Why the print here?

test/test_build.py
```diff
@@ -42,6 +44,24 @@ class ImageBuildResult(NamedTuple):
     metadata: dict = {}
 
 
+def parse_request_params(request):
```
mvo5, line 47: Thanks for extracting this helper!

test/testcases.py
kingsleyzissou: Flip sorry that was for debugging :)

test/testutil.py
kingsleyzissou: Yeah I thought about that too. But it did seem like overkill. I will add comments, like you suggested, pointing back to the original code

README.md
```diff
@@ -46,6 +46,33 @@ sudo podman run \
...
+```bash
+sudo podman run \
+    --rm \
+    -it \
+    --privileged \
+    --pull=newer \
+    --security-opt label=type:unconfined_t \
+    -v $(pwd)/config.json:/config.json \
+    -v $(pwd)/output:/output \
+    -v /var/lib/containers/storage:/var/lib/containers/storage \
```
cgwalters, line 66: Hmm interesting, I'd suggested using `-v /var/lib/containers:/var/lib/containers` elsewhere...but perhaps this is indeed cleaner because AIUI it will avoid gathering e.g. created container roots (as opposed to the underlying images).

README.md
```diff
@@ -46,6 +46,33 @@ sudo podman run \
     quay.io/centos-bootc/fedora-bootc:eln
 ```
 
+### Using local containers
+
+To use containers from local container's storage rather than a registry, we need to ensure two things:
+- the container exists in local storage
```
cgwalters, line 52: I guess I'm ok with this constraint, but in the podman-bootc CLI we intentionally matched `podman run` here and do "pull if not present" (which matches `docker run` of course).

README.md
```diff
@@ -46,6 +46,33 @@ sudo podman run \
...
+    -v $(pwd)/output:/output \
+    -v /var/lib/containers/storage:/var/lib/containers/storage \
+    quay.io/centos-bootc/bootc-image-builder:latest \
+    --type qcow2 \
+    --config /config.json \
+    --local \
+    localhost/bootc:eln
+```
+
+Note we need to pass the `--local` flag, we need to mount the storage path as a volume and then use the name of the local container.
```
cgwalters, line 74: Having an example use the term `localhost/` certainly emphasizes that a local container is supported...but we still support referring to non-`localhost/` images (e.g. `quay.io/...`) right?   So this sentence is misleading in saying "name of the local container" as if that is special in some way, as opposed to an image that happens to already be present in the container storage.

How about:

> When using the `--local` flag, we need to mount the storage path as a volume.  With this enabled, it is assumed that the target container is in the container storage.

test/test_build.py
```diff
@@ -66,9 +73,29 @@ def image_type_fixture(shared_tmpdir, build_container, request, force_aws_upload
...
     ImageBuildResult with the resulting image path and user/password
+    In the case an image is being built from a local container, the
+    function will build the required local container for the test.
     """
-    with build_images(shared_tmpdir, build_container, request, force_aws_upload) as build_results:
-        yield build_results[0]
+    container_ref, images, target_arch, local = parse_request_params(request)
+
+    if not local:
+        with build_images(shared_tmpdir, build_container, request, force_aws_upload) as build_results:
```
ondrejbudai, line 82: I think it's possible to deduplicate these two lines by moving them after the if-else block.

ondrejbudai (commented):
I love this! :heart: One nitpick, but it absolutely be done in a follow-up.

ondrejbudai (approved):

README.md
rhatdan, line 66: Might make more sense to use DEST /usr/lib/containers/storage:ro, since I don't think we want BIB to modify the host storage, or do you plan on modifying the host storage?

/usr/lib/containers/storage should already be setup as an additional store.

bib/cmd/bootc-image-builder/image.go
```diff
@@ -42,6 +42,9 @@ type ManifestConfig struct {
 
 	// TLSVerify specifies whether HTTPS and a valid TLS certificate are required
 	TLSVerify bool
+
+	// Use a local container from the host rather than a repository
```
rhatdan, line 46: I don't think BIB is using container but container images?  Should this be
Use a local container image ...

bib/cmd/bootc-image-builder/main.go
```diff
@@ -422,6 +424,7 @@ func run() error {
 	manifestCmd.Flags().String("rpmmd", "/rpmmd", "rpm metadata cache directory")
 	manifestCmd.Flags().String("target-arch", "", "build for the given target architecture (experimental)")
 	manifestCmd.Flags().StringArray("type", []string{"qcow2"}, "image types to build [qcow2, ami, iso, raw]")
+	manifestCmd.Flags().Bool("local", false, "use a local container rather than a container from a registry")
```
rhatdan, line 427: Do we need a local flag, or should BIB always look for a local image and then pull the image if it is not in local storage?

README.md
ondrejbudai: Hmm, but if we do a read-only mount, we get into issues that podman/skopeo cannot lock the store. Any thoughts about this?

bib/cmd/bootc-image-builder/main.go
ondrejbudai: Is that your preference? I think I would like that as well, but we wanted to do the minimal change for now, and refactor it later.

README.md
kingsleyzissou: So yeah we could mount another store other than `/var/lib/containers/storage`, that's no hassle. Testing locally I was using another store entirely. So maybe we can update the docs accordingly.

The `ro` is a bit more complicated, we run into loads of locking issues. We aren't even able to generate the manifest. Do you have any suggestions for getting around that? fyi @achilleas-k

rhatdan: Readonly means that the images are readonly in the addiitonal store in /usr/lib/containers/storage, while BIB would have its own /var/lib/containers/storage and able to pull images or run containers there.  There is little to protect the container from the user mucking around with the external containers-storage, so it is more of tell the user do not do that.

bib/cmd/bootc-image-builder/main.go
rhatdan: I just don't see why I would volume mount the images in and then not use local. Also I think we should allow non local pulls when storage is mounted.  One image might require many layers that are already stored in the local storage but have different content in the upper layers.

README.md
ondrejbudai: However, when we use the podman/skopeo tools to inspect the containers in the container store, the tools need to put locks **inside the store**. When the store is read/only, podman/skopeo (called by BIB) cannot create the locks, and they will just fail.

bib/cmd/bootc-image-builder/main.go
cgwalters: Defaulting to local storage will be a change for consumers, but I think it's the right time to make that change, and we add a `--remote` flag for the previous default.

ondrejbudai: Sounds sane. Just to double-check: Is it fine that bib will use different versions of podman/skopeo to manipulate the store than the host has? In other words, the format of the containers store is stable, right?

rhatdan: Yes it has been stable. Definitely forward stable.  But I doubt backwards would be an issue.  Besides we are just talking about images here not containers. When we to to using composefs we might need to make sure we don't have issues.

@giuseppe Thoughts?

PR #123, opened 2024-01-15
test: add boot test on arm64 as well

mvo5:
This commit enables arm64 boot tests for the generated image.


test/vm.py
```diff
@@ -85,32 +87,62 @@ class QEMU(VM):
     # TODO: support qemu-system-aarch64 too :)
     QEMU = "qemu-system-x86_64"
```
cgwalters, line 88: It looks like this is dead code now.

cgwalters:
> NOTE that it's failing right now and we strongly suspect that is because the ELN kernel in the aarch64 contianer does not actually boot (at least not in qemu).

cc https://github.com/CentOS/centos-bootc/pull/364 - if we do that we'll also need to adjust this project to clearly focus on the stream9 image.

PR #126, opened 2024-01-16
build: Use tags to trim down containers/storage dependencies

cgwalters:
The default depends on btrfs, etc. that we don't need because ultimately we're just forking off skopeo anyways.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,6 +1,11 @@
 #!/bin/bash
 
 set -euo pipefail
+# Keep this in sync with e.g. https://github.com/containers/podman/blob/2981262215f563461d449b9841741339f4d9a894/Makefile#L51
+# It turns off the esoteric containers-storage backends that add dependencies
+# on things like btrfs that we don't need.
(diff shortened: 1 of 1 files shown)


cgwalters:
Digging into the TMT logs, the fatal error there looks like
> cp: error writing '/output/image/./disk.raw': No space left on device

cgwalters:
This PR is useful for https://github.com/osbuild/bootc-image-builder/pull/129 too to trim down the package dependencies

cgwalters:
I think this one should be quite safe to get in and will be helpful for increasing our compatibility with using c9s as a base, and avoiding needing to pull in much more than a go compiler at build time.

mvo5 (approved):
Thank you, this looks fine!

achilleas-k:
Seeing `cp: error writing '/output/image/./disk.raw': No space left on device` in the testing farm logs.  Did the configuration of the runners change?

achilleas-k (approved):

cgwalters:
Can we just override the TMT tests?

cgwalters:
OK rebased :surfer: and hopefully this is will unblock other PRs.

PR #129, opened 2024-01-17
 Use C9S

cgwalters:
This is currently broken however because `osbuild-depsolve-dnf`
appears to be missing from the yum repositories...probably
because it needs to be added to the giant bizarre JSON file
in https://gitlab.com/redhat/centos-stream/release-engineering/pungi-centos/-/blob/centos-9-stream/prepopulate.json?ref_type=heads

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -1,5 +1,5 @@
-FROM registry.fedoraproject.org/fedora:39 AS builder
-RUN dnf install -y git-core golang gpgme-devel libassuan-devel && mkdir -p /build/bib
+FROM registry.access.redhat.com/ubi9/ubi:latest AS builder
+RUN dnf install -y git-core golang && mkdir -p /build/bib
 COPY bib/go.mod bib/go.sum /build/bib
 RUN cd /build/bib && go mod download
 COPY build.sh /build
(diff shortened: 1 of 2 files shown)


Containerfile
```diff
@@ -1,19 +1,15 @@
...
-FROM registry.fedoraproject.org/fedora:39
-# Install newer osbuild to fix the loop bug, see
-# - https://github.com/osbuild/bootc-image-builder/issues/7
-# - https://github.com/osbuild/bootc-image-builder/issues/9
-# - https://github.com/osbuild/osbuild/pull/1468
-COPY ./group_osbuild-osbuild-fedora-39.repo /etc/yum.repos.d/
-RUN dnf install -y osbuild osbuild-ostree osbuild-depsolve-dnf && dnf clean all
+FROM quay.io/centos/centos:stream9
+# FROM registry.access.redhat.com/ubi9/ubi:latest
+RUN dnf install -y osbuild osbuild-ostree  && dnf clean all
```
mvo5, line 12: Pardon my ignorance on this - we will want to install more tools in the future into the bib container:
- qemu-img so that we can convert the raw -> cqow2
- qemu-user so that we can cross build images when host arch != target arch, the idea is that the bootc image must be available for both arches and then the buildroot is setup with the host arch and we build the target image but some (small) tasks will happen inside the target arch tree via chroot and for that qemu-user and a binfmt_misc handler is needed to handle this transparently

Will this be an issue when we move here?

cgwalters: Yes, moving the buildroot requirements here makes sense to me.  However, that is orthogonal to this PR because we still end up running ELN today regardless, right?

Containerfile
```diff
@@ -1,19 +1,15 @@
-FROM registry.fedoraproject.org/fedora:39 AS builder
-RUN dnf install -y git-core golang gpgme-devel libassuan-devel && mkdir -p /build/bib
+FROM registry.access.redhat.com/ubi9/ubi:latest AS builder
```
lmilbaum, line 1: Would you mind pinning the version instead of using the latest tag?

lmilbaum (changes requested):

Containerfile
cgwalters: You want to use renovate to update it?

I dunno...part of the idea here is: this is RHEL, the change that this breaks us should be 0.0001% - if it does we have totally failed at our jobs.

PR #130, opened 2024-01-17
build: Drop osbuild copr

cgwalters:
The change we needed landed in the main repos;
this is prep for https://github.com/osbuild/bootc-image-builder/pull/129 which rolls in this alongside a larger change.


cgwalters:
(didn't test this specific one locally)

PR #137, opened 2024-01-18
testutil: remove "match" for centos8/9 compatbility

mvo5:
Support for `match` in python is only available in version 3.10. However centos/rhel 8 and 9 have lower versions so this will need to be moved to `if/elif/else` for now.

See https://github.com/osbuild/bootc-image-builder/pull/109


bcl (approved):
Thanks!

PR #138, opened 2024-01-18
bib: use container based buildroot when building bootc disk images (HMS-3318)

mvo5:
Instead of using the trandition rpm based buildroot this commit
switches the buildroot to the target bootc container. The rational
is that the low-level tooling like mkfs etc should match the target
system.

Once that has landed and we can use container based buildroots
I will start working on `bootc install-filesystem` instead of the ostree
pipeline.

PR #139, opened 2024-01-18
bib: add experimental cross arch image building

mvo5:
[This is based on https://github.com/osbuild/bootc-image-builder/pull/138 and only the last two commits are relevant]

This draft PR outlines how we could do cross architecture image building. I.e. you `bootc-image-builder` on a mac and target an `amd64` machine.

The usage would be:
```
sudo podman run ... \
  --type qcow2 \
  --target-arch amd64 \
```
and this is predicated on the idea that the target container is available for both host and target architecture. When that is the case the container for the host architecture is used as a buildroot, the container for the target architecture is the target tree. Note that some stages chroot into the target tree so `qemu-user` is essential for this to work so that target achitecture code can be run transparently (because only very small amounts of code need to run emulated this is pretty fast still).

### DIFF
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -115,9 +115,19 @@ func makeManifest(c *ManifestConfig, cacheRoot string) (manifest.OSBuildManifest
 	}
 
 	// resolve container
-	resolver := container.NewResolver(c.Architecture.String())
+	hostArch := arch.Current()
+	resolverNative := container.NewResolver(hostArch.String())
+	resolverTarget := container.NewResolver(c.Architecture.String())
+
 	containerSpecs := make(map[string][]container.Spec)
 	for plName, sourceSpecs := range manifest.GetContainerSourceSpecs() {
+		var resolver container.Resolver
+		if plName == "build" {
+			resolver = resolverNative
+		} else {
+			resolver = resolverTarget
+		}
+
 		for _, c := range sourceSpecs {
 			resolver.Add(c)
 		}
@@ -152,8 +162,8 @@ func saveManifest(ms manifest.OSBuildManifest, fpath string) error {
 }
 
 func manifestFromCobra(cmd *cobra.Command, args []string) ([]byte, error) {
-	hostArch := arch.Current()
-	repos, err := loadRepos(hostArch.String())
+	buildArch := arch.Current()
+	repos, err := loadRepos(buildArch.String())
 	if err != nil {
 		return nil, err
 	}
@@ -163,6 +173,20 @@ func manifestFromCobra(cmd *cobra.Command, args []string) ([]byte, error) {
 	configFile, _ := cmd.Flags().GetString("config")
 	tlsVerify, _ := cmd.Flags().GetBool("tls-verify")
 	imgType, _ := cmd.Flags().GetString("type")
+	targetArch, _ := cmd.Flags().GetString("target-arch")
+	if targetArch != "" {
+		// TODO: detect if binfmt_misc for target arch is
+		// available, e.g. by mounting the binfmt_misc fs into
+		// the container and inspects the files or by
+		// including tiny statically linked target-arch
+		// binaries inside our bib container
+		fmt.Fprintf(os.Stderr, "WARNING: target-arch is experimental and needs an installed 'qemu-user' package\n")
+		if imgType == "iso" {
+			return nil, fmt.Errorf("cannot build iso for different target arches yet")
+		}
+		buildArch = arch.FromString(targetArch)
+	}
+	// TODO: add "target-variant", see https://github.com/osbuild/bootc-image-builder/pull/139/files#r1467591868
 
 	var config *BuildConfig
 	if configFile != "" {
@@ -179,7 +203,7 @@ func manifestFromCobra(cmd *cobra.Command, args []string) ([]byte, error) {
 		ImgType:      imgType,
 		Config:       config,
 		Repos:        repos,
-		Architecture: hostArch,
+		Architecture: buildArch,
 		TLSVerify:    tlsVerify,
 	}
 	return makeManifest(manifestConfig, rpmCacheRoot)
@@ -315,6 +339,7 @@ func run() error {
 	manifestCmd.Flags().String("config", "", "build config file")
 	manifestCmd.Flags().String("type", "qcow2", "image type to build [qcow2, ami]")
 	manifestCmd.Flags().Bool("tls-verify", true, "require HTTPS and verify certificates when contacting registries")
+	manifestCmd.Flags().String("target-arch", "", "build for the given target architecture (experimental)")
 
 	logrus.SetLevel(logrus.ErrorLevel)
 	buildCmd.Flags().AddFlagSet(manifestCmd.Flags())
--- a/test/vm.py
+++ b/test/vm.py
@@ -1,6 +1,7 @@
 import abc
 import os
 import pathlib
+import platform
 import subprocess
 import sys
 import time
@@ -93,10 +94,8 @@ def find_ovmf():
 
 class QEMU(VM):
     MEM = "2000"
-    # TODO: support qemu-system-aarch64 too :)
-    QEMU = "qemu-system-x86_64"
 
-    def __init__(self, img, snapshot=True, cdrom=None):
+    def __init__(self, img, arch="", snapshot=True, cdrom=None):
         super().__init__()
         self._img = pathlib.Path(img)
         self._qmp_socket = self._img.with_suffix(".qemp-socket")
@@ -104,22 +103,37 @@ class QEMU(VM):
         self._snapshot = snapshot
         self._cdrom = cdrom
         self._ssh_port = None
+        if not arch:
+            arch = platform.machine()
+        self._arch = arch
 
     def __del__(self):
         self.force_stop()
 
-    # XXX: move args to init() so that __enter__ can use them?
-    def start(self, wait_event="ssh", snapshot=True, use_ovmf=False):
-        if self.running():
-            return
-        log_path = self._img.with_suffix(".serial-log")
-        self._ssh_port = get_free_port()
-        self._address = "localhost"
-        qemu_cmdline = [
-            self.QEMU, "-enable-kvm",
+    def _gen_qemu_cmdline(self, snapshot, use_ovmf):
+        if self._arch in ("arm64", "aarch64"):
+            qemu_cmdline = [
+                "qemu-system-aarch64",
+                "-machine", "virt",
+                "-cpu", "cortex-a57",
+                "-smp", "2",
+                "-bios", "/usr/share/AAVMF/AAVMF_CODE.fd",
+            ]
+        elif self._arch in ("amd64", "x86_64"):
+            qemu_cmdline = [
+                "qemu-system-x86_64",
+                "-M", "accel=kvm",
+                # get "illegal instruction" inside the VM otherwise
+                "-cpu", "host",
+            ]
+            if use_ovmf:
+                qemu_cmdline.extend(["-bios", find_ovmf()])
+        else:
+            raise ValueError(f"unsupported architecture {self._arch}")
+
+        # common part
+        qemu_cmdline += [
             "-m", self.MEM,
-            # get "illegal instruction" inside the VM otherwise
-            "-cpu", "host",
             "-serial", "stdio",
             "-monitor", "none",
             "-netdev", f"user,id=net.0,hostfwd=tcp::{self._ssh_port}-:22",
@@ -128,18 +142,23 @@ class QEMU(VM):
         ]
         if not os.environ.get("OSBUILD_TEST_QEMU_GUI"):
             qemu_cmdline.append("-nographic")
-        if use_ovmf:
-            qemu_cmdline.extend(["-bios", find_ovmf()])
         if self._cdrom:
             qemu_cmdline.extend(["-cdrom", self._cdrom])
         if snapshot:
             qemu_cmdline.append("-snapshot")
         qemu_cmdline.append(self._img)
-        self._log(f"vm starting, log available at {log_path}")
+        return qemu_cmdline
+
+    # XXX: move args to init() so that __enter__ can use them?
+    def start(self, wait_event="ssh", snapshot=True, use_ovmf=False):
+        if self.running():
+            return
+        self._ssh_port = get_free_port()
+        self._address = "localhost"
 
         # XXX: use systemd-run to ensure cleanup?
         self._qemu_p = subprocess.Popen(
-            qemu_cmdline,
+            self._gen_qemu_cmdline(snapshot, use_ovmf),
             stdout=sys.stdout,
             stderr=sys.stderr,
         )
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -72,7 +72,7 @@ jobs:
     - name: Install test dependencies
       run: |
         sudo apt update
-        sudo apt install -y podman python3-pytest python3-paramiko python3-boto3 flake8 qemu-system-x86
+        sudo apt install -y podman python3-pytest python3-paramiko python3-boto3 flake8 qemu-system-x86 qemu-efi-aarch64 qemu-system-arm qemu-user-static
     - name: Diskspace (before)
       run: |
         df -h
--- a/plans/all.fmf
+++ b/plans/all.fmf
@@ -12,6 +12,7 @@ provision:
 prepare:
   how: install
   package:
+    - edk2-aarch64
     - podman
     - pytest
     - python3-boto3
@@ -19,6 +20,8 @@ prepare:
     - python3-paramiko
     - python3-pip
     - qemu-kvm
+    - qemu-system-aarch64
+    - qemu-user-static
 execute:
   how: tmt
   script: |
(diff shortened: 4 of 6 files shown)


cgwalters:
One thing to consider here instead of invoking qemu ourselves is to just `podman run --arch amd64 bootc-image-builder` instead.

Experimenting with things around this I ran into https://github.com/containers/podman/issues/15300 I think.

> because only very small amounts of code need to run emulated this is pretty fast still

Is that true?  Either way we're going to be executing the target OS code right?  I think it's basically just that we're overall I/O bound in general.

mvo5:
> One thing to consider here instead of invoking qemu ourselves is to just `podman run --arch amd64 bootc-image-builder` instead.
> 
> Experimenting with things around this I ran into [containers/podman#15300](https://github.com/containers/podman/issues/15300) I think.

Thank you for the pointer and idea - I don't mind taking this route and will happily experiment with it. The reason I picked the approach in this PR is that most of the work that osbuild is doing happen in the buildroot. So making this native arch seemed expedient. However I think your point below is very valid, overall even during that stage osbuild is really not doing much (in a container world at least) so it maybe well be not worth the complexity and we can go all in on (slow) emulation for the buildroot work too.
 
> > because only very small amounts of code need to run emulated this is pretty fast still
> 
> Is that true? Either way we're going to be executing the target OS code right? I think it's basically just that we're overall I/O bound in general.

What I perceive is that this approach cross builds an bootc fedora image from amd64 for aarch64 in under a minute - this is why I wrote this comment. *But* as I wrote above I did not actually test how long it takes to have the buildroot running in emulation or even to just use podman --arch. It may well be that this is still very much good enough because of the i/o bound nature of the workload. I will play with this in the next few days and report back. My main point with this PR was to socialize the idea and get feedback and that I certainly archived :) Thank you!

rhatdan:
I think this is wrong, it is going to be way too slow.  We should run bootc-image-builder in the native arch and then pass into it the --arch flag.

rhatdan:
Everything runs really slowly in emulation mode, if we can run the commands in native and just pull and convert in alternative arch, the performance will be much better.

mvo5:
> I think this is wrong, it is going to be way too slow. We should run bootc-image-builder in the native arch and then pass into it the --arch flag.

When I ran this for testing it was suprisingly fast because the buildroot which does the bulk of the work runs as native code. Only the very few pieces that need to chroot into the build tree during the build run in emulation.

rhatdan:
I believe I was mistaken, when I read
```
sudo podman run ... \
  --type qcow2 \
  --target-arch amd64 \
```
My brain interpeted the --target-arch as the --arch option in Podman, meaning the BIB would be running in emulation.  The --target-arch going to BIB is exactly what I want.

https://www.youtube.com/watch?v=OjYoNL4g5Vg

rhatdan:
This is exactly what we need.

Containerfile
```diff
@@ -1,5 +1,5 @@
 FROM registry.fedoraproject.org/fedora:39 AS builder
-RUN dnf install -y git-core golang gpgme-devel libassuan-devel && mkdir -p /build/bib
+RUN dnf install -y git-core golang gpgme-devel libassuan-devel btrfs-progs-devel lvm2-devel && mkdir -p /build/bib
```
rhatdan, line 2: We should be just build-taging these out.  You don't need to support btrfs or lvm.

bib/cmd/bootc-image-builder/main.go
```diff
@@ -156,14 +166,21 @@ func saveManifest(ms manifest.OSBuildManifest, fpath string) error {
...
-	repos := loadRepos(hostArch.String())
+	buildArch := arch.Current()
+	repos := loadRepos(buildArch.String())
 
 	imgref := args[0]
 	rpmCacheRoot, _ := cmd.Flags().GetString("rpmmd")
 	configFile, _ := cmd.Flags().GetString("config")
 	tlsVerify, _ := cmd.Flags().GetBool("tls-verify")
 	imgType, _ := cmd.Flags().GetString("type")
+	targetArch, _ := cmd.Flags().GetString("target-arch")
```
rhatdan, line 177: We should be handling "target-variant" as well.

Containerfile
```diff
@@ -13,7 +13,7 @@ FROM registry.fedoraproject.org/fedora:39
 # - https://github.com/osbuild/bootc-image-builder/issues/9
 # - https://github.com/osbuild/osbuild/pull/1468
 COPY ./group_osbuild-osbuild-fedora-39.repo /etc/yum.repos.d/
-RUN dnf install -y osbuild osbuild-ostree osbuild-depsolve-dnf podman qemu-img && dnf clean all
+RUN dnf install -y osbuild osbuild-ostree osbuild-depsolve-dnf podman qemu-img qemu-user-static && dnf clean all
```
rhatdan, line 16: qemu-user-static is not available in RHEL, so if we ever move to a RHEL based image this will not work.

Containerfile
mvo5: Thank you! Colin fixed this already, once I rebased to "main" this will go away.

Containerfile
mvo5: Hm, in the general case we need `qemu-user` because some of the osbuild stages will chroot into the target. Bootupd is installed this way today, i.e. we essentially chroot into the deployment root and just call bootupd. Is that a showstopper for this approach or should we go with it for now (given that we have limited time until the demo etc)?

bib/cmd/bootc-image-builder/main.go
mvo5: Happy to add more here - but pardon my ignorance, what does `target-variant` mean in this context?

Containerfile
mvo5: Another thing that might be an issue is the arch specific mkfs things that Colin pointed out in https://github.com/osbuild/bootc-image-builder/issues/153#issuecomment-1910453402 - we need to see what we can do here.

bib/cmd/bootc-image-builder/main.go
rhatdan: Some Arm platforms have multiple variants.  So if BIB is calling skopeo or podman then it is needs to pass the variant as well as the Arch if the user specified a variant.

Containerfile
rhatdan: Lets get this working on Fedora first and then we can worry about RHEL.

rhatdan:
I would like to get this merged to add the feature and then we can continue to work on some of the stuff Colin asked for.

mvo5:
This is ready now, it builds on top of https://github.com/osbuild/bootc-image-builder/pull/148 and does a "x86"->"arm" cross build with boot test. The boot test takes around 30s on my system, the test a total runtime of 5min. 

Once things looks good in here and this is merged I will work on the "arm"->"x86" tests. I know it's an important use-case but I find it a bit painful to test on mac as I need to do this remote only.

bib/cmd/bootc-image-builder/main.go
mvo5: I added a TODO here for now, if you could share some container or any other link where I can inspect this a bit more that would be super helpful. I will then add/fix this in a followup.

mvo5:
The integration test fails right now with:
```
subprocess.CalledProcessError: Command '['chroot', '/run/osbuild/tree', 'useradd', '--groups', 'wheel', '--password', '$6$U379toL5x8.UFL8m$/wFhWBOSctUT0bCEXPw.vdiNf.vb1NYKU7ZK712bW9G3s/vZtgWkrPc9XMvegStvCzBRGkwqzoEg4/WwsGOA90', 'test']' returned non-zero exit status 126.
```
which means that the binfmt-misc handler needs to be registered from inside the container (we cannot rely on having this on the host OS). Hopefully just a SMOP.

shi2wei3:
Just curious, are we going to switch to use bootc install for bib by default or just as an option? does this cross arch build also works after switch over to bootc install?

bib/cmd/bootc-image-builder/main.go
```diff
@@ -315,6 +333,7 @@ func run() error {
 	manifestCmd.Flags().String("config", "", "build config file")
 	manifestCmd.Flags().String("type", "qcow2", "image type to build [qcow2, ami]")
 	manifestCmd.Flags().Bool("tls-verify", true, "require HTTPS and verify certificates when contacting registries")
+	manifestCmd.Flags().String("target-arch", "", "build for the given target architecture")
```
cdrage, line 336: shouldn't we specify that you can only enter `x86_64` and `aarch64` ? since that's what's being added to `entrypoint.sh`: https://github.com/osbuild/bootc-image-builder/pull/139/files#diff-6f9d41d046756f0ddc2fcee0626bdb50100d12b88f293734eff742818e03efa2R6

cdrage (changes requested):

test/vm.py
```diff
@@ -93,33 +94,46 @@ def find_ovmf():
...
+    def _gen_qemu_cmdline(self, snapshot, use_ovmf):
+        if self._arch in ("arm64", "aarch64"):
+            qemu_cmdline = [
+                "qemu-system-aarch64",
+                "-machine", "virt",
+                "-cpu", "cortex-a57",
+                "-smp", "2",
+                "-bios", "/usr/share/AAVMF/AAVMF_CODE.fd",
+            ]
+        elif self._arch in ("amd64", "x86_64"):
```
cdrage, line 122: i also see the options here, so I'm assuming command line it should only accept amd64 or arm64?

cdrage (changes requested):

test/vm.py
rhatdan: We should eventually support all of the platforms that RHEL supports, otherwise our parent company will not be happy.

mvo5:
> Just curious, are we going to switch to use bootc install for bib by default or just as an option? does this cross arch build also works after switch over to bootc install?

The plan is to move to bootc install, there is already work in osbuild/images happening for this.

bib/cmd/bootc-image-builder/main.go
mvo5: Thanks for your review - indeed, the code in this area is a bit loose right now. In theory (if the host has qemu-user{,-static} installed) this will just work magically for all target arches that the container is build for.

However the user experience if the host does not have qemu-user{,-static} is bad hence I added it to the entrypoint.sh but I am having second thoughts about it. The fact that we have to mess with the (non-namespaced, system-wide) binfmt_misc from inside the container is not great. There could also be confusing bugs, because the binfmt_misc registration including the interpreter path is global so if someone has a custome /usr/local/lib/qemu-binfmt-aarch64 or something inside the container this path will not exist and fun errors happen. We also cannot "fix" it from inside the contianer without messing up the host. I think this is a long winded way of saying that we may simply need to error if binfmt_misc is missing and ask the user kindly to install it system-wide. Podman-machine of course can ensure this is avaiable inside the podman machine it creates.

rhatdan: Yes I think this is the users or someone outside of the container images responsibility.  If I am attempting to build a different Arch then the default arch without qemu-user-static-ARCH then then podman build will blow up, long before we attempt to convert the image.  You should just pass the arch down and allow the apps to blow up.
Do not attempt to modify the host system, that would be bad behavior and very unexpected.

mvo5: Thank you again for your feedback! I updated the code to not do anything with binfmt_misc from inside the container. For now it will just print a warning that qemu-user is required, we could do something elaborate here by detecting this but that seems like a step(2) feature.

test/vm.py
mvo5: We definitely need to test more scenarios here but if the container is available for both host and target arch  and qemu-user is available for the target arch the existing code should DTRT even for e.g. ppc64le.

achilleas-k:
FTR:
```
› uname -m
arm64

› time podman run \
    -it --rm \
    --privileged \
    --pull=newer \
    --volume ./output:/output \
    quay.io/centos-bootc/bootc-image-builder:xarch \
    --type qcow2 \
    --target-arch x86_64 \
Generating manifest-qcow2.json ... WARNING: target-arch is experimental and needs an installed 'qemu-user' package
DONE
Building manifest-qcow2.json
source/org.osbuild.skopeo (org.osbuild.skopeo): Getting image source signatures
...
build:          04aded43b0432e5852df5ebf323bc54d73be4570745a47b6116f52a4ef46e269
ostree-deployment:      67e7f9dfb2805e1df572e33c4e36e8954532f0f8175586e73cf8abee2d0a3bf4
image:          c4ef94328cca66b6c15ff1d5c98ecad1b15b395270294c9f80a3f03ea97e4d66
qcow2:          946f9df64eb7daf87a07d28d355fcc08237f3dce134ae31cfc9fef6d0aa45848
Build complete!
Results saved in
/output
podman run -it --rm --privileged --pull=newer --security-opt  --volume         0.05s user 0.03s system 0% cpu 2:17.77 total
```

```
› uname -m
x86_64

› time sudo podman run \
    --rm \
    -it \
    --privileged \
    --pull=newer \
    --volume ./output/:/output \
    quay.io/centos-bootc/bootc-image-builder:xarch \
    --type qcow2 \
    --target-arch arm64 \
Generating manifest-qcow2.json ... WARNING: target-arch is experimental and needs an installed 'qemu-user' package
DONE
Building manifest-qcow2.json
source/org.osbuild.skopeo (org.osbuild.skopeo): Getting image source signatures
...
build:          b09fb0efe1f39f826c46fcb8e9c1f7de13d0a0308de5d664404873c72127ce3f
ostree-deployment:      b14ddbb55e64bc2961d65ee8505246824bba3ac36d12320d0360ee77caa89973
image:          5cb0df803aa1d50d031b1d148fd3696de3f6c582290edb729bcba512f9d6dcb3
qcow2:          510c0be6224910b3636577aefcc4819cd55cbeb5eb4ba3e9e8476255c69bad31
Build complete!
Results saved in
/output
sudo podman run --rm -it --privileged --pull=newer --security-opt  --volume    0.01s user 0.02s system 0% cpu 4:10.20 total
```

This is awesome.

bib/cmd/bootc-image-builder/main.go
```diff
@@ -115,9 +115,19 @@ func makeManifest(c *ManifestConfig, cacheRoot string) (manifest.OSBuildManifest
 	}
 
 	// resolve container
-	resolver := container.NewResolver(c.Architecture.String())
+	hostArch := arch.Current()
+	resolverNative := container.NewResolver(hostArch.String())
+	resolverTarget := container.NewResolver(c.Architecture.String())
```
achilleas-k, line 120: If I'm reading this right, in the standard case, we'll get two identical resolvers and each one will get the same container it will have to resolve.  It's not _terrible_ but a bit wasteful.  Maybe we could instead:
```go
resolvers[hostArch.String()] = container.NewResolver(hostArch.String())
if hostArch.String() != c.Architectyre.String() {
    resolvers[c.Architecture.String()] = container.NewResolver(c.Architecture.String()) 

}

...

if plName == "build" {
    rarch = hostArch.String()
} else {
    rarch = c.Architecture.String()
}
for _, c := range sourceSpecs {
    rarch.Add(c)
}
```

I now also realise that even in this case, we're resolving the same container twice (and have been since we started using the container as a build root).
Again, not terrible, but maybe we can be smarter about it :)

achilleas-k (commented):
Tiny comment about minor optimisations but happy to merge and follow-up in another PR.
Keeping un-approved until https://github.com/osbuild/bootc-image-builder/pull/148 is merged and this is rebased.

bib/cmd/bootc-image-builder/main.go
mvo5: Thank you, I will look at this, maybe in a followup as it's not *that* pressing but I agree that it's a bit wasteful

mvo5:
Hm, this one is running out of diskspace again :/

bib/cmd/bootc-image-builder/main.go
mvo5: Thanks again, I was thinking it could be something like:
```go
index 5e9ed2d..9bb6a5b 100644
@@ -115,13 +115,18 @@ func makeManifest(c *ManifestConfig, cacheRoot string) (manifest.OSBuildManifest
        }
 
        // resolve container
-       hostArch := arch.Current()
-       resolverNative := container.NewResolver(hostArch.String())
-       resolverTarget := container.NewResolver(c.Architecture.String())
+       hostArch := arch.Current().String()
+       targetArch := c.Architecture.String()
+
+       resolverNative := container.NewResolver(hostArch)
+       resolverTarget := resolverNative
+       if hostArch != targetArch {
+               resolverTarget = container.NewResolver(targetArch)
+       }
 
        containerSpecs := make(map[string][]container.Spec)
        for plName, sourceSpecs := range manifest.GetContainerSourceSpecs() {
-               var resolver container.Resolver
+               var resolver *container.Resolver
                if plName == "build" {
                        resolver = resolverNative
                } else {
```
which requires https://github.com/osbuild/images/pull/433 (so that we get pointers to the resolver first) - I have no super strong opinion here though, happy to discuss more if you think this is less obvious or too verbose :)

mvo5:
The main piece of this PR got merged via https://github.com/osbuild/bootc-image-builder/pull/179 and is available now. The testing got split into https://github.com/osbuild/bootc-image-builder/pull/181 mostly to avoid too much spam for this PR as I had to force push a bunch of times to figure the right GH runner setup with enough diskspace out. That seems to be fortunately solved now too. Once 181 is merged I will close this one as it's empty. Thanks everyone for your great insights and help getting this ready. I really appreciate that!

PR #141, opened 2024-01-19
many: refactor usage of check/fail a bit

mvo5:
This PR tweaks the code a bit to pass more errors up and exit only on the caller side. This will make writing unit tests easier and also feels a bit more goish. 

Split into multiple individual commits to see the refactor chain but it's fine to also just merge via a single squash, the history is really not that interesting :)

### DIFF
--- a/bib/cmd/bootc-image-builder/image.go
+++ b/bib/cmd/bootc-image-builder/image.go
@@ -2,7 +2,7 @@ package main
 
 import (
 	cryptorand "crypto/rand"
-	"fmt"
+	"log"
 	"math"
 	"math/big"
 	"math/rand"
@@ -56,11 +56,11 @@ func Manifest(c *ManifestConfig) (*manifest.Manifest, error) {
 	case "iso":
 		img, err = pipelinesForISO(c, rng)
 	default:
-		fail(fmt.Sprintf("Manifest(): unsupported image type %q", c.ImgType))
+		log.Fatalf("Manifest(): unsupported image type %q", c.ImgType)
 	}
 
 	if err != nil {
-		fail(err.Error())
+		log.Fatalf("%s", err)
 	}
 
 	mf := manifest.New()
@@ -73,7 +73,7 @@ func Manifest(c *ManifestConfig) (*manifest.Manifest, error) {
 
 func pipelinesForDiskImage(c *ManifestConfig, rng *rand.Rand) (image.ImageKind, error) {
 	if c.Imgref == "" {
-		fail("pipeline: no base image defined")
+		log.Fatalf("pipeline: no base image defined")
 	}
 	ref := "ostree/1/1/0"
 	containerSource := container.SourceSpec{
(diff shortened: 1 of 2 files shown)


bcl (approved):
I like this. Also makes it easier to move code into libraries if needed later.

cgwalters (commented):
I have no opposition to this, but I think the correct fix is actually to use `RunE` - e.g. `cmdBuild` should just return an `error` type.

mvo5:
> I have no opposition to this, but I think the correct fix is actually to use `RunE` - e.g. `cmdBuild` should just return an `error` type.

Thanks for your review and feedback. Part of the diff is to use `RunE` for `cmdBuild (c.f. https://github.com/osbuild/bootc-image-builder/pull/141/files#diff-cbdce8fdffa710f25830f44fc4b0d5be917240f81c72e59e7003dd62fb62349cR301) is there more this should do or was this particular change just lost in the noise in the diff when reviewing (it's a bit long and repetitive, sorry for that)?

cgwalters (approved):
Ahhh sorry I only skimmed the commit description and my review started by clicking on the first commit, not the end result aka "files changed" where it is clearly correct.  Sorry about that and LGTM.

achilleas-k (approved):
Thanks for the cleanup.  LGTM!

achilleas-k:
One thing to note: When the call to `RunE` fails, by default cobra will print the `help` output.  I find it's nicer to use `RunE` failures to validate input and options but `log.Fatal()` for runtime errors.  Right now, if for example the manifest write fails, you'll get an error message and the help output:
```
❯ touch ./output/manifest-qcow2.json
sudo chattr +i ./output/manifest-qcow2.json
sudo podman run \
    --rm \
    -it \
    --privileged \
    --pull=newer \
    -v ./output:/output \
    localhost/bootc-image-builder:not-a-checker \
    --type qcow2 \
Generating manifest-qcow2.json ... DONE
Error: failed to create output file "/output/manifest-qcow2.json": open /output/manifest-qcow2.json: operation not permitted
Usage:
  bootc-image-builder build

Flags:
      --aws-ami-name string   name for the AMI in AWS (only for type=ami)
      --aws-bucket string     target S3 bucket name for intermediate storage when creating AMI (only for type=ami)
      --aws-region string     target region for AWS uploads (only for type=ami)
      --config string         build config file
  -h, --help                  help for build
      --output string         artifact output directory (default ".")
      --rpmmd string          rpm metadata cache directory (default "/var/cache/osbuild/rpmmd")
      --store string          osbuild store for intermediate pipeline trees (default ".osbuild")
      --tls-verify            require HTTPS and verify certificates when contacting registries (default true)
      --type string           image type to build [qcow2, ami] (default "qcow2")

2024/01/24 12:17:25 error: failed to create output file "/output/manifest-qcow2.json": open /output/manifest-qcow2.json: operation not permitted
```

which I find a bit ugly and counterintuitive (it implies there was a mistake in the flags and options).

I wont block the PR because this is cleaner, but maybe something to follow up on.

cgwalters:
@achilleas-k yes fair, followup in https://github.com/osbuild/bootc-image-builder/pull/152

PR #146, opened 2024-01-22
many: unbreak integration tests

mvo5:
See the individual commits, this PR fixes three things:
1. asks tmt for more diskspace in general
2. move default /tmp for pytest to /var/tmp so that we do not suffer from tmpfs smallness
3. move to last known good quay.io/centos-bootc/fedora-bootc container before issues https://github.com/CentOS/centos-bootc/issues/184 and https://github.com/osbuild/bootc-image-builder/issues/149

This unblocks "main" but we still need to find a proper solution for (3)

--- 

The tests are currently running out of disk space inside the testing-farm environment. This commit should help debug this by adding `df -h` output after the test is run.

It also asks for 200GB disk space (which is probably way more than we need). Note that the docs say that the size is 250GB by default for things running inside the RH ranch so this might be redunant [1].

[1] https://docs.testing-farm.io/Testing%20Farm/0.1/test-request.html#hardware

### DIFF
--- a/test/containerbuild.py
+++ b/test/containerbuild.py
@@ -24,5 +24,9 @@ def container_to_build_ref():
     # making "image_type" an "image" tuple (type, container_ref_to_test)
     return os.getenv(
         "BIB_TEST_BOOTC_CONTAINER_TAG",
-        "quay.io/centos-bootc/fedora-bootc:eln",
+        # using this tag instead of ":eln" until
+        #  https://github.com/CentOS/centos-bootc/issues/184 and
+        #  https://github.com/osbuild/bootc-image-builder/issues/149
+        # are fixed
+        "quay.io/centos-bootc/fedora-bootc:ed19452a30c50900be0b78db5f68d9826cc14a2e402f752535716cffd92b4445",
     )
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -112,4 +112,4 @@ jobs:
         AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
         AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
       run: |
-        pytest -rs -s -vv
+        pytest -rs -s -vv --basetemp="${TMPDIR}/tmp"
--- a/plans/all.fmf
+++ b/plans/all.fmf
@@ -2,6 +2,10 @@ summary: Run all tests inside a VM environment
 provision:
   how: virtual
   image: fedora:39
+  # 120 means 120GB, disk has to be specificed here, the format from:
+  #   https://tmt.readthedocs.io/en/stable/spec/hardware.html#disk
+  # does not work here or below under "hardware"
+  disk: 120
   hardware:
     virtualization:
       is-supported: true
@@ -17,3 +21,6 @@ prepare:
 execute:
   how: tmt
   script: pytest -s -vv --force-aws-upload
+finish:
+  how: shell
+  script: df -h
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,4 @@
+[pytest]
+# do not use /tmp by default as it may be on a tempfs and our tests can
+# generate 10G images (that full of holes so not really 10G but still)
+addopts = --basetemp=/var/tmp


achilleas-k:
Seems to still be failing with the same error.

achilleas-k:
It's only happening for the AMI, which might be one of or a combination of two differences:
1. The ami is raw so it's much larger (10 GiB).
2. The `/output` volume isn't mapped to the host, so wherever the container store is mounted might not have enough space, even if the system root is big enough.

I'm having trouble finding the output of `df -h`.

mvo5:
> It's only happening for the AMI, which might be one of or a combination of two differences:
> 
>     1. The ami is raw so it's much larger (10 GiB).
> 
>     2. The `/output` volume isn't mapped to the host, so wherever the container store is mounted might not have enough space, even if the system root is big enough.
> 
> 
> I'm having trouble finding the output of `df -h`.
> 
> EDIT: It's in the `finish` stage. Does that also run on failure?

It's a bit sad, I don't know why "finish" is not run, according to the source code comment in https://github.com/teemtee/tmt/blob/main/tmt/steps/finish/__init__.py#L78 it should run even when the previous steps failed. I will debug this some more.

mvo5:
So fun times, the size is still not right but also `fedora-booc:eln` install is currently broken, it does not boot and random services fail randomly, running e.g. `loadkeys` in a debug shell results in `error while loading shared libraries: libc.so.6: cannot open shared object file: permission denied`. Seems selinux related, disabling selinux make the error go away and then dmesg shows a lot of selinux denials related to `{ getattr } `and `{ search }` for `ssh-keygen` and `dev=overlay`. I cannot paste proper logs right now as I only have a qemu GUI window, no serial shell or anything :/

cgwalters:
Hmmm...will look at that

mvo5:
I addressed the disk-full issue with one more patch by ensuring we use /var/tmp - /tmp comes from a 4GB tmpfs so the ENOSPC issue seems to be caused by this. We also get a 40GB disk now which is not the 120GB we asked for but better than the 15GB we had before. I will ask the tmt team if there is some sort of cap on the resources one can request. Progress :)

mvo5:
> Hmmm...will look at that

Please let me know if I can help in any way!

achilleas-k (approved):
Thanks for digging in and fixing this!

cgwalters:
@mvo5 haven't reproduced yet but one thing it likely relates to is https://github.com/CentOS/centos-bootc/commit/e6761c23bc61262b88dfe9fe081252faf7a885dc *combined* with me just releasing a new ostree into rawhide (and hence ELN) which actually turns this on.  It means that `/` is an overlayfs by default.  However, my test instance upgraded to this just fine, though there may be some skew.  Looking.

cgwalters:
Let's track this in https://github.com/CentOS/centos-bootc/issues/184

cgwalters:
OK so turns out it's an issue here :smile: see the latest in https://github.com/CentOS/centos-bootc/issues/184

But...I will back out this change temporarily until we fix things here.

achilleas-k:
> OK so turns out it's an issue here 😄 see the latest in [CentOS/centos-bootc#184](https://github.com/CentOS/centos-bootc/issues/184)
> 
> But...I will back out this change temporarily until we fix things here.

This PR should get in regardless (it's just changing testing infra).

I think we should keep the change in the container and focus on fixing the issue rather than reverting and risking forgetting about it or kicking the can down the road.

cgwalters (approved):

test/containerbuild.py
```diff
@@ -24,5 +24,9 @@ def container_to_build_ref():
     # making "image_type" an "image" tuple (type, container_ref_to_test)
     return os.getenv(
         "BIB_TEST_BOOTC_CONTAINER_TAG",
-        "quay.io/centos-bootc/fedora-bootc:eln",
+        # using this tag instead of ":eln" until
+        #  https://github.com/CentOS/centos-bootc/issues/184 and
+        #  https://github.com/osbuild/bootc-image-builder/issues/149
+        # are fixed
+        "quay.io/centos-bootc/fedora-bootc:ed19452a30c50900be0b78db5f68d9826cc14a2e402f752535716cffd92b4445",
```
cgwalters, line 31: Yeah, we definitely need auto-rotating datestamped tags...

achilleas-k: If/when we get those, let's also have a regular auto-update job here so that changes in the base container are handled in isolated PRs.

PR #165, opened 2024-01-31
Rename `iso` -> `anaconda-iso`

cgwalters:
This PR just renames the `iso` artifact to `anaconda-iso`.

The reason is that I would like to potentially introduce a *different* generic "live iso" that generates *exactly the input container* just as a Live ISO, very similar to how Fedora CoreOS (and one of its parents, the original Container Linux) do it.

This (proposed) `live-iso` would have several use cases, such as always running from a PXE boot.  I also think this is the long term architecture for an ISO, as opposed to the current Anaconda.  It would also be the same as how e.g. Fedora Workstation does it, where the installer is just an app in the system.

Anyways again, this is just a rename.  For compatibility we continue to honor the `iso` as an alias.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ The following image types are currently available via the `--type` argument:
 |-----------------------|---------------------------------------------------------------------------------------|
 | `ami`                 | [Amazon Machine Image](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/AMIs.html) |
 | `qcow2` **(default)** | [QEMU](https://www.qemu.org/)                                                         |
-| `iso`                 | An unattended Anaconda installer that installs to the first disk found.               |
+| `anaconda-iso`        | An unattended Anaconda installer that installs to the first disk found.               |
 
 ## ☁️ Cloud uploaders
 
--- a/bib/cmd/bootc-image-builder/image.go
+++ b/bib/cmd/bootc-image-builder/image.go
@@ -50,7 +50,7 @@ func Manifest(c *ManifestConfig) (*manifest.Manifest, error) {
 	switch c.ImgType {
 	case "ami", "qcow2", "raw":
 		return manifestForDiskImage(c, rng)
-	case "iso":
+	case "anaconda-iso", "iso":
 		return manifestForISO(c, rng)
 	default:
 		return nil, fmt.Errorf("Manifest(): unsupported image type %q", c.ImgType)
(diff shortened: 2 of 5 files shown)


mvo5 (dismissed):
Thank you

noelmiller:
@cgwalters Just to clarify, Are you saying this would be a live ISO that you could use to install to disk using anconda from a live system rather than having an automated install with a kickstart file like it currently is now?

cgwalters:
This PR is just changing the name of the existing artifact.

But yes the new ISO type I am proposing would support what you are talking about.  (Though there's some nontrivial details here, like to have the GUI anaconda I think we should embed it really as a flatpak etc.)

There is however some steps to bridge things here...if we support configuring the kickstart for our existing anaconda-iso (in particular, leaving off things like the partitioning) then I think it automatically kicks anaconda into interactive mode.  (But yeah this is still not what one wants...IMO for honestly any case, but there's just a huge amount of Fedora-derivative inertia around Anaconda as it is going on here)

noelmiller:
> This PR is just changing the name of the existing artifact.
> 
> But yes the new ISO type I am proposing would support what you are talking about. (Though there's some nontrivial details here, like to have the GUI anaconda I think we should embed it really as a flatpak etc.)
> 
> There is however some steps to bridge things here...if we support configuring the kickstart for our existing anaconda-iso (in particular, leaving off things like the partitioning) then I think it automatically kicks anaconda into interactive mode. (But yeah this is still not what one wants...IMO for honestly any case, but there's just a huge amount of Fedora-derivative inertia around Anaconda as it is going on here)

I think the only benefit I see to having an Anaconda interactive mode (allowing for partitioning, user creation, etc.) is it allows bootc-image-builder to support building things like Silverblue or Universal Blue ISO images which would be hugely helpful to the Desktop Linux community. 

Being able to use this tool to build desktops would be so cool!

mvo5 (approved):
Still looks fine to me

mvo5:
> Hm tmt tests fell over with
> 
> > E       ModuleNotFoundError: No module named 'qmp'
> 
> which is surely unrelated?
> 
> Will look...
> 
> (Side note, I am just not a fan of dynamic languages anymore...and tangentially related to this was looking at refactoring the qemu Go wrapper we have in coreos-assembler (see [coreos/coreos-assembler#3712](https://github.com/coreos/coreos-assembler/pull/3712) ) so that in theory it can be shared and reused with podman-machine)

Fwiw, I agree and would love to use rust/go more …

achilleas-k (approved):
LGTM

I think I'd prefer something that includes the word `installer` in it, to signify purpose, but names are hard.

PR #166, opened 2024-01-31
github: switch to GH macos m1 runners

mvo5:
Now that GH provides M1 macOS runners we should use them :) See https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -87,10 +87,7 @@ jobs:
 
   integration-macos:
     name: "Integration macos"
-    # disabled GH runner as it takes ~50min to run this test, self-hosted
-    # is much faster (~15min)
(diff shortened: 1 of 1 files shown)


cgwalters:
> Error: qemu exited unexpectedly with exit code -1, stderr: qemu-system-aarch64: -accel hvf: Error: HV_UNSUPPORTED

Yeah, won't work because we can't do nested virt here.

cgwalters:
Honestly, I think we can get 80% of the testing value by running through the `podman machine` workflow on a Linux host - and even closer if we use aarch64 to do that.  My go-to for this has usually been GCP which supports nested virt, although apparently not with aarch64 yet.

PR #174, opened 2024-02-05
Add Containerfile for building from osbuild and images sources for development

achilleas-k:
New Containerfile that requires build contexts pointing to checked out sources for osbuild/osbuild and osbuild/images to make building development versions of all the components more convenient.

PR includes a README describing how to build the container.

PR #176, opened 2024-02-06
Add selinux-policy-targeted and distribution-gpg-keys

cgwalters:
This is part of reusing the bib container to build base images; rpm-ostree wants to have *a* policy to bootstrap from currently.

For distribution-gpg-keys, it allows `.repo` files to reference those unconditionally.

Also, reuse a "make package lists declarative" model I've used in other places.  It'd be nice to try to semi-standardize this of course, or even better upstream it into dnf.

We're looking at adding a different Containerfile in https://github.com/osbuild/bootc-image-builder/pull/174 and this way it can source the same requirements.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -13,7 +13,8 @@ FROM registry.fedoraproject.org/fedora:39
 # - https://github.com/osbuild/bootc-image-builder/issues/9
 # - https://github.com/osbuild/osbuild/pull/1468
 COPY ./group_osbuild-osbuild-fedora-39.repo /etc/yum.repos.d/
-RUN dnf install -y osbuild osbuild-ostree osbuild-depsolve-dnf podman qemu-img && dnf clean all
+COPY ./package-requires.txt .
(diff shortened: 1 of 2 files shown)


cgwalters:
To be clear my goal is to be able to apply this change to the centos-bootc repo:

```
diff --git a/.github/workflows/build-image.yml b/.github/workflows/build-image.yml
index 13b9623..d500f4f 100644
--- a/.github/workflows/build-image.yml
+++ b/.github/workflows/build-image.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
 
     container:
-      image: quay.io/centos-bootc/builder:latest
+      image: quay.io/centos-bootc/bootc-image-builder:latest
       options: --privileged
 
     strategy:
```

Because just from a mechanics perspective, we're already shipping nearly all the relevant code in this container image anyways right now.

That said, I guess I could try to teach rpm-ostree not to hard require selinux for this case, but I can't imagine there's any problems in having the extra dependencies here for now.

achilleas-k (approved):

PR #148, opened 2024-01-22
test: test both  `quay.io/centos-bootc/{centos,fedora}-bootc:{eln,stream9}`

mvo5:
Now that `bib` supports both fedora and centos bootc we need to start testing both as part of the integration suite.

PR #178, opened 2024-02-07
test: cache images when possible 

mvo5:
Build on top of https://github.com/osbuild/bootc-image-builder/pull/148

This PR adds code to speed up the tests by re-using images as much as possible. Ideally it would always share but because GH runners have constraint diskspace it will cleanup when the free space falls below 1GB (a bit arbitrary but should be a reasonable heuristic).


achilleas-k (approved):
Really nice!

PR #179, opened 2024-02-07
bib: add cross arch building

mvo5:
This commit adds experimental cross arch building.

I.e. you `bootc-image-builder` on a mac and target an `amd64` machine.

The usage would be:
```
sudo podman run ... \
  --type qcow2 \
  --target-arch amd64 \
```
and this is predicated on the idea that the target container is available for both host and target architecture. When that is the case the container for the host architecture is used as a buildroot, the container for the target architecture is the target tree. Note that some stages chroot into the target tree so `qemu-user` is essential for this to work so that target achitecture code can be run transparently (because only very small amounts of code need to run emulated this is pretty fast still).

It will need qemu-user{,-static} installed on the host. Sadly we cannot just install qemu-user into the container because the binfmt_misc system is global and not namespaced. So while adding the binfmt_misc handler from inside the container would work with enough privs it would alter the global machine state and we should not do it.

[This is https://github.com/osbuild/bootc-image-builder/pull/139 without the testing which runs out of diskspace because we need to pull the arm64 container and create an arm64 disk image which is too much for the already tight disk space situation
I will keep working on #139 to see what I can do about cleaning up diskspace but to unblock the podman team we can consider merging this?]

### DIFF
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -115,9 +115,19 @@ func makeManifest(c *ManifestConfig, cacheRoot string) (manifest.OSBuildManifest
 	}
 
 	// resolve container
-	resolver := container.NewResolver(c.Architecture.String())
+	hostArch := arch.Current()
+	resolverNative := container.NewResolver(hostArch.String())
+	resolverTarget := container.NewResolver(c.Architecture.String())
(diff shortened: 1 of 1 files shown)


achilleas-k (approved):
Yes.  Good.  Ship it!

PR #181, opened 2024-02-08
test: test cross building (amd64->aarch64)

mvo5:
This is a continuation of https://github.com/osbuild/bootc-image-builder/pull/139 with the tests and a bit of extra setup in GH so that a disk with more space is used for the image building. That seems to solve the diskspace issue.


achilleas-k (approved):
Really nice!  Great to see it all green too!

PR #182, opened 2024-02-08
bib: avoid duplicating resolvers

mvo5:
When not doing a cross arch build use the same resolver as host and target arch are the same.

This addresses the comment in https://github.com/osbuild/bootc-image-builder/pull/139#discussion_r1479679146


achilleas-k (approved):
:shipit:

PR #187, opened 2024-02-09
Manifest generation unit tests

achilleas-k:
Adding some quick unit tests to check that:
- Manifest generation and serialization doesn't fail with valid configs.
- Manifest generation fails with expected errors with invalid configs.
- Manifest serialization fails with expected errors with invalid resolved sources.
- Serialized manifests contain expected stages.

These are quite basic and don't cover anything that wouldn't be caught by our more comprehensive integration tests, but are useful for catching regressions quickly (especially for rapid iteration during development).

### DIFF
--- a/bib/cmd/bootc-image-builder/main_test.go
+++ b/bib/cmd/bootc-image-builder/main_test.go
@@ -222,8 +222,60 @@ func TestManifestSerialization(t *testing.T) {
 		},
 	}
 
+	baseConfig := getBaseConfig()
 	userConfig := getUserConfig()
 	testCases := map[string]manifestTestCase{
+		"ami-base": {
+			config:     baseConfig,
+			imageType:  "ami",
+			containers: diskContainers,
+			expStages: map[string][]string{
+				"build": {"org.osbuild.container-deploy"},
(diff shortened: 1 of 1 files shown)


cgwalters (commented):
I have no expertise in this so all I can say is this looks superficially sane.  Just hitting Go lint checks.

achilleas-k:
> Just hitting Go lint checks.
Fixed.  Saw those earlier and fixed them but somehow forgot to push -_-

> I have no expertise in this so all I can say is this looks superficially sane. 
Yeah and they don't really do much right now, most of what they will catch is if, for example, something changes in images that causes a manifest call to fail in bib.  Or the most useful thing is if some change drops, for example, the inclusion of the user stage when a user is defined.
I'm hoping this can serve as a foundation for more comprehensive tests in a similar vein.  We don't need to be failing tests whenever a manifest changes for example (something we used to do in composer a long time ago), but it'd be good if we failed (or had to modify) a test when an important stage stops appearing in a pipeline.

cgwalters (approved):

mvo5:
\o/

Issue #184, opened 2024-02-08
rhtap not pushing new images


cgwalters:
```
$ skopeo inspect -n docker://quay.io/centos-bootc/bootc-image-builder  | jq -r .Created
2024-02-07T14:50:36.65773167Z
$
```

And we have a few PRs merged since then.  

Looking at the pipeline I see e.g. [this run](https://console.redhat.com/preview/application-pipeline/workspaces/centos-bootc/applications/bootc-image-builder/commit/e25342b5a265013212cc826f2b8b75794ed13a65) which seems to have built the latest commit but then...it got stuck deploying or something?

cgwalters:
cc @lmilbaum who may know more about this

cgwalters:
Ah the fatal error is

```
[push-snapshot] + cosign copy -f quay.io/redhat-user-workloads/centos-bootc-tenant/bootc-image-builder/bootc-image-builder@sha256:e563f634b5a4fde70d6a5ccd3d47ec010b61a5113fe0adf46a0b8474229172fd 'quay.io/centos-bootc/bootc-image-builder:{floatingTag}'
[push-snapshot] Error: could not parse reference: quay.io/centos-bootc/bootc-image-builder:{floatingTag}
```

cgwalters:
This may be fixed by https://github.com/redhat-appstudio/release-service-catalog/commit/083a86528dc56b0c7379922cb1d6322d7a6532a3

cgwalters:
This was fixed

PR #188, opened 2024-02-09
README.md: Fix a variety of markdown lints

cgwalters:
The real motivation is to just get a new "dummy" commit in to see if https://github.com/osbuild/bootc-image-builder/issues/184 is fixed.


bcl (approved):
Ack

PR #189, opened 2024-02-09
Add ansible tests

schuellerf:
The first trial to add ansible to the tests.

I just saw that the first test "get_distro_version.yml" could also be changed like "check_user.yml" to actively check against a given value. (if we have the distro name in this context)

Then the code is for sure nicer...

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -19,3 +19,23 @@ To install `tmt` on fedora at least those packages are needed:
 ```shell
 sudo dnf install tmt tmt+provision-virtual
 ```
+
+Ansible container
+-----------------
+To execute real tests against the VM ansible is used. Ansible is run from within a container.
+
+⚠️ When changing `execution-environment.yml` you need to rebuild and commit `ansible-container/` as described below
+
+```shell
+ansible-builder create --context ansible-container
+# optionally as this is also done by the test framework
+podman build -t bootc-image-builder-ansible-runner ansible-container/
+```
+
+To avoid having to install `ansible-builder` (via `pip install ansible-builder`), the resulting folder `ansible-container`
+is added to git.
+
+A manual test of the setup (when the Test VM is running) can be:
+```shell
+podman run --rm -ti --net=host -v ./ansible-playbooks:/runner/ansible-playbooks:Z localhost/bootc-image-builder-ansible-runner ansible-playbook -i localhost, ansible-playbooks/get_distro_version.yml
+```
--- a/test/ansible-container/README.md
+++ b/test/ansible-container/README.md
@@ -1,0 +1,4 @@
+# WARNING
+Do not change anything in this directory!
+
+See ["Ansible container" in ../README.md](../README.md#ansible-container) and [../execution-environment.yml](../execution-environment.yml)!
--- a/test/ansible-playbooks/get_distro_version.yml
+++ b/test/ansible-playbooks/get_distro_version.yml
@@ -1,0 +1,14 @@
+---
+- hosts: all
+  gather_facts: yes
+  vars:
+    ansible_password: "{{ lookup('env', 'ANSIBLE_PASSWORD') | default('password') }}"
+    # TODO capture the SSH key from the VM if possible, feed into the ansible container and re-enable host key checking
+    ansible_ssh_common_args: '-o StrictHostKeyChecking=no'
+  tasks:
+    - name: Print distribution
+      debug:
+        msg: "Distribution: {{ ansible_distribution }}"
+    - name: Print version
+      debug:
+        msg: "Version: {{ ansible_distribution_version }}"
--- a/test/test_build.py
+++ b/test/test_build.py
@@ -13,7 +13,7 @@ import pytest
 
 # local test utils
 import testutil
-from containerbuild import build_container_fixture  # noqa: F401
+from containerbuild import build_container_fixture, build_ansible_container_fixture  # noqa: F401
 from testcases import gen_testcases
 from vm import AWS, QEMU
 
@@ -38,6 +38,42 @@ class ImageBuildResult(NamedTuple):
     journal_output: str
     metadata: dict = {}
 
+
+def get_distribution_and_version(test_vm, build_ansible_container, username, password):
+    """Run Ansible playbook to get distribution and version"""
+
+    output = test_vm.execute_ansible(build_ansible_container,
+                                     "./test/ansible-playbooks/get_distro_version.yml",
+                                     username,
+                                     password )
+
+    print(output)
+
+    # Parse output to extract distribution and version
+    distribution = None
+    version = None
+    dist_key = "Distribution:"
+    version_key = "Version:"
+    for line in output.splitlines():
+        if dist_key in line:
+            distribution = line.split(dist_key)[1].strip(" \"'")
+        elif version_key in line:
+            version = line.split(version_key)[1].strip(" \"'")
+
+    return distribution, version
+
+
+def check_user(test_vm, build_ansible_container, username, password, user_to_test):
+    """Run Ansible playbook to get distribution and version"""
+
+    output = test_vm.execute_ansible(build_ansible_container,
+                                     "./test/ansible-playbooks/check_user.yml",
+                                     username,
+                                     password,
+                                     env_vars=[f"BIB_TEST_USER={user_to_test}"])
+
+    print(output)
+    return " changed=0 " in output
 
 @pytest.fixture(scope='session')
 def shared_tmpdir(tmpdir_factory):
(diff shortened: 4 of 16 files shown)


cgwalters (commented):
There's some conceptual overlap between ansible-for-tests and tmt right?

Also there's a lot going on related to testing...see e.g. stuff in https://gitlab.com/bootc-org/tests

I am curious if you've talked to any of the folks there?

test/ansible-container/README.md
```diff
@@ -0,0 +1,4 @@
+# WARNING
+Do not change anything in this directory!
```
mvo5, line 2: There should be a "why" here :) Maybe `Do not change anything in this directory! It is auto-generated and all changes will be overriden` or similar (whatever the deeper reason is)

test/ansible-playbooks/get_distro_version.yml
```diff
@@ -0,0 +1,14 @@
+---
+- hosts: all
+  gather_facts: yes
+  vars:
+    ansible_password: "{{ lookup('env', 'ANSIBLE_PASSWORD') | default('password') }}"
```
mvo5, line 5: Does ansible support some sort of include/shared vars or similar? It seems we need to set `ansible_{password,ssh_common_args}` in all tests so it might be useful to think about how to do it in a single place. OTOH, just two scripts right now and YAGNI (but in this case I'm pretty sure we will need it again).

test/test_build.py
```diff
@@ -39,6 +39,43 @@ class ImageBuildResult(NamedTuple):
     metadata: dict = {}
 
 
+def get_distribution_and_version(test_vm, build_ansible_container, username, password):
+    """Run Ansible playbook to get distribution and version"""
+
+    output = test_vm.execute_ansible(build_ansible_container,
+                                     "./test/ansible-playbooks/get_distro_version.yml",
```
mvo5, line 46: (nipitck) ideally this use `__file__` so that it does not matter from where the test is run

test/test_build.py
```diff
@@ -39,6 +39,43 @@ class ImageBuildResult(NamedTuple):
...
+                                     username,
+                                     password)
+
+    print(output)
```
mvo5, line 50: is this a debug leftover? If not, maybe worth adding some context to the print, e.g. `print(f"output from ansible get_distro_version playboot {output}")` or something

test/test_build.py
```diff
@@ -39,6 +39,43 @@ class ImageBuildResult(NamedTuple):
...
+def check_user(test_vm, build_ansible_container, username, password, user_to_test):
+    """Run Ansible playbook to get distribution and version"""
+
+    output = test_vm.execute_ansible(build_ansible_container,
+                                     "./test/ansible-playbooks/check_user.yml",
+                                     username,
+                                     password,
+                                     env_vars=[f"BIB_TEST_USER={user_to_test}"])
+
+    print(output)
```
mvo5, line 75: same comment as above

test/test_build.py
```diff
@@ -39,6 +39,43 @@ class ImageBuildResult(NamedTuple):
...
+        elif version_key in line:
+            version = line.split(version_key)[1].strip(" \"'")
+
+    return distribution, version
+
+
+def check_user(test_vm, build_ansible_container, username, password, user_to_test):
+    """Run Ansible playbook to get distribution and version"""
+
+    output = test_vm.execute_ansible(build_ansible_container,
```
mvo5, line 69: I know this is meant to grow, but right now it feels very heavy to build an ansible container, run a playbook, get the output and return the user. All of this is essentially: `test_vm.run(f"getent passwd {user_to_test}")`. A bit similar for the get_distro() it does essentially `test_vm.run("cat /etc/os-release")`

test/test_build.py
```diff
@@ -180,13 +217,20 @@ def test_image_is_generated(image_type):
...
     with QEMU(image_type.img_path, arch=image_type.img_arch) as test_vm:
         exit_status, _ = test_vm.run("true", user=image_type.username, password=image_type.password)
         assert exit_status == 0
-        exit_status, output = test_vm.run("echo hello", user=image_type.username, password=image_type.password)
-        assert exit_status == 0
-        assert "hello" in output
+
+        distribution, version = get_distribution_and_version(test_vm, build_ansible_container,
+                                                             image_type.username, image_type.password)
+        assert distribution in ["Fedora", "CentOS"]
```
mvo5, line 227: I wonder if we should test in a more targeted way here, i.e. ensure that we get exactly the distro we want, right now that feels a little bit vague.

test/test_build.py
```diff
@@ -180,13 +217,20 @@ def test_image_is_generated(image_type):
 
 @pytest.mark.skipif(platform.system() != "Linux", reason="boot test only runs on linux right now")
 @pytest.mark.parametrize("image_type", gen_testcases("direct-boot"), indirect=["image_type"])
-def test_image_boots(image_type):
+def test_image_boots(image_type, build_ansible_container):
```
mvo5, line 220: Tests that use the image_type fixture use cached images. I would prefer to just add a new `test_ansible` or similar and not overload the existing test (removing things from it is fine though of course).

test/vm.py
```diff
@@ -80,6 +80,51 @@ def __enter__(self):
     def __exit__(self, exc_type, exc_value, traceback):
         self.force_stop()
 
+    def execute_ansible(self, ansible_container_tag, playbook, username, password, env_vars=[], check=True):
```
mvo5, line 83: I wonder if we should move this out of VM into it's own testutil helper (or something). Right now the VM class is pretty targeted and has a thin profile. Adding this as a class method make it less so and AFAICT this helper can just use the VM and does not need to be part of it?

test/test_build.py
```diff
@@ -39,6 +39,43 @@ class ImageBuildResult(NamedTuple):
...
+    for line in output.splitlines():
+        if dist_key in line:
+            distribution = line.split(dist_key)[1].strip(" \"'")
+        elif version_key in line:
+            version = line.split(version_key)[1].strip(" \"'")
+
+    return distribution, version
+
+
+def check_user(test_vm, build_ansible_container, username, password, user_to_test):
```
mvo5, line 66: The function is called `check_` and the doc string says it `gets ...`. I wonder if the function should be called `get_user` instead?

test/test_build.py
```diff
@@ -39,6 +39,43 @@ class ImageBuildResult(NamedTuple):
...
+    """Run Ansible playbook to get distribution and version"""
```
mvo5, line 67: This is a copy/paste leftover it seems

mvo5 (commented):
Thanks for your PR. The idea of using ansible to test is interesting and opens a lot of potential I think. I am not sure yet what is the best approach yet though.

I added some questions/ideas/suggestions in the code but I have some higher level questions/discussion points (I may be off in many ways as I have not used ansible in a long time but I hope some of it is still useful):

1. Right now the ansible playbooks gather things but the python code asserts. This adds indirections that make it harder to read the code. I.e. If I look at get_distribution_and_version() I need to also go into the playbook. I wonder if instead we should put the playbooks into control of testing, i.e. we would not define a python test and a playboot but instead setup a bunch of vars or configs for ansible and just run the playbook against the image? E.g. an entire "test_ansible.py" that just (re)uses the generated images and runs the test playbook(s) that do all the work, python would just be a thin driver to run the ansible container?
2. I am not a fan of auto-generated code in a repo (for the ansible container) but I guess there is no way around this? Or could we just `dnf install ansible` in our test env and skip ansible tests if no ansible is around? (sorry again for my ignorance) 
3. Right now ansible adds not very much, this will surely change but it makes me wonder if we should park things until we have a better/more comprehensive use case.

I hope I don't sound negative or anything, I really enjoyed reading this PR and thinking about the implications and (re)learn about ansible :) It's super interesting, thanks for this.

schuellerf:
>     1. Right now the ansible playbooks gather things but the python code asserts. This adds indirections that make it harder to read the code. I.e. If I look at get_distribution_and_version() I need to also go into the playbook. I wonder if instead we should put the playbooks into control of testing, i.e. we would not define a python test and a playboot but instead setup a bunch of vars or configs for ansible and just run the playbook against the image? E.g. an entire "test_ansible.py" that just (re)uses the generated images and runs the test playbook(s) that do all the work, python would just be a thin driver to run the ansible container?
I think concept-wise it should work like in the `check_user.yml` where already in ansible it fails. But just in the context of the current test I didn't find the "distro name"  to hand it over to `get_distro_version.yml` (which would then be renamed to "check_*")

>     2. I am not a fan of auto-generated code in a repo (for the ansible container) but I guess there is no way around this? Or could we just `dnf install ansible` in our test env and skip ansible tests if no ansible is around? (sorry again for my ignorance)
In my experience installing "any" ansible shipped with the current distro causes problems and missing features...
but we could easily switch and check when it fails... I also really don't like auto-generated code in git and also the generation of this code is only "hinted" in README and not scripted... :-/
I'll check/change…

>     3. Right now ansible adds not very much, this will surely change but it makes me wonder if we should park things until we have a better/more comprehensive use case.
My point of view is that this is just a generic concept of "choosing/evaluating the testmethod" and the real usecase would be the end of end test (surely living in another repo)
 
> I hope I don't sound negative or anything, I really enjoyed reading this PR and thinking about the implications and (re)learn about ansible :) It's super interesting, thanks for this.

Thanks!

test/ansible-container/README.md
schuellerf: This was the intention of the next line "See …" and people should go read there 🤔
but it's a good input I'll add it if I don't remove the whole directory 😉

test/ansible-playbooks/get_distro_version.yml
schuellerf: yes, is possible, but you got it - just two scripts for now … it will change

test/test_build.py
schuellerf: thanks for the detailed review - the doc string is wrong

test/test_build.py
schuellerf: yes, this is really bad - I just could not come up with a good example where I know all the parameters to check (other than the user check below) and the parameters of the image creation in this test just is "distro and user"

test/README.md
```diff
@@ -19,3 +19,23 @@ To install `tmt` on fedora at least those packages are needed:
...
 ```
+
+Ansible container
+-----------------
+To execute real tests against the VM ansible is used. Ansible is run from within a container.
+
+⚠️ When changing `execution-environment.yml` you need to rebuild and commit `ansible-container/` as described below
+
+```shell
+ansible-builder create --context ansible-container
```
achilleas-k, line 30: Can we add a job that runs this in a github runner and fails if there are changes?  We do similar things with gofmt.

schuellerf:
We (@achilleas-k , @mvo5 and me) decided that bootc-image-builder is not well suited for an experiment like this, not to pollute this repo. So we'll first have a look at the [bootc-org/tests](https://gitlab.com/bootc-org/tests) mentioned before and then either collaborate there or create a separate "image builder test tool" which is able to read image-builder "input" (blueprint / json) and validates the input with a running image. This test should be setup in a way that it can still be used in bootc-image-builder if necessary.

PR #190, opened 2024-02-09
devel/Containerfile: Unify with main Containerfile

cgwalters:
Use `package-requires.txt` here too, this helps
keep things in sync.


bcl (approved):
Looks good

PR #194, opened 2024-02-12
Drain logic from entrypoint into Go code

cgwalters:
This is prep for adding multiple entrypoints; right now the Go code has a `build` verb that is always injected by this wrapper.

Move the default values into the Go code.

A future change will likely make things so that users are required to pass `build` (or perhaps `build-disk`) and this shell script goes away entirely.

### DIFF
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -335,7 +335,7 @@ func run() error {
 		SilenceUsage:          true,
 	}
 	rootCmd.AddCommand(manifestCmd)
-	manifestCmd.Flags().String("rpmmd", "/var/cache/osbuild/rpmmd", "rpm metadata cache directory")
+	manifestCmd.Flags().String("rpmmd", "/rpmmd", "rpm metadata cache directory")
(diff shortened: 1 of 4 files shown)


bib/cmd/bootc-image-builder/main.go
```diff
@@ -344,7 +344,7 @@ func run() error {
 	logrus.SetLevel(logrus.ErrorLevel)
 	buildCmd.Flags().AddFlagSet(manifestCmd.Flags())
 	buildCmd.Flags().String("output", ".", "artifact output directory")
```
cgwalters, line 346: Just to note here, this continues to work because of the new `WORKDIR /output` in the container.

achilleas-k (approved):
LGTM!

FWIW, I'm fine with dropping the entrypoint script in this same PR as well.
Though that would require deciding on the verb and updating docs so it's fine either way.

cgwalters:
Man, the testing farm tests are broken again...I don't understand the error.

mvo5:
> Man, the testing farm tests are broken again...I don't understand the error.

Oh no! Let me have a look

PR #196, opened 2024-02-13
workflow: add workaround for failing tmt tests

mvo5:


cgwalters:
Looks like tmt is fixed :tada: so I think we may not need this one

Issue #149, opened 2024-01-22
root filesystem label is container_file_t when it should be root_t


cgwalters:
Migrating this back from https://github.com/CentOS/centos-bootc/issues/184

Because we want to be able to add a proper `Closes` in this repository.

---

One thing I notice here...and I'm not yet certain if it's a bib regression or not, but looking at the disk image before it's booted:

```
$ guestfish --ro -a disk.qcow2                                                                                                                                                                                                                   
><fs> run
list-filesystems
><fs> list-filesystems
/dev/sda1: unknown
/dev/sda2: vfat
/dev/sda3: ext4
/dev/sda4: ext4
><fs> mount /dev/sda4 /
><fs> getxattrs /
[0] = {
  attrname: security.selinux
  attrval: system_u:object_r:container_file_t:s0\x00
}
><fs> 
```

That's just really broken, we shouldn't end up with a physical disk image root labeled `container_file_t`!  It looks like actually *all* of the labels up to the deployment root are similarly broken (they should be something like `root_t` or `usr_t`).

However once we get to the deployment things are fine:

```
><fs> getxattrs /ostree/deploy/default/deploy/3ef1290eacdb05e50127ed5a920e264f228dae248addb10d98224a2e04918c2c.0/etc/fstab
[0] = {
  attrname: security.selinux
  attrval: system_u:object_r:etc_t:s0\x00
}
><fs> getxattrs /ostree/deploy/default/deploy/3ef1290eacdb05e50127ed5a920e264f228dae248addb10d98224a2e04918c2c.0/etc/passwd 
[0] = {
  attrname: security.selinux
  attrval: system_u:object_r:passwd_file_t:s0\x00
}
><fs> 
```

And it's specifically that `/ostree/deploy/default/backing` is also `container_file_t`, and the overlayfs picks up that context and that breaks everything.

cgwalters:
BTW I don't have much experience with osbuild debugging...if someone can add a brief crash course into the README.md here on e.g. "how can I set a breakpoint right after `org.osbuild.selinux` runs that'd be awesome.  I read through https://www.osbuild.org/guides/developer-guide/index.html but didn't find anything like that.

Edit, ok `osbuild --help` shows a likely `--break` argument, but when I do e.g. `osbuild --break org.osbuild.selinux /tmp/manifest-qcow2.json` gives me...

```
source/org.osbuild.skopeo-index (org.osbuild.skopeo-index): Getting image list signatures
[...]

PR #197, opened 2024-02-13
Ensure root_t label for /store

cgwalters:
The way osbuild works is to synthesize a filesystem tree in the store, then copy it to the disk.  This ensures the label for the store is `root_t` which ends up being the labeling for the "infrastructure" bits in the `/ostree` repository in the target root.

This in turn is blocking a lot of things.

Closes: https://github.com/osbuild/bootc-image-builder/issues/149

### DIFF
--- a/bib/internal/setup/setup.go
+++ b/bib/internal/setup/setup.go
@@ -12,13 +12,21 @@ import (
 
 // EnsureEnvironment mutates external filesystem state as necessary
 // to run in a container environment.  This function is idempotent.
-func EnsureEnvironment() error {
+func EnsureEnvironment(storePath string) error {
 	osbuildPath := "/usr/bin/osbuild"
 	if util.IsMountpoint(osbuildPath) {
 		return nil
 	}
 
-	// A hardcoded security label from Fedora derivatives
+	// Forcibly label the store to ensure we're not grabbing container labels
+	rootType := "system_u:object_r:root_t:s0"
+	// This papers over the lack of ensuring correct labels for the /ostree root
+	// in the existing pipeline
+	if err := util.RunCmdSync("chcon", rootType, storePath); err != nil {
+		return err
+	}
+
+	// A hardcoded security label from Fedora derivatives for osbuild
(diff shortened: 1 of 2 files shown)


bib/internal/setup/setup.go
```diff
@@ -53,6 +53,14 @@ func EnsureEnvironment() error {
 	if err := util.RunCmdSync("mount", "--bind", destPath, osbuildPath); err != nil {
 		return err
 	}
+
+	// And we also forcibly label the store to ensure we're not grabbing container labels
+	rootType := "system_u:object_r:root_t:s0"
```
cgwalters, line 58: One thing I'd like to test as a followup; now that we have a policy in bib since https://github.com/osbuild/bootc-image-builder/pull/176 we can probably just do `chcon --reference /` here instead of hardcoding a context.  And similarly `chcon --reference=/usr/bin/rpm` instead of hardcoding install_t.

cgwalters:
The other thing I did briefly mess around with here is that in theory the existing `org.osbuild.ostree.selinux` stage could grow support for labeling the base dirs, but given that we can hopefully push towards bootc install to-filesystem anyways, I think this short term hack will be fine.

cgwalters:
Thoughts?

achilleas-k (dismissed):
LGTM.

AIUI, this will not be necessary when we switch to using `bootc install` (osbuild/images#462).

cgwalters:
> AIUI, this will not be necessary when we switch to using bootc install (https://github.com/osbuild/images/pull/462).

Probably/hopefully yes, but OTOH I think this is generally right as a generic baseline anyways.

rhatdan:
LGTM

cgwalters:
OK this one just needs a quick re-stamp for approval, I just did a minor tweak to avoid a textual/logical conflict with https://github.com/osbuild/bootc-image-builder/pull/238

mvo5 (dismissed):

cgwalters:
Oops, stray whitespace change still led to a conflict.  Fixed.

PR #198, opened 2024-02-14
Mount /boot ro by default

cgwalters:
See https://github.com/containers/bootc/pull/294
This is particularly motivated by https://github.com/CentOS/centos-bootc-dev/pull/27 because with that suddenly `dnf` will appear to start working but trying to do anything involving the kernel (i.e. mutating `/boot`) will end in sadness, and this puts a stop to that.

(This also relates of course to ye olde https://github.com/osbuild/bootc-image-builder/issues/18
 where we want the partitioning setup in the default case
 to come from the container)

### DIFF
--- a/bib/cmd/bootc-image-builder/partition_tables.go
+++ b/bib/cmd/bootc-image-builder/partition_tables.go
@@ -9,6 +9,11 @@ import (
 const (
 	MebiByte = 1024 * 1024        // MiB
 	GibiByte = 1024 * 1024 * 1024 // GiB
+	// BootOptions defines the mountpoint options for /boot
+	// See https://github.com/containers/bootc/pull/341 for the rationale for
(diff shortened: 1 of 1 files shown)


bib/cmd/bootc-image-builder/partition_tables.go
```diff
@@ -9,6 +9,11 @@ import (
 const (
 	MebiByte = 1024 * 1024        // MiB
 	GibiByte = 1024 * 1024 * 1024 // GiB
+	// BootOptions defines the mountpoint optins for /boot
```
bcl, line 12: `optins` typo

mvo5 (approved):
Thank you

ericcurtin:
I did read some patches a while pack about adding EROFS support to grub, so you can truely have a ro /boot filesystem with better performance, but they are not ready I believe...

PR #200, opened 2024-02-14
add doc on cross-arch builds

sallyom:
Add documentation for running cross-architecture bib

I might have used too many words :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,20 +133,22 @@ Usage:
     <imgref>
 
 Flags:
-      --chown string    chown the ouput directory to match the specified UID:GID
-      --config string   build config file
-      --tls-verify      require HTTPS and verify certificates when contacting registries (default true)
-      --type string     image type to build [qcow2, ami] (default "qcow2")
+      --chown string           chown the ouput directory to match the specified UID:GID
+      --config string          build config file
+      --tls-verify             require HTTPS and verify certificates when contacting registries (default true)
+      --type string            image type to build [qcow2, ami] (default "qcow2")
+      --target-arch string     architecture to build image for (default is the native architecture)
 ```
 
 ### Detailed description of optional flags
 
-| Argument         | Description                                                      | Default Value |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -105,18 +105,20 @@ Usage:
     <imgref>
 
 Flags:
-      --config string   build config file
-      --tls-verify      require HTTPS and verify certificates when contacting registries (default true)
-      --type string     image type to build [qcow2, ami] (default "qcow2")
+      --config string          build config file
+      --tls-verify             require HTTPS and verify certificates when contacting registries (default true)
+      --type string            image type to build [qcow2, ami] (default "qcow2")
+      --target-arch string     architecture to build image for (default is the native architecture)
```
cgwalters, line 111: (Not a new issue but it's unclear to me why we need to duplicate this table)

cgwalters (dismissed):

mvo5 (dismissed):
Thank you! And sorry that this review took forever, it somehow slipped through.

README.md
```diff
@@ -130,6 +132,17 @@ The following image types are currently available via the `--type` argument:
 | `qcow2` **(default)** | [QEMU](https://www.qemu.org/)                                                         |
 | `iso`                 | An unattended Anaconda installer that installs to the first disk found.               |
 
+## 💾 Target architecture
+
+Specify the target architecture of the system on which the disk image will be installed on. By default,
```
mvo5, line 137: I wonder if we should slip in the word "experimental" here somewhere but then we print it on the commandline and while this needs more real world testing we are somewhat confident that the feature will work fine :)

mvo5 (approved):
I rebased this and fixed the conflicts. Sorry that this has not landed yet, I will try to get it in now.

README.md
mvo5: Nevermind, this is no longer experimental it now used for a while.

PR #222, opened 2024-02-29
Add troubleshooting guide

achilleas-k:
Add a troubleshooting guide and a script that builds an image from a bootable container using `bootc install to-disk`, effectively removing BIB from the pipeline.  The guide and script are meant for investigating build or boot issues when it's not clear if the source of the issue is BIB, or something in the bootable container itself.

I'd like for the troubleshooting guide to be extended with more information about what users can try to investigate or work around failures.  We could also later link to https://gitlab.com/bootc-org/podman-bootc-cli with information on how to use it for debugging.

I initially considered linking to the guide from an issue template on GitHub but I think for now we can keep it here and link to it when we want more information from users reporting issues.

@cgwalters I tested the script with the current quay.io/centos-bootc/centos-bootc:stream9 (image ID `f1e476c5f5d61c0eb50f5d5e2e161595d29a8ebc500940d532e90c8fd483b5df`) and it worked but failed to boot with UEFI in qemu.  BIOS boot worked.

Is this something with the way I run the container?

### DIFF
--- a/devel/Troubleshooting.md
+++ b/devel/Troubleshooting.md
@@ -3,6 +3,7 @@
 If your container image fails to build using bootc-image-builder, or if it builds successfully but fails to boot, it's often not clear if the problem lies with bootc-image-builder or the container itself.
 
 To test building an image without bootc-image-builder, try the [bootc-install](devel/bootc-install) script in this repository.
+A more detailed guide for building a disk image using `bootc install` can be found [in the bootc docs](https://github.com/containers/bootc/blob/main/docs/install.md#executing-bootc-install)
 
 **IMPORTANT**
 Before running the script, note that it creates a file called `disk.raw` in the working directory. Make sure this action doesn't overwrite any existing files.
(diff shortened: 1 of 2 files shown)


cgwalters:
> failed to boot with UEFI in qemu. BIOS boot worked.

If you don't pass `--generic-image`, bootc targets the just same boot mode as the host (if your initial VM is BIOS you don't get EFI, and vice versa).  Using `to-disk --via-loopback` automatically enables `--generic-image`.

cgwalters (dismissed):

cgwalters:
It's kind of tempting I guess to assume `--generic-image` when running in a VM and only do the "match the host" on bare metal which nowadays should generally always get you an EFI only system.

achilleas-k:
> (But OTOH any magic we do like that then undercuts the "test baremetal flows via VMs" which gets dangerous)

I agree, I think the current state makes sense.  I did in fact test this in a VM booted through BIOS.  I should update those devel VM scripts :)

achilleas-k:
Script is much simpler now.  Thanks Colin!

ondrejbudai (approved):
Cool!

PR #225, opened 2024-03-01
Allow users to specify multiple types

rhatdan:
When producting a disk image, I want to be able to specify multiple at the same time. This would speed up builds and allow creating of an AMI, ISO and qcow2 all in one run.

### DIFF
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -297,7 +297,7 @@ func cmdBuild(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("chowning is not allowed in output directory")
 	}
 
-	manifest_fname := fmt.Sprintf("manifest-%s.json", imgTypes[0])
+	manifest_fname := fmt.Sprintf("manifest-%s.json", strings.Join(imgTypes, "-"))
 	fmt.Printf("Generating %s ... ", manifest_fname)
 	mf, err := manifestFromCobra(cmd, args)
 	if err != nil {
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -187,6 +187,14 @@ func manifestFromCobra(cmd *cobra.Command, args []string) ([]byte, error) {
 	rpmCacheRoot, _ := cmd.Flags().GetString("rpmmd")
 	targetArch, _ := cmd.Flags().GetString("target-arch")
 	tlsVerify, _ := cmd.Flags().GetBool("tls-verify")
+
+	// translate anaconda-iso to iso to avoid multiple image type checks
+	for idx := range imgTypes {
+		if imgTypes[idx] == "anaconda-iso" {
+			imgTypes[idx] = "iso"
+		}
+	}
+
 	if targetArch != "" {
 		// TODO: detect if binfmt_misc for target arch is
 		// available, e.g. by mounting the binfmt_misc fs into
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -342,10 +342,10 @@ func cmdBuild(cmd *cobra.Command, args []string) error {
 
 	fmt.Println("Build complete!")
 	if upload {
-		for _, imgType := range imgTypes {
+		for idx, imgType := range imgTypes {
 			switch imgType {
 			case "ami":
-				diskpath := filepath.Join(outputDir, imgType, "disk.raw")
+				diskpath := filepath.Join(outputDir, exports[idx], "disk.raw")
 				if err := uploadAMI(diskpath, targetArch, cmd.Flags()); err != nil {
 					return err
 				}
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -222,10 +222,21 @@ func manifestFromCobra(cmd *cobra.Command, args []string) ([]byte, error) {
 		config = &BuildConfig{}
 	}
 
+	// Disk image types should all share mostly the same manifest but with
+	// different export pipelines.
+	// Right now the qcow2 contains all the pipelines required for ami and raw,
+	// so if one of the image types is qcow2, build that and export pipelines
+	// accordingly if needed.
+	// NOTE: THIS WILL CHANGE WITH THE INTRODUCTION OF NEW IMAGE TYPES
+	imgType := imgTypes[0]
+	if slices.Contains(imgTypes, "qcow2") {
+		imgType = "qcow2"
+	}
+
 	manifestConfig := &ManifestConfig{
 		Architecture: buildArch,
 		Config:       config,
-		ImgType:      imgTypes[0],
+		ImgType:      imgType,
 		Imgref:       imgref,
 		Repos:        repos,
 		TLSVerify:    tlsVerify,
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -21,6 +21,7 @@ import (
 	"github.com/osbuild/images/pkg/rpmmd"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+	"golang.org/x/exp/slices"
 )
 
 //go:embed fedora-eln.json
@@ -182,7 +183,7 @@ func manifestFromCobra(cmd *cobra.Command, args []string) ([]byte, error) {
 
 	imgref := args[0]
 	configFile, _ := cmd.Flags().GetString("config")
-	imgType, _ := cmd.Flags().GetString("type")
+	imgTypes, _ := cmd.Flags().GetStringArray("type")
 	rpmCacheRoot, _ := cmd.Flags().GetString("rpmmd")
 	targetArch, _ := cmd.Flags().GetString("target-arch")
 	tlsVerify, _ := cmd.Flags().GetBool("tls-verify")
@@ -193,10 +194,13 @@ func manifestFromCobra(cmd *cobra.Command, args []string) ([]byte, error) {
 		// including tiny statically linked target-arch
 		// binaries inside our bib container
 		fmt.Fprintf(os.Stderr, "WARNING: target-arch is experimental and needs an installed 'qemu-user' package\n")
-		if imgType == "iso" {
+		if slices.Contains(imgTypes, "iso") {
 			return nil, fmt.Errorf("cannot build iso for different target arches yet")
 		}
 		buildArch = arch.FromString(targetArch)
+	}
+	if slices.Contains(imgTypes, "iso") && len(imgTypes) > 1 {
+		return nil, fmt.Errorf("cannot build iso with different target types")
 	}
 	// TODO: add "target-variant", see https://github.com/osbuild/bootc-image-builder/pull/139/files#r1467591868
 
@@ -213,7 +217,7 @@ func manifestFromCobra(cmd *cobra.Command, args []string) ([]byte, error) {
 	manifestConfig := &ManifestConfig{
 		Architecture: buildArch,
 		Config:       config,
-		ImgType:      imgType,
+		ImgType:      imgTypes[0],
 		Imgref:       imgref,
 		Repos:        repos,
 		TLSVerify:    tlsVerify,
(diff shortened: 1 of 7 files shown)


bib/cmd/bootc-image-builder/main.go
```diff
@@ -348,7 +340,7 @@ func run() error {
 	rootCmd.AddCommand(manifestCmd)
 	manifestCmd.Flags().String("rpmmd", "/rpmmd", "rpm metadata cache directory")
 	manifestCmd.Flags().String("config", "", "build config file")
-	manifestCmd.Flags().String("type", "qcow2", "image type to build [qcow2, ami]")
+	manifestCmd.Flags().StringArray("type", []string{"qcow2"}, "image type to build [qcow2, ami]")
```
cgwalters, line 343: s/type/types/ right?

bib/cmd/bootc-image-builder/main.go
```diff
@@ -210,7 +212,7 @@ func manifestFromCobra(cmd *cobra.Command, args []string) ([]byte, error) {
 
 	manifestConfig := &ManifestConfig{
 		Imgref:       imgref,
-		ImgType:      imgType,
+		ImgType:      imgTypes[0],
```
cgwalters, line 215: I am not an osbuild expert but I think we need a distinct manifest for each image.

bib/cmd/bootc-image-builder/main.go
rhatdan: You mean on the description `image types to build,,,`
We still wan to use `--type qcow2 --type raw --type iso`
I think `--type qcow2,raw,iso` also works.

bib/cmd/bootc-image-builder/main.go
rhatdan: Yes I wanted to ask that, I also know very little about it.

rhatdan: From looking at osbuild it says you can specify --export XYZ multiple times, for one manifest. I might be mistaken, and need to call osbuild multiple times ones for each diskimage type. and create multiple manifest.  The documentation is not very clear. Definitely do not want to be pulling the image multiple times from the registry.

achilleas-k: You're right, you can export the result of multiple pipelines.  The ISO build is too different to share any meaningful state with the disk images, but the AMI and QCOW2 basically share the whole manifest, and the QCOW2 has an extra pipeline at the end to convert the image format.  So we can build one manifest and export two artifacts, the `image` and the `qcow2`, and get two outputs from one build job.

rhatdan: Why is it called image and not ami, raw, hyperv, qcow2 ...

As we roll this forward I would prefer the name of the directories and content to match the specified --type?

achilleas-k: > Why is it called image and not ami, raw, hyperv, qcow2 ...

That's just the name of the internal pipeline as it's named in the manifest.  It's plain `image` because the ami is just a raw disk so the pipeline name is generic.

rhatdan: I want the images that are output named to match the image name or let me name the image. I was thinking that 
`BIB quay.io/podman/foobar --type ami --type raw --type hperv --type qcow` I would end up with a directry tree of 

raw/roobar.raw
ami/foobar.ami
hyperv/foobar.hyperv
qcow2/foobar.qcow2

OR


image/roobar.raw
image/foobar.ami
image/foobar.hyperv
image/foobar.qcow2

From a user perspective, I like the second one best, but either can be worked with.

achilleas-k: This can all be handled trivially from inside the code, so nothing serious to block on right now.  Let's do it in a small follow-up.

To clarify, my comments above were descriptions about internal naming which matter in the code, not suggestions for external-facing things like flags or directory names.

Putting everything in the same subdirectory of the output path with appropriate extensions makes total sense.  Also, naming the files after the container name is a nice idea, let's do that (again another follow-up).  I'll open an issue so we don't lose the ideas.

achilleas-k:
I took the liberty of making the single-manifest-multiple-images scenario work and added tests.

achilleas-k:
And resolved conflicts

achilleas-k:
@mvo5 I'd love a review of the pytest changes from you if you have time

rhatdan:
Needs a rebase.

bib/cmd/bootc-image-builder/main.go
```diff
@@ -208,6 +221,17 @@ func manifestFromCobra(cmd *cobra.Command, args []string) ([]byte, error) {
 		config = &BuildConfig{}
 	}
 
+	// Disk image types should all share mostly the same manifest but with
+	// different export pipelines.
+	// Right now the qcow2 contains all the pipelines required for ami and raw,
+	// so if one of the image types is qcow2, build that and export pipelines
+	// accordingly if needed.
+	// NOTE: THIS WILL CHANGE WITH THE INTRODUCTION OF NEW IMAGE TYPES
+	imgType := imgTypes[0]
```
mvo5, line 230: Could this be simplfied to something like:
```go
index 072e132..8b8127e 100644
@@ -228,7 +228,7 @@ func manifestFromCobra(cmd *cobra.Command, args []string) ([]byte, error) {
        // accordingly if needed.
        // NOTE: THIS WILL CHANGE WITH THE INTRODUCTION OF NEW IMAGE TYPES
        imgType := imgTypes[0]
-       if slices.Contains(imgTypes, "qcow2") {
+       if len(imgTypes) > 1 {
                imgType = "qcow2"
        }
```
(or am I missing something?) - because qcow2 is a superset of all the other images we could always export it. It feels slightly easier to read to me because I don't need to think about what it means if one passes raw,ami.

test/test_build.py
```diff
@@ -170,15 +181,20 @@ def del_ami():
     journal_log_path.write_text(journal_output, encoding="utf8")
     bib_output_path.write_text(bib_output, encoding="utf8")
 
-    yield ImageBuildResult(
-        image_type, generated_img, target_arch, username, password,
-        bib_output, journal_output, metadata)
+    results = []
+    for image_type in image_types:
```
mvo5, line 185: I'm slightly uneasy that most of the existing tests now get a list which makes everything a bit more verbose. I wonder if we could just extract most of image_type_fixture into a normal python helper and then just call that helper in a dedicated test (because AIUI we cannot re-use a cached image, right?). And maybe in this test we export all the targets we support ? wdyt?

mvo5 (commented):
Thanks, this looks fine and I like the added tests! Few ideas inline and one suggestion but if this needs to land quickly we could land and push the tests later (AIUI it is important to land soon for Dan?).

bib/cmd/bootc-image-builder/main.go
achilleas-k: I think I prefer to keep it as is because it makes the logic clear and it might avoid confusion when we need to change it in the future (for example when we bring in other clouds).

> I don't need to think about what it means if one passes raw,ami

I think in this case it's better if we do think about it, because the current code doesn't imply that qcow2 is the superset of all possible disk images.  I mean it is right now, but I'm thinking of the (maybe near-) future scenario when it's not.

On the other hand, we could just always create the qcow2 manifest and simply export what we need.

We should probably eventually make a super-manifest of all the cloud/disk image types and always generate that one.  It would look like this
```mermaid
flowchart
build-->ostree-deployment-->image
image-->qcow2
image-->vmdk
image-->vpc
image-->gcp
image-->hyperv
```
and we'd just match image-type args to exports and be done with it.

test/test_build.py
mvo5: [..]
> > (because AIUI we cannot re-use a cached image, right?)
> 
> Not sure what you mean. Why not?

Sorry, what I mean is right now if we say "qcow2" in the fixture and we have it already we never run bib (which makes sense). But for the qcow2+raw case we will not be able to reuse the qcow2 even if it exists. I think this is already encoded in the tests so I'm probably not adding anything new here :)

achilleas-k: This is actually something I'm wondering about.  It's probably trivial but now I'm thinking what if we test one scenario and then break the other.  We can test the multibuild and the single build scenarios by having two cases, `raw+qcow2` and `ami`, but what if we refactor something and somehow break the single `qcow2` case?  We'd never catch it :/

achilleas-k:
The tests work but there's one small issue.  When exporting the `image` pipeline from a `qcow2` manifest, the filename is different than when exporting the `image` pipeline from a `raw` or `ami` manifest.  This is because we don't control the filenames for internal pipelines, just the last one.  It's easy to fix, but we have to do it in osbuild/images.

achilleas-k:
https://github.com/osbuild/images/pull/505

cgwalters:
In general we need to clean up the image stuff and not have 5 copies of the list of supported disk image types, as that will need to grow.  I'm thinking it'd be good to have something like `bootc-image-builder list-supported-outputs` or something that outputs a machine-readable version too that we can use to keep the docs up to date.

bib/cmd/bootc-image-builder/main.go
```diff
@@ -313,14 +339,16 @@ func cmdBuild(cmd *cobra.Command, args []string) error {
...
-			if err := uploadAMI(diskpath, targetArch, cmd.Flags()); err != nil {
-				return err
+		for idx, imgType := range imgTypes {
+			switch imgType {
+			case "ami":
+				diskpath := filepath.Join(outputDir, exports[idx], "disk.raw")
+				if err := uploadAMI(diskpath, targetArch, cmd.Flags()); err != nil {
+					return err
+				}
+			default:
```
mvo5, line 349: (niptick) AFAICT we don't need the default:continue, it's what happy anyway

test/test_build.py
```diff
@@ -288,3 +347,14 @@ def test_iso_installs(image_type):
         vm.start(use_ovmf=True)
         exit_status, _ = vm.run("true", user=image_type.username, password=image_type.password)
         assert exit_status == 0
+
+
+@pytest.mark.parametrize("images", gen_testcases("multidisk"), indirect=["images"])
+def test_multi_build_request(images):
```
mvo5, line 353: \o/

test/test_manifest.py
```diff
@@ -5,17 +5,16 @@
...
     pytest.skip("no podman, skipping integration tests that required podman", allow_module_level=True)
 
 from containerbuild import build_container_fixture  # noqa: F401
 from testcases import gen_testcases
 
 
-@pytest.mark.parametrize("image_type", gen_testcases("manifest"))
-def test_manifest_smoke(build_container, image_type):
-    container_ref = image_type.split(",")[0]
+@pytest.mark.parametrize("images", gen_testcases("manifest"))
```
mvo5, line 15: (super nitpick) I don't think this is strictly needed, the original code is confusing, I will do a followup with a small cleanup.

test/test_build.py
```diff
@@ -52,23 +54,47 @@ def image_type_fixture(shared_tmpdir, build_container, request, force_aws_upload
...
+
     username = "test"
     password = "password"
 
-    # image_type can be long and the qmp socket (that has a limit of 100ish
-    # AF_UNIX) is derrived from the path
-    # so hash the image_type instead of just using it
-    for_image_type = request.param
-    output_path = shared_tmpdir / format(abs(hash(for_image_type)), "x")
+    # params can be long and the qmp socket (that has a limit of 100ish
```
mvo5, line 93: Given that we build them all now the caching can be simplified I think, I will double check and (maybe) do a followup with that

mvo5 (approved):
Thank you, this is very nice!

rhatdan:
Can we get https://github.com/osbuild/bootc-image-builder/pull/251 in then so we can rebase this.  Would love to have this support in container image by Monday for Demo.

mvo5:
> Can we get #251 in then so we can rebase this. Would love to have this support in container image by Monday for Demo.

Sure thing, I will work on this. Note that there is also a assertion failure in one of the tests right now (our tests) and I will fix that too.

mvo5 (dismissed):

mvo5:
> We should add this to README.

Thank you, I added something minimal to the README and happy to do more in a follouwp

ondrejbudai (approved):
Thanks!

mvo5 (approved):
Thank you!

cgwalters (dismissed):

mvo5 (dismissed):

mvo5 (dismissed):

cgwalters:
It looks like this is hung up on MacOS runner capacity?  Oh I see, it's self-hosted?  I'm a bit uncertain if we really need to do MacOS testing at all at least on PRs as opposed to having it be part of a periodic and opt-in per PR.

rhatdan:
I agree running tests in a VM on a MAC is not that important, and should almost never fail, without failing in a linux hosted Linux VM

mvo5 (approved):

achilleas-k:
Thanks for moving this across the finish line!!

PR #227, opened 2024-03-02
Allow caller to specify the --chown UID:GID of the output directory.

rhatdan:


README.md
```diff
@@ -105,18 +105,20 @@ Usage:
     <imgref>
 
 Flags:
+      --chown string    chown the ouput directory to match the specified UID:GID
```
cgwalters, line 108: Isn't the real fix here for osbuild to stop trying to invoke chown() on its own in general?

If we need to chown, I'd advocate for automatically matching the uid/gid of the target output directory by default (I don't think osbuild does that).

rhatdan: The issue here is that we want to run this command on top of a rootless user via sudo or via the API server, and need a way to allow the rootless user to use the disk images after they are created.

```
podman build ...
podman push ...
sudo podman run BIB -v output:/output --chown $UID:$GID 
podman manifest add output/IMAGE.NAME
```

Currently I have hacked up scripts to do `sudo chown`

achilleas-k:
This option should probably interact with the change we made back in https://github.com/osbuild/bootc-image-builder/pull/57.

Right now, we have this logic:

Check if we can `chown` files in the output directory:
https://github.com/osbuild/bootc-image-builder/blob/d7869c8fe6c857829a8eb3d9dc1a71e52b7580ff/bib/cmd/bootc-image-builder/main.go#L266-L269

If not, disable owner preservation in osbuild:
https://github.com/osbuild/bootc-image-builder/blob/d7869c8fe6c857829a8eb3d9dc1a71e52b7580ff/bib/cmd/bootc-image-builder/main.go#L298-L302

We should probably disable the behaviour when `!canChown`, otherwise we'll get an error on `chownR()` (e.g. on macOS).

cgwalters:
We had a realtime chat on this and my takeaway was:

- It makes sense to support this
- However, given that the actual *goal* you had was to add these into OCI artifacts, IMO it would make a lot of sense to support having bib generate those artifacts directly.  However *that* then apparently in your case would get into pulling from rootful storage to rootless

I'd also reiterate here my baseline push that I think we should have Linux users entering our system do one of two things:

- Use podman machine (with rootful!)
- Use rootful for everything; it's important to emphasize that for e.g. production cases we want to support using OCP for this stuff, and there's no "rootless" containers there so the issue doesn't exist

mvo5 (commented):
Thanks Dan, this looks great. It has a conflict and I drafted a small test (attached). If you don't mind we can take care of the conflict and add the test but I don't want to push to your branch without your consent :) 

[0001-test-add-integration-tests-for-new-chown-feature.patch.txt](https://github.com/osbuild/bootc-image-builder/files/14523128/0001-test-add-integration-tests-for-new-chown-feature.patch.txt)

Just for reference the test takes 14s on my machien with a warmed podman (as it reuses the main build container) for all three test cases.

mvo5 (approved):
Thanks, I pushed the conflict resolving and added a small err prefix (thanks to Ondrej) - I think this is fine and we can add tests etc in a followup (I have something lined up).

PR #228, opened 2024-03-02
bootc-image-builder.sh script

ericcurtin:
It's not much, but it makes for a little less typing/copying commands around, so you can run things like:

sudo ./bootc-image-builder.sh -v .:/output --arch aarch64 --type raw quay.io/centos-bootc/fedora-bootc:eln

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,15 +32,9 @@ by adding a volume-mount for the local file as well as the `--config` flag to th
 The following command will create a QCOW2 disk image. First, create `./config.json` as described above to configure user access.
 
 ```bash
-sudo podman run \
-    --rm \
-    -it \
-    --privileged \
-    --pull=newer \
-    --security-opt label=type:unconfined_t \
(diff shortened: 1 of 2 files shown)


cgwalters (changes requested):
We can't do this by default for multiple reasons.

- Inherently such an approach will require the user to manually clone the repository or just curl the script, and make it executable
- But then we need a mechanism to *keep that script up to date* on the user's machine; for coreos-assembler which is a similar project with similar problems recommends copy-pasting a bash alias instead, which is slightly better

But the real fix...we are already developing https://gitlab.com/bootc-org/podman-bootc-cli and I would like to add a `build-disk` verb there.  We will be keeping that project up to date and plan to grow it more into an opinionated bootc-world CLI entrypoint that integrates with podman.

ericcurtin:
Didn't know that podman-bootc-cli was in the works, that seems like the best solution indeed!

PR #235, opened 2024-03-05
test: revert pr#234 (re-enable anaconda-iso test)

mvo5:
This is a reminder/check that we need to re-enable the anaconda iso test. It will fail right now until the iso install is fixed.


mvo5:
Closing as the disable landed in a different way.

PR #238, opened 2024-03-06
bib: mount `devtmpfs` inside the container too

mvo5:
This ensures that the new `partscan` feature in osbuild works. By default the containers only have a static snapshot of /dev on a tmpfs. This means that anything later added by losetup will be missing inside the container.

It also means that https://github.com/osbuild/osbuild/pull/1468 can be reverted.

This should unblock https://github.com/osbuild/bootc-image-builder/pull/120 and also https://github.com/osbuild/images/pull/462

### DIFF
--- a/bib/internal/setup/setup.go
+++ b/bib/internal/setup/setup.go
@@ -53,6 +53,14 @@ func EnsureEnvironment() error {
 	if err := util.RunCmdSync("mount", "--bind", destPath, osbuildPath); err != nil {
 		return err
 	}
+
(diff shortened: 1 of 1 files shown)


cgwalters (approved):

cgwalters:
Tempting to merge over red because we know this can't break the ISO...

cgwalters:
Ah man, shouldn't have clicked the "merge main" button, I thought it was fast forward.  I assume we don't want merge commits in this repo.

PR #239, opened 2024-03-06
workflow: make `pytest` less verbose

mvo5:
This is a bit of an experiment - we currently run `pytest` with `--capture=no` which means we get realtime updates as the tests run in GH actions. This is nice but it also leads to extremly verbose test output which is hard to read because on failure all test output is visible not just the failed one(s).

So this switches to the default mode of pytest to catpure the output of successful tests and only show the output of tests that failed. We lose the ability to view tests in realtime but that is (nowdays) rarely needed (plus it's trivial to run the tests locally).


cgwalters:
I think this is better?  The test output is still pretty verbose though

- dnf package installs are inherently verbose (unnecessarily IMO)
- osbuild is also somewhat chatty in places, like pulling container images
- the qemu console output is verbose
etc.

rhatdan:
LGTM
Anything that can be done to make things less chatty.

mvo5:
> I think this is better? The test output is still pretty verbose though
> 
>     * dnf package installs are inherently verbose (unnecessarily IMO)
> 
>     * osbuild is also somewhat chatty in places, like pulling container images
> 
>     * the qemu console output is verbose
>       etc.

Sorry, I'm a muppet (sometimes), I did not update all our pytest invocations. I did that now and it's much less verbose now. Well, it's still verbose but only for things that fail by default pytest will just swallow all the outputs and hide them now.

cgwalters (dismissed):
Ah indeed, the tests are nearly silent now.

(Almost perhaps *too* silent...it'd be nice to somehow see some key lines to know the tests are not just no-ops :smile: Dunno how hard that is an obviously nonblocking)

rhatdan:
Needs a rebase.

mvo5:
> Ah indeed, the tests are nearly silent now.
> 
> (Almost perhaps _too_ silent...it'd be nice to somehow see some key lines to know the tests are not just no-ops 😄 Dunno how hard that is an obviously nonblocking)

Yeah, agreed. I tweaked it a little bit, I think it might be fine now - sadly there is no way right now to show the last few lines of a successful test with pytest (maybe when I have a few spare minutes could try to fix that upstream).

achilleas-k (approved):

PR #241, opened 2024-03-06
Vendor Go dependencies

cgwalters:
I'm hitting the pain where I want to hack on osbuild/images, and run in a container.
We have the devel containerfile, but it requires some mental energy to use.  Also, for the case where I don't also need to hack on the osbuild binary it's a bit overkill.

Vendoring our Go dependencies makes the *main* Containerfile immediately useful for local builds, I can `go mod replace` with a local dir (just remembering to `go mod vendor`) or even if I just want to truly hack something up I can edit the `vendor/` dir directly.

We don't have to merge this but...I may just keep this around and reuse it and reapply in situations like this. Unsure.


achilleas-k (approved):
I'm in favour of merging this but I'd like Ondrej's approval too before merging.

mvo5:
Fwiw, I'm not a fan of important 1,5 million lines of code into git when there is go.mod,sub - I will not block this though as I see we already do this elsewhere. Still not a fan.

cgwalters:
OK yeah...I am very conflicted on this myself.  We actually have a *very* analogous situation in bootc where there's:

- A Rust binary, that includes some code but
- A Rust package ostree-rs-ext that includes a *lot* of heavy lifting, but not all, and the rest is in
- A C binary and shared library

Sometimes, a change requires hacking on all 3, but a bit more often I just want to update ostree-rs-ext and use that change in bootc.

The analogy here to be clear is:

- bib
- osbuild/images
- osbuild/osbuild

I would definitely not want to vendor ostree-rs-ext in bootc (Rust culture does not generally vendor via git).  Go has a lot longer history here where vendoring was common before go modules were finalized.

Soo...even though I submitted this PR I am neutral on it.  Maybe slightly negative even.  So moving to draft, and if no one argues strongly for this we can close it, and I will look at optimizing the hacking flow a bit more.

achilleas-k:
> I can `go mod replace` with a local dir (just remembering to `go mod vendor`) or even if I just want to truly hack something up I can edit the `vendor/` dir directly.

I guess in the end this also works even when we _don't_ have the deps vendored upstream (it'll just vendor everything the first time), so maybe this isn't a great reason to do it.

For the devel containerfile, I wonder if using the user's `GOMODCACHE` during the build would be a nice solution for quick local builds.  It'll slow down the context setup, but it avoids redownloading all the dependencies each time.  My one worry is root owning things in the user's cache.

cgwalters:
Yeah.  OK let's just close this.  It's not hard to redo if we ever want to.

PR #245, opened 2024-03-07
github: split test setup and the actual run

ondrejbudai:
When I open the log for test run in GitHub actions, I want to see just
the test run, not pip install deps. Let's split these two things
to make the log more readable.


mvo5 (approved):
Thank you!

PR #249, opened 2024-03-07
devel: Add Makefile and Containerfile.hack

cgwalters:
Two conveniences:

- I commonly work on multiple projects in multiple languages and I find it very useful to have typing `make` do something useful by default.
- This `Containerfile.hack` I found is a lot faster and more understandable for the case of just wanting to hack on the Go code including changes to osbuild/images versus what is in `devel/Containerfile` which is much more heavyweight and assumes we're also hacking on osbuild.

### DIFF
--- a/devel/Containerfile.hack
+++ b/devel/Containerfile.hack
@@ -1,0 +1,15 @@
+# This Containerfile is targeted to a specialized path where you are changing
+# pure Go code in this project and you're building on the host system with
+# a compatible userspace (e.g. c9s or fedora).  We just take the already
+# built binary and inject it on top of the main upstream container image.
+#
+# Crucially, using this flow imposes no constraints on your build setup,
+# so that adding e.g. `replace github.com/osbuild/images => ../images`
+# will Just Work.
+#
+# To use this, do e.g.:
+#
+# make && podman build --no-cache -t localhost/bib -v $(pwd)/bin:/srcbin -f devel/Containerfile.hack .
+#
+FROM quay.io/centos-bootc/bootc-image-builder:latest
+RUN install /srcbin/bootc-image-builder /usr/bin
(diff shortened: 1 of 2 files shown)


devel/Containerfile.hack
```diff
@@ -0,0 +1,15 @@
...
+# a compatible userspace (e.g. c9s or fedora).  We just take the already
+# built binary and inject it on top of the main upstream container image.
+#
+# Crucially, using this flow imposes no constraints on your build setup,
+# so that adding e.g. `replace github.com/osbuild/images => ../images`
+# will Just Work.
+#
+# To use this, do e.g.:
+#
+# make && podman build --no-cache -t localhost/bib -v $(pwd)/bin:/srcbin -f devel/Containerfile.hack .
```
rhatdan, line 12: Might need a :z on the volume mount.

rhatdan:
I am a big :100: on the Makefile, everytime I come to this repo, I type make and then remember I have to do ./build.sh

devel/Containerfile.hack
cgwalters: I don't personally because I now always use podman machine even on Linux, which basically obviates that problem

rhatdan: Ok although when we get to virtiofsd it will happen and I think this will not work without machine nor on a Mac and perhaps windows machine, if labeling is supported over the remote connection.

ondrejbudai (approved):

devel/Containerfile.hack
ondrejbudai: Why not just `COPY` instead of mounting the dir? Am I missing something? Let's not block on this, though.

cgwalters: Good question, it's a pattern I copied from elsewhere, I added a rationale in https://github.com/osbuild/bootc-image-builder/pull/277 - does that make sense?

PR #251, opened 2024-03-07
Add support for vmdk

cgwalters:
Requires: https://github.com/osbuild/images/pull/504

---

We want to support people trying this out on e.g. VSphere as that is a popular environment.


rhatdan:
LGTM

rhatdan:
You need to update the README.md file as well.

mvo5:
PR#504 has landed now, was this the only thing that kept it in draft?

mvo5 (dismissed):
This looks fine, I would love to have a test but given that timing is important here I will work on this in a followup

ondrejbudai (changes requested):
We should add the new types into README.


Might be also worth adding a Matrix for which environments are the new types recommended. I suspect many people will want an image for vSphere, but they will not want to think which type to choose, so making the choice for them should help them a lot. Alternatively, we can rename the image type to just vsphere to make it very explicit and make bib well-aligned with the rest of Image Builder portfolio.

cgwalters:
With the code here now, `--type ova` built but in digging in a bit more (I was not very familiar with ova format) I now realize from a glance at the code it's a tarball around 3 files.  And the export was broken (I think I saw there was an `.ovf` file and missed that we needed `.ova), with this on top

```
index 66b1c9d..1aefe2f 100644
@@ -291,7 +291,7 @@ func cmdBuild(cmd *cobra.Command, args []string) error {
        case "vmdk":
                exports = []string{"vmdk"}
        case "ova":
-               exports = []string{"ova"}
+               exports = []string{"archive"}
        case "anaconda-iso", "iso":
                exports = []string{"bootiso"}
        default:
```

I now get:

```
org.osbuild.tar: 2871a252794ddcd301cee5e92cb2de773756ea32f859d14a365b6d3ab0e18ce7 {
  "filename": "disk.ova",
  "format": "ustar",
  "paths": [
    "disk.ovf",
    "disk.mf",
    "disk.vmdk"
  ]
}
tar: disk.ovf: Cannot stat: No such file or directory
tar: disk.mf: Cannot stat: No such file or directory
tar: disk.vmdk: Cannot stat: No such file or directory
tar: Exiting with failure status due to previous errors
```

Any tips on that?

cgwalters:
Well in the interest of forward progress I dropped the `ova` bits for now, so just `vmdk`.

cgwalters:
> Alternatively, we can rename the image type to just vsphere to make it very explicit and make bib well-aligned with the rest of Image Builder portfolio.

Hmm...it's tricky because in theory vmdk and ova are standards that apply outside vsphere too; according to wikipedia they're also used by e.g. virtualbox.

ondrejbudai (approved):
Thanks!

PR #253, opened 2024-03-08
test: add test that ensures "vmdk" images are build (HMS-3758)

mvo5:
The vmdk images cannot be booted easily in qemu so we will just
check that the image is there but do no functional testing.

This is a followup for https://github.com/osbuild/bootc-image-builder/pull/251

### DIFF
--- a/test/test_build.py
+++ b/test/test_build.py
@@ -170,14 +170,12 @@ def build_images(shared_tmpdir, build_container, request, force_aws_upload):
                 image_type, generated_img, target_arch, username, password,
                 bib_output, journal_output))
 
-    # return cached results only if we have exactly the amount of images
-    # requested. the reason is that for multi-images we cannot just return
-    # partial results. And because tests run in random order it maybe that
-    # a multi-image test runs after a single image test that already generated
-    # one of the types the multi images requested.
-    # TODO: rework the whole helper and extract the cache check and the build
-    # into their (tested) functions
-    if len(results) == len(image_types):
+    # Because we always build all image types, regardless of what was requested, we should either have 0 results or all
+    # should be available, so if we found at least one result but not all of them, this is a problem with our setup
+    assert not results or len(results) == len(image_types), \
+        f"unexpected number of results found: requested {image_types} but got {results}"
+
+    if results:
         yield results
         return
 
--- a/test/test_build.py
+++ b/test/test_build.py
@@ -17,7 +17,7 @@ import pytest
 # local test utils
 import testutil
 from containerbuild import build_container_fixture  # noqa: F401
-from testcases import (DIRECT_BOOT_IMAGE_TYPES, INSTALLER_IMAGE_TYPES,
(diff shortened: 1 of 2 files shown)


cgwalters:
IMO we don't need to test all images on every PR, as it will quickly get unsustainable (or at least, costly and slow).

The best pattern I think is to have a "periodic" (daily, or so) that does integration testing, and allow opt-in on specific PRs.  This is the pattern we use in OpenShift for example, where you can type
`/test e2e-gcp` to opt-in to testing on GCP.

cgwalters (commented):
Looks sane to me overall though!

test/test_build.py
```diff
@@ -192,11 +190,13 @@ def build_images(shared_tmpdir, build_container, request, force_aws_upload):
                 # upload forced but credentials aren't set
                 raise RuntimeError("AWS credentials not available (upload forced)")
 
-        # we're either building an iso or all the disk image types
-        if image_types[0] in INSTALLER_IMAGE_TYPES:
-            types_arg = [f"--type={image_types[0]}"]
+        # qemu boot images can be build with a single innovation
```
achilleas-k, line 193:
```suggestion
        # qemu boot images can be build with a single invocation
```
?

test/test_build.py
```diff
@@ -192,11 +190,13 @@ def build_images(shared_tmpdir, build_container, request, force_aws_upload):
...
+        # TODO: ensure that *all* images can be build with the same call, to
```
achilleas-k, line 194:
```suggestion
        # TODO: ensure that *all* images can be built with the same call, to
```

test/testcases.py
```diff
@@ -2,7 +2,16 @@
...
+QEMU_BOOT_IMAGE_TYPES = ("qcow2", "raw")
+
+# supported images that can be booted in a cloud
+CLOUD_BOOT_IMAGE_TYPES = ("ami",)
+
+# images that can *not* be booted directly from qemu
+NON_QEMU_BOOT_IMAGE_TYPES = ("vmdk",)
+
+# images that can *not* be booted directly from qemu
+NON_QEMU_BOOT_IMAGE_TYPES = ("vmdk",)
```
achilleas-k, line 14: Duplicate lines.

achilleas-k (changes requested):
Looks good but I have a few nitpicks below (comments).

achilleas-k (approved):

mvo5:
> IMO we don't need to test all images on every PR, as it will quickly get unsustainable (or at least, costly and slow).
> 
> The best pattern I think is to have a "periodic" (daily, or so) that does integration testing, and allow opt-in on specific PRs. This is the pattern we use in OpenShift for example, where you can type `/test e2e-gcp` to opt-in to testing on GCP.

Yeah, it's a good point, it will become unsustainable at some point, we are hoping to use some tricks to delay the point a little bit though.

achilleas-k (approved):
:shipit:

PR #255, opened 2024-03-11
workflow: cancel in-progress tests on PR updates (HMS-3697)

mvo5:
Similar to the osbuild PR#1636 this comit cancels CI runs for PRs that got updated.

This changes the behavior in a way that whenever a PR gets updated all still-in-progress runs get canceled and new runs get spawned.


ondrejbudai (approved):
Yay!

PR #257, opened 2024-03-12
uploader: test the progress upload output with a regex

achilleas-k:
The length of the progress bar depends on the terminal width so let's make the test matching flexible.


mvo5 (approved):
Love it, thank you!

PR #258, opened 2024-03-12
Use the all-manifest for disk image builds

achilleas-k:
_Requires osbuild/images#512_

The bootc-disk image manifest now always contains all pipelines needed to generate all image types. This way we can build any combination of disk images depending on what we need to export without needing to conditionally add or remove pipelines.

Set the filename as just "disk" since extensions are added automatically for each file.

### DIFF
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -223,17 +223,10 @@ func manifestFromCobra(cmd *cobra.Command, args []string) ([]byte, error) {
 		config = &BuildConfig{}
 	}
 
-	// Disk image types should all share mostly the same manifest but with
-	// different export pipelines.
-	// Right now the qcow2 contains all the pipelines required for ami and raw,
-	// so if one of the image types is qcow2, build that and export pipelines
-	// accordingly if needed.
-	// NOTE: THIS WILL CHANGE WITH THE INTRODUCTION OF NEW IMAGE TYPES
+	// Disk image types all share the same manifest but with different export pipelines.
+	// ISO images can't be built alongside other image types.
+	// Therefore, the first element is enough.
 	imgType := imgTypes[0]
-	if slices.Contains(imgTypes, "qcow2") {
-		imgType = "qcow2"
-	}
-
 	manifestConfig := &ManifestConfig{
 		Architecture: buildArch,
 		Config:       config,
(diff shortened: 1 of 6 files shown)


bib/cmd/bootc-image-builder/main.go
```diff
@@ -223,17 +223,10 @@ func manifestFromCobra(cmd *cobra.Command, args []string) ([]byte, error) {
...
-	// Disk image types should all share mostly the same manifest but with
-	// different export pipelines.
-	// Right now the qcow2 contains all the pipelines required for ami and raw,
-	// so if one of the image types is qcow2, build that and export pipelines
-	// accordingly if needed.
-	// NOTE: THIS WILL CHANGE WITH THE INTRODUCTION OF NEW IMAGE TYPES
+	// Disk image types all share the same manifest but with different export pipelines.
+	// ISO images can't be built alongside other image types.
+	// Therefore, the first element is enough.
 	imgType := imgTypes[0]
```
ondrejbudai, line 229: I think we should refactor this at some point. The manifest config now only needs to know whether to build a disk image, or an ISO, so I think we should decide what's being built right here. Additionally, we should handle in this logic the case when ISO and a disk image was requested (either by erroring out, or rather generating two manifests and running osbuild twice).

ondrejbudai (commented):
Love it. See my comment inline, though, I would love to read your opinion.

bib/cmd/bootc-image-builder/main.go
achilleas-k: I agree.  I considered introducing an enum `disk` vs `iso` and collapsing the image types into those two, replacing the `ImgType` field in the `ManifestConfig` by that type:
https://github.com/achilleas-k/bootc-image-builder/blob/7f0ed3c1aab100115830ce356a6fb3c2a998fba2/bib/cmd/bootc-image-builder/image.go#L29-L32

But I kept flip-flopping on the idea and decided to focus on the functional change instead.

I can do this now if you want.  Otherwise it can be a followup?


> Additionally, we should handle in this logic the case when ISO and a disk image was requested (either by erroring out, or rather generating two manifests and running osbuild twice).

This is already covered further up in the function:
https://github.com/achilleas-k/bootc-image-builder/blob/7f0ed3c1aab100115830ce356a6fb3c2a998fba2/bib/cmd/bootc-image-builder/main.go#L211-L213

mvo5 (dismissed):
Thank you! I like it, I made a inline suggestion that might be a potential followup but the benefits (especially for the speed) justify merging it now IMHO

bib/cmd/bootc-image-builder/main.go
achilleas-k: Sure.  I'll add it to this PR then.

achilleas-k:
@mvo5 Went a slightly different direction from your path but it's very similar.  Take a look!

mvo5 (approved):
This is very nice, I love it!

PR #259, opened 2024-03-12
main: check if local storage is mounted when `--local` flag is used (HMS-3792)

kingsleyzissou:
This PR is a follow up to https://github.com/osbuild/bootc-image-builder/pull/120 and addresses some of the comments on the PR.

The most notable change is switching to use local storage as the default. A helper function is introduced to check if the image is in the local store and, if not, copy it into the store.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,19 +40,20 @@ sudo podman run \
     --security-opt label=type:unconfined_t \
     -v $(pwd)/config.json:/config.json \
     -v $(pwd)/output:/output \
+    -v /var/lib/containers/storage:/var/lib/containers/storage \
     quay.io/centos-bootc/bootc-image-builder:latest \
     --type qcow2 \
     --config /config.json \
     quay.io/centos-bootc/fedora-bootc:eln
 ```
 
+NOTE: local storage is being used by default. If the `--local` flag is not provided, as in the above example,
+the latest image will be pulled into the local storage.
+
 ### Using local containers
 
-To use containers from local container's storage rather than a registry, we need to ensure two things:
-- the container exists in local storage
-- mount the local container storage
-
-Since the container is run in `rootful` only root container storage paths are allowed.
+To skip pulling an image into local storage and use an existing container image, the `--local` flag can be used,
+as below:
 
 ```bash
 sudo podman run \
@@ -63,15 +64,12 @@ sudo podman run \
     --security-opt label=type:unconfined_t \
     -v $(pwd)/config.json:/config.json \
     -v $(pwd)/output:/output \
-    -v /var/lib/containers/storage:/var/lib/containers/storage \
     quay.io/centos-bootc/bootc-image-builder:latest \
     --type qcow2 \
     --config /config.json \
     --local \
     localhost/bootc:eln
 ```
-
-When using the --local flag, we need to mount the storage path as a volume. With this enabled, it is assumed that the target container is in the container storage.
 
 ### Running the resulting QCOW2 file on Linux (x86_64)
 
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 
 	"github.com/osbuild/bootc-image-builder/bib/internal/setup"
+	"github.com/osbuild/bootc-image-builder/bib/internal/util"
 	"github.com/osbuild/images/pkg/arch"
 	"github.com/osbuild/images/pkg/blueprint"
 	"github.com/osbuild/images/pkg/cloud/awscloud"
@@ -29,9 +30,10 @@ import (
 var reposStr string
 
 const (
-	distroName       = "fedora-39"
-	modulePlatformID = "platform:f39"
-	releaseVersion   = "39"
+	distroName            = "fedora-39"
+	modulePlatformID      = "platform:f39"
+	releaseVersion        = "39"
+	containersStoragePath = "/var/lib/containers/storage"
 )
 
 type BuildConfig struct {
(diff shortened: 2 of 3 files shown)


bib/internal/util/util.go
```diff
@@ -26,3 +26,48 @@ func RunCmdSync(cmdName string, args ...string) error {
...
+		fmt.Sprintf("docker://%s", imgref),
+		storageRef,
+	)
+}
+
+func EnsureContainerIsInStore(imgref string) error {
+	driver := "overlay"
+	graphRoot := "/var/lib/containers/storage"
+	runRoot := "/run/containers/storage"
+	storageImgRef := fmt.Sprintf("containers-storage:[%s@%s+%s]%s", driver, graphRoot, runRoot, imgref)
```
cgwalters, line 55: I'm curious why the explicit driver/runroot were needed versus inheriting the defaults.   I don't see a problem with it, just seems worthy of a comment I guess.

bib/internal/util/util.go
```diff
@@ -26,3 +26,48 @@ func RunCmdSync(cmdName string, args ...string) error {
...
+
+	exists, err := checkContainerImageExists(storageImgRef)
+	if err != nil {
+		return fmt.Errorf("Failed to check if container is in local store: %v", err)
+	}
+
+	if exists {
+		fmt.Println("Container image exists in local store")
```
cgwalters, line 63: A print seems unnecessary here to me.

bib/internal/util/util.go
```diff
@@ -26,3 +26,48 @@ func RunCmdSync(cmdName string, args ...string) error {
...
+
+func EnsureContainerIsInStore(imgref string) error {
+	driver := "overlay"
+	graphRoot := "/var/lib/containers/storage"
+	runRoot := "/run/containers/storage"
+	storageImgRef := fmt.Sprintf("containers-storage:[%s@%s+%s]%s", driver, graphRoot, runRoot, imgref)
+
+	exists, err := checkContainerImageExists(storageImgRef)
+	if err != nil {
+		return fmt.Errorf("Failed to check if container is in local store: %v", err)
```
cgwalters, line 59: Here and elsewhere

- Use `%w` and not `%v` for wrapping errors
- Go style does not capitalize the first word (not my favorite, but it is what it is)

bib/internal/util/util.go
```diff
@@ -26,3 +26,48 @@ func RunCmdSync(cmdName string, args ...string) error {
 	}
 	return nil
 }
+
+func checkContainerImageExists(storageRef string) (bool, error) {
+	cmd := exec.Command("skopeo", "inspect", storageRef)
```
cgwalters, line 31: An important best practice with `skopeo inspect` is to use `-n`; otherwise it may enumerate tags in the remote registry which is *expensive* in the general case when dealing with remote registries.

I don't think it matters much here, but just noting.

May help to pass e.g. `-f {{.Digest}}` or something so skopeo knows we don't need much information.

README.md
```diff
@@ -67,11 +44,12 @@ sudo podman run \
...
     --config /config.json \
-    --local \
-    localhost/bootc:eln
+    quay.io/centos-bootc/fedora-bootc:eln
 ```
 
-When using the --local flag, we need to mount the storage path as a volume. With this enabled, it is assumed that the target container is in the container storage.
+Note: we need to mount the storage path as a volume, since using containers-storage is the new default. If the target container is not
+already in the local store, `bootc-image-builder` will attempt to copy it into containers-storage. The `--remote` flag can
+be used to skip this and use the image from the remote registry.
```
cgwalters, line 52: Hmm I think what we should document first is noting the new behavior that sometimes (but definitely not always, one will want to e.g. `podman pull` the updated image first.

Let's just mention that, as one might think to use `--remote` for this, but I think that's more for a distinct build system oriented flow.

cgwalters: We could also of course add a `--pull=` flag, just like `podman run` has.  For the bootc self-install flow, we get that for free because it is just exactly a `podman run` invocation of the container itself.

cgwalters (commented):
Thanks for working on this!  Looks generally sane and should be a nice improvement.

bib/internal/util/util.go
kingsleyzissou: I had made them configurable and then decided against it. I'll tidy this up

bib/internal/util/util.go
kingsleyzissou: Thanks that's great feedback, will fix that up

bib/cmd/bootc-image-builder/main.go
```diff
@@ -422,6 +432,8 @@ func run() error {
 	manifestCmd.Flags().String("rpmmd", "/rpmmd", "rpm metadata cache directory")
 	manifestCmd.Flags().String("target-arch", "", "build for the given target architecture (experimental)")
 	manifestCmd.Flags().StringArray("type", []string{"qcow2"}, "image types to build [qcow2, ami, iso, raw]")
+	manifestCmd.Flags().Bool("remote", false, "use a container image from a remote registry")
```
kingsleyzissou, line 435: @ondrejbudai made a good point, we could maybe remove this flag entirely and just have the single codepath?

bib/cmd/bootc-image-builder/main.go
```diff
@@ -422,6 +432,8 @@ func run() error {
...
+	manifestCmd.Flags().String("containers-store", "/var/lib/containers/storage", "containers-storage directory")
```
kingsleyzissou, line 436: I was meant to remove this. Will update

kingsleyzissou:
Closing in favour of #293

PR #262, opened 2024-03-13
Switch to using containers-storage even without --local

ondrejbudai:
This pull request switches to using containers-storage even without the --local flag.

### DIFF
--- a/bib/cmd/bootc-image-builder/image.go
+++ b/bib/cmd/bootc-image-builder/image.go
@@ -123,15 +123,7 @@ func manifestForDiskImage(c *ManifestConfig, rng *rand.Rand) (*manifest.Manifest
 	mf := manifest.New()
 	mf.Distro = manifest.DISTRO_FEDORA
 	runner := &runner.Linux{}
-	containerSources := []container.SourceSpec{
-		{
-			Source:    c.Imgref,
-			Name:      c.Imgref,
-			TLSVerify: &c.TLSVerify,
-			Local:     true,
-		},
-	}
-	err = img.InstantiateManifestFromContainers(&mf, containerSources, runner, rng)
+	err = img.InstantiateManifestFromContainers(&mf, []container.SourceSpec{containerSource}, runner, rng)
 
 	return &mf, err
 }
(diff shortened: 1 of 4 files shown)


achilleas-k:
I think we're all going to get stuck on the pull vs no-pull decision (i.e. whether we should behave like `--pull=newer`).

> I think that pulling always is not the greatest idea since we switched to a container storage, but it keeps the backward compatibility.

In my mind, running bib is similar to a `podman build` and I *think* I expect it to behave the same way.  I could be thinking about it very differently from how users do though, so maybe always pulling makes sense?  I honestly don't know.

If I had to choose right now, I'd default to only pulling when `missing`.

ondrejbudai:
My thinking is that we will probably want to adjust the behaviour of `--local` and handling the host's container storage in the future (see #259 and comments from Dan in #120), and this will probably be a somewhat breaking change. Thus, this PR aims to be fully backward compatible, so we don't break users more often neccessary.

In other words, I'm absolutely fine with changing the behaviour, but once we know what the final behaviour should be.

cgwalters:
> If I had to choose right now, I'd default to only pulling when missing.

Right; `podman-bootc` also explicitly goes to similar effort to implement pull-if-not-present semantics and I think it makes sense.

If we always pulled it would break offline development, and we'd at least need a flag to not pull.

Unfortunately there's not a `podman pull --policy=if-not-present` or so (a tricky thing AFAIK is often these CLI flags need to be added to docker too in order to avoid divergence).

ondrejbudai:
I agree, Colin, and you can still achieve the same behaviour by using `--local` and mounting in the container storage. The point of this PR is to unify the codepaths, so we can start doing stuff like in #272.

I know that we should unify the behaviour with `podman run` and `podman-bootc`, but I think this needs a bigger discussion, thus I decided not to introduce any functional changes in this PR so we can land it smoothly to unblock other things.

achilleas-k:
> I decided not to introduce any functional changes in this PR so we can land it smoothly to unblock other things.

Good call.  Let's get this merged.

achilleas-k (approved):

cgwalters:
Yep makes sense to me for the record!

rhatdan:
You could do a hack of podman run --pull=missing IMAGE true
podman run --help| grep missing

rhatdan:
Or
podman image exists IMAGE || podman pull IMAGE

PR #272, opened 2024-03-15
bib: respect the root fs type from the container

ondrejbudai:
Needs #262 or an alternative, this is mostly a showcase of how bib can be expanded after we can actually run the container.

Prior to this commit, we always assumed that the root fs is ext4. However, it's possible to redefine this in the container, so let's respect this.

### DIFF
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -298,7 +298,7 @@ func cmdBuild(cmd *cobra.Command, args []string) error {
 	}
 
 	manifest_fname := fmt.Sprintf("manifest-%s.json", strings.Join(imgTypes, "-"))
-	fmt.Printf("Generating %s ... ", manifest_fname)
+	fmt.Printf("Generating manifest %s\n", manifest_fname)
 	mf, err := manifestFromCobra(cmd, args)
 	if err != nil {
 		panic(err)
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strconv"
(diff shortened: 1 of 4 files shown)


achilleas-k (approved):
:rocket: Finally!!

ondrejbudai:
ftr, this needs tests and I also want to implement this properly in `images`, this is mostly a showcase for now. :)

bib/cmd/bootc-image-builder/main.go
```diff
@@ -100,7 +102,59 @@ func loadConfig(path string) (*BuildConfig, error) {
...
+	}
+
+	if err := json.Unmarshal(buffer.Bytes(), &bootcConfig); err != nil {
+		return "", fmt.Errorf("failed to unmarshal bootc configuration: %w", err)
+	}
+
+	// filesystem.root.type is the preferred way instead of the old root-fs-type top-level key.
+	// See https://github.com/containers/bootc/commit/558cd4b1d242467e0ffec77fb02b35166469dcc7
+	fsType := bootcConfig.Filesystem.Root.Type
+	supportedFS := []string{"ext4", "xfs"}
```
cgwalters, line 129: I'm curious about the allowlist here; it seems distribution dependent at the minimum right?  E.g. some distros/OSes use btrfs, there's also bcachefs coming in, etc.

Maybe it'd be better to have a denylist instead, i.e. not allow doing `vfat` for example.  We could I guess do a sanity check that `mkfs.$type` exists in the root I guess to give a better error perhaps?

mvo5: I wonder if this is the right place for policy like this (i.e. I would rather just remove it). It seems to me the best place for this would be something like a "bootc lint" or "bootc build" that does checks. If we add the policy here we need to duplicate it accross `bootc install to-disk` and `anaconda` so having it in one central place seems more maintainable to me?

ondrejbudai:
This was moved to #383.

PR #277, opened 2024-03-16
build: Add .dockerignore and a comment

cgwalters:
It's best practice to trim out stuff we don't need in the container build from the context directory; this way if I e.g. edit `tests` it won't invalidate the build caches.

Similarly, it's a really good idea to exclude things like `bin/` from the builds - you don't want to accidentally somehow be using incorrect cached *final* binaries.

But `Containerfile.hack` is all about bypassing that, so add a comment for why we use the bind mount.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+bin
+devel
+plans
+test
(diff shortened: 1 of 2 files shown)


cgwalters:
Blah, testing farm hitting ENOSPC again.

rhatdan:
LGTM

cgwalters:
@rhatdan mind adding an official GH review approval?

ondrejbudai (approved):

PR #286, opened 2024-03-19
bib: connect stdout/stderr when doing the initial podman pull

mvo5:
This should help with the potential UX issue that a podman pull can take a while and for the user it looks like it's hanging.

### DIFF
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -144,6 +144,13 @@ func makeManifest(c *ManifestConfig, cacheRoot string) (manifest.OSBuildManifest
 	// We might want to change this behaviour in the future to match podman.
 	if !c.Local {
 		pullCmd := exec.Command("podman", "pull", "--arch", c.Architecture.String(), c.Imgref)
+		// This is a bit of a hack, we cannot use "stdout"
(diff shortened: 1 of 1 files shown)


bib/cmd/bootc-image-builder/main.go
```diff
@@ -110,6 +110,8 @@ func makeManifest(c *ManifestConfig, cacheRoot string) (manifest.OSBuildManifest
 	// We might want to change this behaviour in the future to match podman.
 	if !c.Local {
 		pullCmd := exec.Command("podman", "pull", "--arch", c.Architecture.String(), c.Imgref)
+		pullCmd.Stdout = os.Stdout
```
ondrejbudai, line 113: This will break the manifest subcommand, I learned this the hard way. However, `stderr` should be just fine, right?

mvo5: Excellent point!

mvo5: I updated this now and also added a short comment about an alternative approach (move it into manifestFromCobra and check what command is used)

mvo5:
Closing for now as we now print a lot message when pulling the image (which I think is not printed by default but that is a separate issue)

PR #288, opened 2024-03-19
doc:add --cpu host in virt-install cli

### LINKED ISSUE #287: Sample virt-install invocation for eln requires explicit CPU flag
I am trying generate an image  straight-up from readme example:

```
sudo podman run \
    --rm \
    -it \
    --privileged \
    --pull=newer \
    -v $(pwd)/config.json:/config.json \
    -v $(pwd)/output:/output \
    --type qcow2 \
    --config /config.json \
```
even tried same with : quay.io/centos-bootc/centos-bootc:stream9

On trying to boot I get : 
```
 Error : kernel panic: not syncing: Attempted to kill init! exit code =0x00007f00
```
image

say-paul:
fixes #287

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,7 @@ qemu-system-x86_64 \
 ```bash
 sudo virt-install \
     --name fedora-bootc \
+    --cpu host \
     --vcpus 4 \
     --memory 4096 \
     --import --disk ./output/qcow2/disk.qcow2,format=qcow2 \


cgwalters:
(Personally I think telling people to copy-paste qemu invocations is unsustainable and we should drop these and refer to tools maintained elsewhere, but yep)

cgwalters:
Hmm actually, we may want `--os-variant rhel9` here? I think that will help ensure the right baseline?

ondrejbudai:
@say-paul Can you check that `--os-variant rhel9` also fixes the issue? I agree that this would be preferred.

say-paul:
tried it with `rhl9`, `rhel9-unknown` , but ended up with issue #287

cgwalters (approved):
I think there's a better doc on this but see https://www.spinics.net/lists/centos-devel/msg22018.html

I'd expect that `--os-variant fedora-eln` would work here too.  But this fine.

PR #290, opened 2024-03-19
bib: make /var/lib/containers/storage a volume

ondrejbudai:
6e965d2413f7bdc8844bf09c9c9968c6c170d96d changed bib to always go
through a container storage. However, if the container runtime running
bib was using overlayfs (the default), we ended up in a situation that
/var/lib/containers/storage was overlayfs. When bib itself tried to
worked with these containers, it called podman/skopeo that tried to
create overlayfs inside overlayfs, which is a forbidden combination.

This worked by accident with centos-boocs/fedora and centos containers,
because they contain fuse-overlayfs, and podman/skopeo apparently
silently switched to using fuse. This isn't only slow, but it also
didn't work with container images without fuse-overlayfs. One example
of such an image is the iot-bootable-container. E.g.
registry.gitlab.com/redhat/services/products/image-builder/ci/images/iot-bootable-container:fedora-39-x86_64-49d623cc26287730f87d4c9eebadefd2b180dea8a41b00efe7f3b1c636f221d7

Let's fix this by making /var/lib/containers/storage a proper volume,
which means that this directory will be bind-mounted to a real
filesystem by the container engine, instead of being on an overlay.

I tested this by successfully building a disk image from the
iot-bootable-container container image above.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -25,3 +25,4 @@ VOLUME /output
 WORKDIR /output
 VOLUME /store
 VOLUME /rpmmd
+VOLUME /var/lib/containers/storage


achilleas-k (approved):

cgwalters:
I'm a bit confused, with `-v /var/lib/containers/storage:/var/lib/containers/storage` in the container run invocation we end up with a pure bind mount to the host, so that shouldn't be an overlayfs.  Are you sure the setup here isn't somehow omitting that option?

ondrejbudai:
It's still valid to call bib without `-v /var/lib/containers/storage:/var/lib/containers/storage`. Thus, we need to make sure that this directory is a "real filesystem", and not an overlay.

cgwalters:
Ah OK I had missed this bit: https://github.com/osbuild/bootc-image-builder/pull/262/commits/1498632a8b8d977578e8501f98a47dadecc5fc61#diff-cbdce8fdffa710f25830f44fc4b0d5be917240f81c72e59e7003dd62fb62349cR105

So...fair enough.  It'd be more code but perhaps conceptually clearer if in this case we did one of these in the entrypoint:

- `ln -sf /store/containers-storage /var/lib/containers/storage`
- `mount -t tmpfs tmpfs /var/lib/containers`

etc.

The way I see it is for this case if I'm explicitly caching the osbuild `store` as a distinct volume, I would expect it to hold cached external containers too instead of having to juggle a distinct mount.

rhatdan:
Belated LGTM

PR #291, opened 2024-03-19
main: double container size as the min size of "/" (HMS-3810)

mvo5:
As a starting point for the image size we use the container size and just double it.

### DIFF
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -32,6 +32,12 @@ const (
 	distroName       = "fedora-39"
 	modulePlatformID = "platform:f39"
 	releaseVersion   = "39"
+
+	// As a baseline heuristic we double the size of
+	// the input container to support in-place updates.
+	// This is planned to be more configurable in the
+	// future.
+	containerSizeToDiskSizeMultiplier = 2
 )
 
 type BuildConfig struct {
@@ -101,6 +107,7 @@ func loadConfig(path string) (*BuildConfig, error) {
 	return &conf, nil
 }
 
+// getContainerSize returns the size of an already pulled container image in bytes
 func getContainerSize(imgref string) (uint64, error) {
 	var stdout, stderr strings.Builder
 
@@ -136,7 +143,7 @@ func makeManifest(c *ManifestConfig, cacheRoot string) (manifest.OSBuildManifest
 		return nil, fmt.Errorf("cannot get container size: %w", err)
 	}
 	c.Filesystems = []blueprint.FilesystemCustomization{
-		{Mountpoint: "/", MinSize: cntSize * 2},
+		{Mountpoint: "/", MinSize: cntSize * containerSizeToDiskSizeMultiplier},
 	}
 
 	manifest, err := Manifest(c)
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -101,6 +101,23 @@ func loadConfig(path string) (*BuildConfig, error) {
 	return &conf, nil
 }
 
+func getContainerSize(imgref string) (uint64, error) {
+	var stdout, stderr strings.Builder
+
+	cmd := exec.Command("podman", "image", "inspect", imgref, "--format", "{{.Size}}")
+	cmd.Stdout = &stdout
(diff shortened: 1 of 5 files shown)


bib/cmd/bootc-image-builder/main.go
```diff
@@ -114,6 +130,13 @@ func makeManifest(c *ManifestConfig, cacheRoot string) (manifest.OSBuildManifest
 			return nil, fmt.Errorf("failed to pull container image: %w", err)
 		}
 	}
+	cntSize, err := getContainerSize(c.Imgref)
+	if err != nil {
+		return nil, fmt.Errorf("cannot get container size: %w", err)
+	}
+	c.Filesystems = []blueprint.FilesystemCustomization{
+		{Mountpoint: "/", MinSize: cntSize * 2},
```
cgwalters, line 138: I'm not fully following the logic under `disk.NewPartitionTable`; will the `/` filesystem still be grown in this scenario to fit the overall disk?

cgwalters: (And if this filesystem is larger than `DEFAULT_SIZE`, does it go past `DEFAULT_SIZE`?)

mvo5: The PartitionTable.relayout comments says the following:
```
// Dynamically calculate and update the start point for each of the existing
// partitions. Adjusts the overall size of image to either the supplied
// value in `size` or to the sum of all partitions if that is larger.
// Will grow the root partition if there is any empty space.
// Returns the updated start point.
```
so unless I misread this it should DTRT *but* I will write some tests to ensure that this is really the case to ensure we do not regress here.

bib/cmd/bootc-image-builder/main.go
```diff
@@ -101,6 +101,23 @@ func loadConfig(path string) (*BuildConfig, error) {
 	return &conf, nil
 }
 
+func getContainerSize(imgref string) (uint64, error) {
```
cgwalters, line 104: Very minor nit but I have tried in code I've been writing recently to add a docstring even for functions like this.  This is maybe a good example because when talking about sizes the units are important, and while it'd be surprising if the units were anything other than bytes...probably still worth stating?

(One thing I don't understand is why Go has the convention of repeating the type name, it makes quick doc comments more annoying to write)

```suggestion
// getContainerSize returns the size of an already pulled container image in bytes
func getContainerSize(imgref string) (uint64, error) {
```

bib/cmd/bootc-image-builder/main.go
```diff
@@ -114,6 +131,13 @@ func makeManifest(c *ManifestConfig, cacheRoot string) (manifest.OSBuildManifest
 			return nil, fmt.Errorf("failed to pull container image: %w", err)
 		}
 	}
+	cntSize, err := getContainerSize(c.Imgref)
+	if err != nil {
+		return nil, fmt.Errorf("cannot get container size: %w", err)
+	}
+	c.Filesystems = []blueprint.FilesystemCustomization{
+		{Mountpoint: "/", MinSize: cntSize * 2},
```
cgwalters, line 139: One other quick thing here, also something like:

```suggestion
	        // As a baseline heuristic we double the size of the input container to
	        // support in-place updates.  This is planned to be more configurable in the future.
		{Mountpoint: "/", MinSize: cntSize * 2},
```

or so?

cgwalters (dismissed):

bib/cmd/bootc-image-builder/main.go
cgwalters: Or perhaps even better a `const containerSizeToDiskSizeMultiplier = 2` at the top with a comment like that?

bib/cmd/bootc-image-builder/main.go
mvo5: Excellent suggestion, added

bib/cmd/bootc-image-builder/main.go
mvo5: Another great point, fixed.

cgwalters (approved):

mvo5:
Sorry for the force pushes, I cleaned the commits up a bit and I think this is fnally ready and tests are also happy.

achilleas-k (approved):

PR #297, opened 2024-03-21
test: map /var/lib/containers/storage when running bib

mvo5:
Now that we always pull the containers this may speed up things when layers are found in the container storage.

I finally measured this, at least locally it has a noticable effect. I tested a single run of 
```
$ sudo pytest -s -vv './test/test_build.py::test_image_boots[quay.io/centos-bootc/centos-bootc:stream9,raw]'
```
- with the patch a single test run  takes `2:39min` on my system
- without `4:40min`

### DIFF
--- a/test/test_build.py
+++ b/test/test_build.py
@@ -233,6 +233,7 @@ def build_images(shared_tmpdir, build_container, request, force_aws_upload):
             "podman", "run", "--rm",
             "--privileged",
             "--security-opt", "label=type:unconfined_t",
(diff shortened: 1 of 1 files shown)


cgwalters:
I guess I still think we *always* want to default to `--local`, in which case yes we should do this too by default.

rhatdan:
I agree, --local=true should be default.  If I have the image in local storage, then use it, otherwise pull it.

ondrejbudai (approved):

PR #298, opened 2024-03-21
main: check s3 bucket access before starting (HMS-3099)

### LINKED ISSUE #220: Check that the given S3 bucket exists before building an image
It's super cheap, and it would improve the UX so much. :)

schuellerf:
Check AWS S3 bucket and permission before starting the build.

Resolves #220 

PR https://github.com/osbuild/images/pull/539 needs to be merged & deployed **before**

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -1,6 +1,8 @@
 FROM registry.fedoraproject.org/fedora:39 AS builder
 RUN dnf install -y git-core golang gpgme-devel libassuan-devel && mkdir -p /build/bib
 COPY bib/go.mod bib/go.sum /build/bib
+ARG GOPROXY=https://proxy.golang.org,direct
+RUN go env -w GOPROXY=$GOPROXY
 RUN cd /build/bib && go mod download
 COPY build.sh /build
 COPY bib /build/bib
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,36 @@
-# There's no heavy lifting here; this file is just a convenience
-# for people whose muscle memory for building projects starts with typing `make`.
-all:
+.PHONY: all
+all: build-binary build-container
+
+.PHONY: clean
+clean:
+	# not sure if we should remove generated stuff
+	# keep the output directory itself
+	#-rm -rf output/*
+	rm -rf bin
+	@echo "removing test files that might be owned by root"
+	sudo rm -rf /var/tmp/bib-tests
+
+.PHONY: test
+test:
+	@echo "Be aware that the tests take a really long time"
+	@echo "Running tests as root"
+	sudo -E pip install --user -r test/requirements.txt
+	sudo -E pytest -s -v
+
+.PHONY: build-binary
+build-binary:
 	./build.sh
+
+.PHONY: build-container
+build-container:
+	sudo podman build --tag bootc-image-builder .
+
+.PHONY: push-check
+push-check: build-binary build-container test
+	cd bib; go fmt ./...
+	@if [ 0 -ne $$(git status --porcelain --untracked-files|wc -l) ]; then \
+	    echo "There should be no changed or untracked files"; \
+	    git status --porcelain --untracked-files; \
+	    exit 1; \
+	fi
+	@echo "All looks good - congratulations"
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -313,16 +313,42 @@ func cmdBuild(cmd *cobra.Command, args []string) error {
 		if !slices.Contains(imgTypes, "ami") {
 			return fmt.Errorf("aws flags set for non-ami image type (type is set to %s)", strings.Join(imgTypes, ","))
 		}
-		// initialise the client to check if the env vars exist before building the image
+
+		// check as many permission prerequisites as possible before starting
 		client, err := awscloud.NewDefault(region)
 		if err != nil {
 			return err
 		}
 
-		fmt.Printf("Checking AWS permission by listing regions...\n")
-		if _, err := client.Regions(); err != nil {
+		fmt.Println("Checking AWS region access...")
+		regions, err := client.Regions()
+		if err != nil {
+			return fmt.Errorf("retrieving AWS regions for '%s' failed: %w", region, err)
+		}
+
+		if !slices.Contains(regions, region) {
+			return fmt.Errorf("given AWS region '%s' not found", region)
+		}
+
+		fmt.Println("Checking AWS bucket...")
+		bucketName, _ := cmd.Flags().GetString("aws-bucket")
+		buckets, err := client.Buckets()
+		if err != nil {
+			return fmt.Errorf("retrieving AWS list of buckets failed: %w", err)
+		}
+		if !slices.Contains(buckets, bucketName) {
+			return fmt.Errorf("bucket '%s' not found in the given AWS account", bucketName)
+		}
+
+		fmt.Println("Checking AWS bucket permissions...")
+		writePermission, err := client.CheckBucketPermission(bucketName, awscloud.S3PermissionWrite)
+		if err != nil {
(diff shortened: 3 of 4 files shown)


bib/cmd/bootc-image-builder/main.go
```diff
@@ -313,16 +313,51 @@ func cmdBuild(cmd *cobra.Command, args []string) error {
...
 		client, err := awscloud.NewDefault(region)
 		if err != nil {
 			return err
 		}
 
-		fmt.Printf("Checking AWS permission by listing regions...\n")
-		if _, err := client.Regions(); err != nil {
-			return err
+		fmt.Printf("Checking AWS region access...\n")
+		var regions []string
```
mvo5, line 324: It's common in go to just avoid this var declaration as the call below will create `regions` and infer the type, so:
```go
-               var regions []string
-               regions, err = client.Regions();
+               regions, err := client.Regions()
```
should be fine here.

bib/cmd/bootc-image-builder/main.go
```diff
@@ -313,16 +313,51 @@ func cmdBuild(cmd *cobra.Command, args []string) error {
...
+		if !regionAvailable {
+			return fmt.Errorf("given AWS region '%s' not found", region)
+		}
+
+		fmt.Printf("Checking AWS bucket...\n")
+		bucketName, _ := cmd.Flags().GetString("aws-bucket")
+		var buckets []string
+		buckets, err = client.Buckets()
+		if err != nil { return fmt.Errorf("retrieving AWS list of buckets failed: %w", err) }
+		foundBucket := false
```
mvo5, line 343: We should be able to use `slices.Contains()` here because we are on go 1.20 and import `golang.org/x/exp/slices`:
```go
-               var buckets []string
-               buckets, err = client.Buckets()
-               if err != nil { return fmt.Errorf("retrieving AWS list of buckets failed: %w", err) }
-               foundBucket := false
-               for _, b := range buckets {
-                       if bucketName == b {
-                               foundBucket = true
-                       }
+               buckets, err := client.Buckets()
+               if err != nil {
+                       return fmt.Errorf("retrieving AWS list of buckets failed: %w", err)
                }
-               if ! foundBucket {
+               if !slices.Contains(buckets, bucketName) {
                        return fmt.Errorf("bucket '%s' not found in the given AWS account", bucketName)
                }
```

bib/cmd/bootc-image-builder/main.go
```diff
@@ -313,16 +313,51 @@ func cmdBuild(cmd *cobra.Command, args []string) error {
...
-		if _, err := client.Regions(); err != nil {
-			return err
+		fmt.Printf("Checking AWS region access...\n")
+		var regions []string
+		regions, err = client.Regions();
+		if err != nil {
+			return fmt.Errorf("retrieving AWS regions for '%s' failed: %w", region, err)
+		}
+
+		regionAvailable := false
```
mvo5, line 330: We should be able to use `slices.Contains()` here (just like below):
```go
-               regionAvailable := false
-               for _, r := range regions {
-                       if r == region { regionAvailable = true }
-               }
-               if !regionAvailable {
+               if !slices.Contains(regions, region) {
                        return fmt.Errorf("given AWS region '%s' not found", region)
                }
```

bib/cmd/bootc-image-builder/main.go
```diff
@@ -313,16 +313,51 @@ func cmdBuild(cmd *cobra.Command, args []string) error {
...
+		for _, r := range regions {
+			if r == region { regionAvailable = true }
+		}
+		if !regionAvailable {
+			return fmt.Errorf("given AWS region '%s' not found", region)
+		}
+
+		fmt.Printf("Checking AWS bucket...\n")
+		bucketName, _ := cmd.Flags().GetString("aws-bucket")
+		var buckets []string
```
mvo5, line 340: Same as above, declaring this here is uncommon in go, just doing:
```go
-               var buckets []string
-               buckets, err = client.Buckets()
+               buckets, err := client.Buckets()
```
should be fine and feels more idiomatic

mvo5 (commented):
Just some really quick drive-by comments for your consideration. Also `go fmt` is needed or our linters will be unhappy (or should be if they are not already :)

bib/cmd/bootc-image-builder/main.go
schuellerf: but re-declares "err" :-/

Containerfile
```diff
@@ -1,6 +1,7 @@
 FROM registry.fedoraproject.org/fedora:39 AS builder
 RUN dnf install -y git-core golang gpgme-devel libassuan-devel && mkdir -p /build/bib
 COPY bib/go.mod bib/go.sum /build/bib
+RUN go env -w GOPROXY=https://proxy.golang.org,direct
```
cgwalters, line 4: Non-blocking/optional: Fully supportive of this change, that said in the general case I think I would vote to make this an `ARG GOPROXY=https://proxy.golang.org` instead so it can be overridden by downstream build systems where desired.

Makefile
```diff
@@ -1,4 +1,39 @@
 # There's no heavy lifting here; this file is just a convenience
```
cgwalters, line 1: Note the changes below make this comment sort of a lie now :smile: 

Maybe say: "Avoid adding substantial dedicated code here, more than 5 lines of inline shell script for a rule should invoke an external tool"?

bib/cmd/bootc-image-builder/main.go
```diff
@@ -313,16 +313,45 @@ func cmdBuild(cmd *cobra.Command, args []string) error {
...
+		regions, err = client.Regions()
+		if err != nil {
+			return fmt.Errorf("retrieving AWS regions for '%s' failed: %w", region, err)
+		}
+
+		if !slices.Contains(regions, region) {
+			return fmt.Errorf("given AWS region '%s' not found", region)
+		}
+
+		fmt.Printf("Checking AWS bucket...\n")
```
cgwalters, line 334: Minor/optional and not new, but this could be `fmt.Println()` and avoid the `\n`.

(Rust of course got this right, the `println!()` macro is saner...)

bib/cmd/bootc-image-builder/main.go
mvo5: Not sure I follow - when I do the following:
```go
index 9f00131..5783e81 100644
@@ -321,8 +321,7 @@ func cmdBuild(cmd *cobra.Command, args []string) error {
                }
 
                fmt.Printf("Checking AWS region access...\n")
-               var regions []string
-               regions, err = client.Regions()
+               regions, err := client.Regions()
                if err != nil {
                        return fmt.Errorf("retrieving AWS regions for '%s' failed: %w", region, err)
                }
```
it still compiles for me, go is fine with this because `regisions` is a new var so that is declared and err is just assigned, see https://go.dev/doc/effective_go#redeclaration and is quite common.

Or am I missing something?

schuellerf: IntelliJ doesn't like this - but if it's an accepted re-declaration I'm happy to use this as it's nicer for sure!

Makefile
schuellerf: ... I'll just remove the comment :-D

bib/cmd/bootc-image-builder/main.go
mvo5: Out of curiosity, what does InteliJ say here? Is it a warning or an error? It's super common in go code to use this pattern of just redeclartions.

schuellerf: Actually it's not even a warning its more a hint. The variable is just underlined as if it would be a typo :-/
I just checked and it should be a "helper" to notify me that I do redeclare. The underline just distracted me... I'll ignore from now on...

ondrejbudai (approved):

PR #301, opened 2024-03-22
Revert "build: Enable prefetch of go deps in konflux pipeline"

ondrejbudai:
This reverts commit dfdc210238f7f6c2b6c2dd06ed0ace8246b11da3.

Since this commit, all Konflux pipelines are failing. Let's try
reverting it to see if it fixes the issue.

cc @ralphbean


ondrejbudai:
Konflux is green, I think this worked. :)

mvo5 (approved):

PR #302, opened 2024-03-22
github: disable tests on macos

ondrejbudai:
The tests are broken for weeks, probably due to broken podman store
on our macos runner. This already happened in the past and it always
requires a manual intervention to fix it.

Since we don't think that these tests provide much value over what
we are already testing on Linux in GitHub Actions and in Testing Farm,
let's disable them.


mvo5 (approved):
Thank you!

Issue #289, opened 2024-03-19
Mount containers storage as read-only


kingsleyzissou:
To unblock using `containers-storage` we are mounting the host's containers-storage to the container's
`/var/lib/containers/storage`. Ideally we would like to mount the host storage as read-only to `/usr/lib/containers/storage`
When trying to enable this for `bootc-image-builder` and `osbuild` we ran into:
https://github.com/containers/storage/issues/1864

Creating this issue just to keep track of this.

Issue #142, opened 2024-01-19
Support containers auth


achilleas-k:
Currently we don't support using containers that require authorization.

For building, osbuild uses skopeo to pull a container and will implicitly read the auth file from the standard locations described in `containers-auth.json(5)`.  However, bootc-image-builder wont be able to resolve the container to generate the manifest unless we explicitly set the path to the auth file with `resolver.AuthFilePath = path`.

We should support (and document) a process for users to mount an auth file into the BIB container and read it during the resolve process.

This limitation can also be worked around with the local container store feature once that's finished (pull the container on the host while authed and then use the local store to build).

ondrejbudai:
I agree that this should be solved with the local container store, not sure if we want to reimplement auth in bib...

achilleas-k:
Yeah I think I wrote this before we knew exactly what we were going to do with local store.  I'll consider this fixed once we get that.

ondrejbudai:
We implementing pulling from the host's container storage, and this is the recommended of doing authenticated pulls (so bib doesn't need to reimplement auth itself). See https://github.com/osbuild/bootc-image-builder?tab=readme-ov-file#using-local-containers for docs.

PR #303, opened 2024-03-22
README: update local storage docs

kingsleyzissou:
Since we are always pulling the latest image by default the container image will always be in the host's local storage. Update the readme to reflect this.

Additionally, address one or two of the minor comments from https://github.com/osbuild/bootc-image-builder/pull/120.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,19 +40,20 @@ sudo podman run \
     --security-opt label=type:unconfined_t \
     -v $(pwd)/config.json:/config.json \
     -v $(pwd)/output:/output \
+    -v /var/lib/containers/storage:/var/lib/containers/storage \
     quay.io/centos-bootc/bootc-image-builder:latest \
     --type qcow2 \
     --config /config.json \
     quay.io/centos-bootc/fedora-bootc:eln
 ```
 
+NOTE: local storage is being used by default. If the `--local` flag is not provided, as in the above example,
+the latest image will be pulled into the local storage.
+
 ### Using local containers
 
-To use containers from local container's storage rather than a registry, we need to ensure two things:
-- the container exists in local storage
-- mount the local container storage
-
-Since the container is run in `rootful` only root container storage paths are allowed.
+To skip pulling an image into local storage and use an existing container image, the `--local` flag can be used,
+as below:
 
 ```bash
 sudo podman run \
@@ -70,8 +71,6 @@ sudo podman run \
     --local \
     localhost/bootc:eln
 ```
-
-When using the --local flag, we need to mount the storage path as a volume. With this enabled, it is assumed that the target container is in the container storage.
 
 ### Running the resulting QCOW2 file on Linux (x86_64)
 
--- a/test/test_build.py
+++ b/test/test_build.py
@@ -77,10 +77,7 @@ def image_type_fixture(shared_tmpdir, build_container, request, force_aws_upload
     """
     container_ref, images, target_arch, local = parse_request_params(request)
 
-    if not local:
-        with build_images(shared_tmpdir, build_container, request, force_aws_upload) as build_results:
-            yield build_results[0]
-    else:
+    if local:
         cont_tag = "localhost/cont-base-" + "".join(random.choices(string.digits, k=12))
 
         # we are not cross-building local images (for now)
@@ -93,8 +90,8 @@ def image_type_fixture(shared_tmpdir, build_container, request, force_aws_upload
             f"containers-storage:[overlay@/var/lib/containers/storage+/run/containers/storage]{cont_tag}"
         ])
 
-        with build_images(shared_tmpdir, build_container, request, force_aws_upload) as build_results:
-            yield build_results[0]
+    with build_images(shared_tmpdir, build_container, request, force_aws_upload) as build_results:
+        yield build_results[0]
 
 
 @pytest.fixture(name="images", scope="session")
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -40,19 +40,20 @@ sudo podman run \
...
+    -v /var/lib/containers/storage:/var/lib/containers/storage \
     quay.io/centos-bootc/bootc-image-builder:latest \
     --type qcow2 \
     --config /config.json \
     quay.io/centos-bootc/fedora-bootc:eln
 ```
 
-### Using local containers
+NOTE: local storage is being used by default. If the `--local` flag is not provided, as in the above example,
+the latest image will be pulled into the local storage.
```
cgwalters, line 51: There's no pulling involved here, right?  We're just...using the image that is local.  But now I'm confused as if it's default, why do we need to talk about it?

kingsleyzissou: We changed this up in https://github.com/osbuild/bootc-image-builder/pull/262 so that we always pull the container image

achilleas-k:
What's the status of this?  Do we still need it?

kingsleyzissou:
> What's the status of this? Do we still need it?

I think @ondrejbudai had some reservations about it but I might be misremembering... but if I am, what are your thoughts here?

kingsleyzissou:
Yeah those are all valid points. The `--local` flag isn't very descriptive. I'll wait for Colin's feedback on that.

Regarding mounting the store `r/o` I don't think it's possible because we mount the container images in `osbuild`:
- https://github.com/osbuild/bootc-image-builder/issues/289
- https://github.com/containers/storage/pull/1873

ondrejbudai:
This makes me wonder if we can create a container from a read-only image (this should be possible, right?) and then mount the container. Alternatively, we can teach osbuild to use podman as a buildroot, but that's a more long-term project.

cgwalters:
> Regarding --local: let's recap what it does today: --local is equal to --pull=always (if bib is compared to podman build or podman run). --local=false equals to --pull=never.

It's the opposite right?  `--local` means `--pull=never`.

BTW the big problem here is not actually about *pulling* it's about things like bib not using the container auth stack https://github.com/osbuild/bootc-image-builder/issues/142 by default, which still applies without specifying `--local`.

cgwalters:
My take here is we should change bib to exactly match the podman default: pull-if-not-exists.

Anyone who wants to update the image can already do `podman pull` on it explicitly too.

But I am also fine with having a `--pull=always` in bib.

kingsleyzissou:
> It's the opposite right? --local means --pull=never.
This is just another sign that this is really bad UX and a bad flag name 😅 but yeah you're right `--local == --pull=never`

kingsleyzissou:
> But I am also fine with having a --pull=always in bib (but if it exists it must honor the container auth stack and actually the full host container config for /etc/containers around image registry remapping, etc.). IOW it needs to defer to executing podman pull on the host or equivalent.

I think we're doing that [already](https://github.com/osbuild/bootc-image-builder/blob/291a8afecbfc48450f047fb2e4fc158197032b0e/bib/cmd/bootc-image-builder/main.go#L220), no?

test/test_build.py
```diff
@@ -77,10 +77,7 @@ def image_type_fixture(shared_tmpdir, build_container, request, force_aws_upload
     """
     container_ref, images, target_arch, local = parse_request_params(request)
 
-    if not local:
```
mvo5, line 80: Would you mind opening a separate PR with this fix? It seems orthogonal to the discussion in here (and useful on it's own)

kingsleyzissou: Yeah absolutely, will do that now

kingsleyzissou:
Small POC here: https://github.com/osbuild/bootc-image-builder/pull/418

rhatdan:
The default should be --pull=missing.

rhatdan:
$ podman run --help | grep -- --pull

README.md
rhatdan: This is a bad default, we are experiencing bad behaviour with ai-lab-recipes, because the images are huge.  Default should be to only pull if the image is missing.

rhatdan:
Not sure why you need to mount host storage read/write.  Lock files should be able to be used if you use an ADDITIONAL STORE. 

If you mount /var/lib/containers/storage:/usr/lib/containers/storage:RO and have storage.conf setup with an additional store pointed at /usr/lib/containers/storage then everything will work without having a writeable storage directory.

Latest fedora podman has this setup by default.

 $ grep -C 3 additionalimagestore /usr/share/containers/storage.conf 

# AdditionalImageStores is used to pass paths to additional Read/Only image stores
# Must be comma separated list.
additionalimagestores = [
"/usr/lib/containers/storage",
]

kingsleyzissou:
> Not sure why you need to mount host storage read/write.  Lock files should be able to be used if you use an ADDITIONAL STORE. 
> 
> 
> 
> If you mount /var/lib/containers/storage:/usr/lib/containers/storage:RO and have storage.conf setup with an additional store pointed at /usr/lib/containers/storage then everything will work without having a writeable storage directory.
> 
> 
> 
> Latest fedora podman has this setup by default.
> 
> ```
> 
>  $ grep -C 3 additionalimagestore /usr/share/containers/storage.conf 
> 
> # AdditionalImageStores is used to pass paths to additional Read/Only image stores
> 
> # Must be comma separated list.
> 
> additionalimagestores = [
> 
> "/usr/lib/containers/storage",
> 
> ]
> 
> ```

The issue we are having is we mount the container in an `osbuild` [stage](https://github.com/osbuild/osbuild/blob/c3da95780a89019b8e2253104a40925a7ec704d8/stages/org.osbuild.container-deploy#L16) and copy the contents into the build root. Trying to mount a container from the RO store errors out and the image build fails. Is there any way to work around this?

PR #304, opened 2024-03-22
bib: switch to use `bootc install to-filesystem` (HMS-3453)

mvo5:
This PR switches bib to the version of the images library from PR https://github.com/osbuild/images/pull/462 that defaults to use `bootc install to-filesystem`. Because some things will work differently in this model this PR touches a lot of things:
1. It removes customizations and the `-config` option as we currently have only very limited support for them in the new world (for the iso the customizations are still supported via --iso-config)
2. It updates the README to have a new default example so that users can still play around and login into a generated image (should probably add an example for --iso-config there now though)
3. switches the testing to use the bootc -root-authorized-keys support instead of the previous passwd based testing

This is ready now but I expect test failures until https://github.com/containers/bootc/pull/407 is available in our tested containers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,32 +23,35 @@ $ podman machine start
 
 ## 🚀 Examples
 
-The following example builds a [Fedora ELN](https://docs.fedoraproject.org/en-US/eln/) bootable container into a QCOW2 image for the architecture you're running
-the command on.
-
-The `fedora-bootc:eln` base image does not include a default user. This example injects a [user configuration file](#-build-config)
-by adding a volume-mount for the local file as well as the `--config` flag to the bootc-image-builder container.
-
-The following command will create a QCOW2 disk image. First, create `./config.json` as described above to configure user access.
+The following example builds a derrived [Fedora ELN](https://docs.fedoraproject.org/en-US/eln/) bootable container into a QCOW2 image for the architecture you're running the command on.
 
 ```bash
+cat > Containerfile <<'EOF'
+FROM quay.io/centos-bootc/fedora-bootc:eln
+RUN useradd alice --password "$(openssl passwd -6 "bob")" --groups wheel
+EOF
+sudo podman build -f Containerfile -t my-bootc
 sudo podman run \
     --rm \
     -it \
     --privileged \
     --pull=newer \
     --security-opt label=type:unconfined_t \
-    -v $(pwd)/config.json:/config.json \
     -v $(pwd)/output:/output \
+    -v /var/lib/containers/storage:/var/lib/containers/storage \
     quay.io/centos-bootc/bootc-image-builder:latest \
     --type qcow2 \
-    --config /config.json \
-    quay.io/centos-bootc/fedora-bootc:eln
+	--local \
+    localhost/my-bootc
 ```
+
+There is more in-depth discussion how to customize the image and the users in the [bootc documentation](https://bootc-org.gitlab.io/documentation/guide/building.html#logins-and-users).
 
 ### Using local containers
 
-To use containers from local container's storage rather than a registry, we need to ensure two things:
+By default (without the `--local` flag) bootc-image-builder will use
+the podman registry. To use containers from local container's storage,
+we need to ensure two things:
 - the container exists in local storage
 - mount the local container storage
 
@@ -61,12 +64,10 @@ sudo podman run \
     --privileged \
     --pull=newer \
     --security-opt label=type:unconfined_t \
-    -v $(pwd)/config.json:/config.json \
     -v $(pwd)/output:/output \
     -v /var/lib/containers/storage:/var/lib/containers/storage \
     quay.io/centos-bootc/bootc-image-builder:latest \
     --type qcow2 \
-    --config /config.json \
     --local \
     localhost/bootc:eln
 ```
@@ -134,7 +135,6 @@ Usage:
 
 Flags:
       --chown string    chown the ouput directory to match the specified UID:GID
-      --config string   build config file
       --tls-verify      require HTTPS and verify certificates when contacting registries (default true)
       --type string     image type to build [qcow2, ami] (default "qcow2")
 ```
@@ -144,7 +144,6 @@ Flags:
 | Argument         | Description                                                      | Default Value |
 |------------------|------------------------------------------------------------------|:-------------:|
 | **--chown**      | chown the ouput directory to match the specified UID:GID         |       ❌      |
-| **--config**     | Path to a [build config](#-build-config)                         |       ❌      |
 | **--tls-verify** | Require HTTPS and verify certificates when contacting registries |    `true`     |
 | **--type**       | [Image type](#-image-types) to build                             |    `qcow2`    |
 
@@ -261,79 +260,6 @@ The following volumes can be mounted inside the container:
 | `/output` | Used for storing the resulting artifacts               |    ✅     |
 | `/store`  | Used for the [osbuild store](https://www.osbuild.org/) |    No    |
 | `/rpmmd`  | Used for the DNF cache                                 |    No    |
-
-## 📝 Build config
-
-A build config is a JSON file with customizations for the resulting image. A path to the file is passed via  the `--config` argument. The customizations are specified under a `blueprint.customizations` object.
-
-As an example, let's show how you can add a user to the image:
-
-Firstly create a file `./config.json` and put the following content into it:
-
-```json
-{
-  "blueprint": {
-    "customizations": {
-      "user": [
-        {
-          "name": "alice",
-          "password": "bob",
-          "key": "ssh-rsa AAA ... user@email.com",
-          "groups": [
-            "wheel"
-          ]
-        }
-      ]
-    }
-  }
-}
-```
-
-Then, run `bootc-image-builder` with the following arguments:
-
-```bash
-sudo podman run \
-    --rm \
-    -it \
-    --privileged \
-    --pull=newer \
-    --security-opt label=type:unconfined_t \
-    -v $(pwd)/config.json:/config.json \
-    -v $(pwd)/output:/output \
-    quay.io/centos-bootc/bootc-image-builder:latest \
-    --type qcow2 \
-    --config /config.json \
-    quay.io/centos-bootc/fedora-bootc:eln
-```
-
-### Users (`user`, array)
-
-Possible fields:
-
-| Field      | Use                                        | Required |
-|------------|--------------------------------------------|:--------:|
-| `name`     | Name of the user                           |    ✅    |
-| `password` | Unencrypted password                       |    No    |
-| `key`      | Public SSH key contents                    |    No    |
-| `groups`   | An array of secondary to put the user into |    No    |
-
-Example:
-
-```json
-{
-  "user": [
-    {
-      "name": "alice",
-      "password": "bob",
-      "key": "ssh-rsa AAA ... user@email.com",
-      "groups": [
-        "wheel",
-        "admins"
-      ]
-    }
-  ]
-}
-```
 
 ## Building
 
--- a/bib/cmd/bootc-image-builder/image.go
+++ b/bib/cmd/bootc-image-builder/image.go
@@ -261,8 +261,6 @@ func manifestForISO(c *ManifestConfig, rng *rand.Rand) (*manifest.Manifest, erro
 		},
 	}
 
-	img.ISOLabelTmpl = "Container-Installer-%s"
-
 	var customizations *blueprint.Customizations
 	if c.Config != nil && c.Config.Blueprint != nil {
 		customizations = c.Config.Blueprint.Customizations
--- a/bib/cmd/bootc-image-builder/image.go
+++ b/bib/cmd/bootc-image-builder/image.go
@@ -78,8 +78,6 @@ func manifestForDiskImage(c *ManifestConfig, rng *rand.Rand) (*manifest.Manifest
 
 	img := image.NewBootcDiskImage(containerSource)
 	img.Users = users.UsersFromBP(customizations.GetUsers())
-	img.Groups = users.GroupsFromBP(customizations.GetGroups())
-
 	img.KernelOptionsAppend = []string{
 		"rw",
 		// TODO: Drop this as we expect kargs to come from the container image,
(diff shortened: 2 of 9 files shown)


cgwalters:
> It removes customizations and the -config option as we currently have only very limited support for them in the new world

Hmmm...I know I keep saying this but it seems good to try to keep at least some of this.  I haven't dug into the details of the relevant pipelines but as far as I am aware it should be logically equivalent to basically do:

- create filesystems
- run ostree stages
- run blueprint/kickstart type stuff in a container targeting new deployment

- create filesystems
- run bootc install to-filesystem
- run same blueprint/kickstart type stuff

Is the problem here mainly that we've lost an ergonomic way to do "find default deployment root" (subject of some debate in osbuild/osbuild).  That seems pretty easy to fix if so...something like:

```
status=$(mktemp)
bootc install to-filesystem --output-status-json=$status ...
```

And our serialized `status` would have the relative path to deployment.

Or hmmm...maybe actually `bootc install` itself could help spawn a container in the target root itself?  That probably makes a lot of sense actually, as it would help enforce things like only writing to `/etc` and `/var` or so?

cgwalters:
Or maybe a `bootc install mount`?  i.e.:

- <create filesystems using whatever>
- bootc install to-filesystem /path/to/fs
- tmproot=$(mktemp -d)
- bootc install mount-initial-root /path/to/fs ${tmproot}
```

Then one could just directly write to `${tmproot}/etc` and `${tmproot}/var` for example?

**Comment by mvo5** [comment]
> > It removes customizations and the -config option as we currently have only very limited support for them in the new world
[..]
>     * run bootc install to-filesystem
> 
>     * run same blueprint/kickstart type stuff
> Is the problem here mainly that we've lost an ergonomic way to do "find default deployment root" (subject of some debate in osbuild/osbuild). That seems pretty easy to fix if so...something like:
> 
> ```
[..]

Thank you for your thoughts on this! The issue so far was that I had the misconception that we should not write to the deployment root directly. We have ways in osbuild to find the default deployment since https://github.com/osbuild/osbuild/pull/1553 so that should be straightforward. So it seems we could re-add customizations via:
1. write our customization to //ostree/deploy/default/deploy/<hash>.0/etc/* 
2. ensure correct selinux labeling for all the files we touch

If that sounds correct we could look into this in a followup. Unfortunately the main thing we need to re-add would be users so we need to carefully explore this - I would love to look into json user records for this, but it needs at least a configuration change to sshd_config [0]. There is the extra complication that it does not solve the issue of clashing UIDs, i.e. if there is a user created by bib with uid=1000 and the bootc-image (for whatever reason) has files belonging to UID=1000 (for a different username). But that is something we can/could detect and just error (not on day-2 though, that would need support from bootc).

Alternatively we could just keep using the org.osbuild.ostree.passwd stage [1] (which is what we do today).

Thoughts very welcome - but that all sounds like followup material?


[0] https://www.freedesktop.org/software/systemd/man/latest/userdbctl.html#Integration%20with%20SSH
[1] https://github.com/osbuild/osbuild/blob/main/stages/org.osbuild.ostree.passwd#L29

**Comment by cgwalters** [comment]
Yeah, it is possible to create conflicts by configuring users statically both in the disk image and container image today.  (btw I did https://containers.github.io/bootc/building/users-and-groups.html which...doesn't explicitly touch on this but should)

I think we have a design question here: When creating users in bib, should it be static (i.e. we create `/home/someuser` literally in the disk image, or should it actually be done more systemd-sysusers or JSON user record style.

I lean more towards the latter it avoids the physical `/etc/passwd` conflicts - by having both operate dynamically on boot.

This relates to whether SSH keys are maintained via tmpfiles.d in `/etc` or injected statically.

This all said, using sysusers for *interactive* users and JSON user records (or altfiles, or something else) is treading a bit of new ground and it may be simplest to follow the "static" path.

**Comment by cgwalters** [comment]
>  I would love to look into json user records for this, but it needs at least a configuration change to sshd_config [0]. 

It's more than that, systemd-homed isn't enabled in c9s note.  (It's...a giant thing with a lot of system implications, so)

**Comment by achilleas-k** [comment]
So for this PR, do we hold off until we figure out the customization story, or do we merge it and work on post-deploy customizations separately (in the meantime telling users to do everything in the base container)?

**Comment by mvo5** [comment]
Hm, tests fail with:
```
...
 Installed: "centos/grub.cfg"
Trimming boot
fstrim: /run/osbuild/mounts/boot: FITRIM ioctl failed: Function not implemented
ERROR Installing to filesystem: Task Trimming boot failed: ExitStatus(unix_wait_status(256))
  File "/run/osbuild/bin/org.osbuild.bootc.install-to-filesystem", line 50, in <module>
    r = main(args["options"], args["inputs"], args["paths"])
  File "/run/osbuild/bin/org.osbuild.bootc.install-to-filesystem", line 45, in main
    subprocess.run(pargs, env=env, check=True)
  File "/usr/lib64/python3.9/subprocess.py", line 528, in run
subprocess.CalledProcessError: Command '['bootc', 'install', 'to-filesystem', '--source-imgref', 'containers-storage:[overlay@/run/osbuild/containers/storage+/run/containers/storage]65d4b6a5081e0e82e014ea4083e7002cfbc18ee014084063e434f5bc360af85e', '--skip-fetch-check', '--generic-image', '--root-ssh-authorized-keys', '/tmp/bootc-ssh-auth-keys-4aj7p2kw', '--karg', 'rw', '--karg', 'console=tty0', '--karg', 'console=ttyS0', '/run/osbuild/mounts']' returned non-zero exit status 1.
mount/part1 (org.osbuild.fat): umount: /store/tmp/buildroot-tmp-mclvonqd/mounts/boot/efi unmounted
mount/part2 (org.osbuild.ext4): umount: /store/tmp/buildroot-tmp-mclvonqd/mounts/boot unmounted
mount/part3 (org.osbuild.ext4): umount: /store/tmp/buildroot-tmp-mclvonqd/mounts/ unmounted
```

**Comment by achilleas-k** [comment]
Seems https://github.com/containers/bootc/commit/b8f2b6dfd6785a008821ff65c503f96b06f968b6#diff-fe45e8cd8b566039edc4c6a4e80f355754d32078b45e465a407e22e27c39741aR649 should fix that.  What version are we using right now in tests?

**Comment by mvo5** [comment]
> Seems [containers/bootc@b8f2b6d#diff-fe45e8cd8b566039edc4c6a4e80f355754d32078b45e465a407e22e27c39741aR649](https://github.com/containers/bootc/commit/b8f2b6dfd6785a008821ff65c503f96b06f968b6#diff-fe45e8cd8b566039edc4c6a4e80f355754d32078b45e465a407e22e27c39741aR649) should fix that. What version are we using right now in tests?

Thanks! Nice one I'm not sure what we have, maybe we should print this as part of the metadata for the tests what version of bootc is used. I will look into this, will make understanding issues like this a lot easier :)

**Comment by achilleas-k** [comment]
> what version of bootc is used

Since we're pulling containers, the container ID should be enough, but that might be adding extra steps to the debugging process so maybe you're right.  Perhaps running the container and doing a `rpm -qa` would be a good bit of metadata to have.

**Comment by mvo5** [comment]
Hm, this will not work as is I think - centos-bootc is at booc 0.1.7 afaict but we need https://github.com/containers/bootc/pull/407 from 0.1.9 to make the ssh login work. We could explore alternatives of course to the root ssh keys …

File: README.md [hunk 1]
```diff
@@ -262,79 +261,6 @@ The following volumes can be mounted inside the container:
 | `/store`  | Used for the [osbuild store](https://www.osbuild.org/) |    No    |
 | `/rpmmd`  | Used for the DNF cache                                 |    No    |
 
-## 📝 Build config
-
-A build config is a JSON file with customizations for the resulting image. A path to the file is passed via  the `--config` argument. The customizations are specified under a `blueprint.customizations` object.
```
cgwalters, line 267: Would it be helpful to have a realtime sync on this?  I sent an invite.

README.md
```diff
@@ -144,7 +144,6 @@ Flags:
 | Argument         | Description                                                      | Default Value |
 |------------------|------------------------------------------------------------------|:-------------:|
 | **--chown**      | chown the ouput directory to match the specified UID:GID         |       ❌      |
-| **--config**     | Path to a [build config](#-build-config)                         |       ❌      |
```
cgwalters, line 147: The thing is this is used elsewhere so I think we may be partially committed short term.

(There is some things related to this, like the fact I don't quite understand why it's JSON and not TOML as seems would be expected for blueprints, but that's an aside)

bib/cmd/bootc-image-builder/main.go
```diff
@@ -455,11 +473,13 @@ func run() error {
...
 	rootCmd.AddCommand(manifestCmd)
 	manifestCmd.Flags().Bool("tls-verify", true, "require HTTPS and verify certificates when contacting registries")
-	manifestCmd.Flags().String("config", "", "build config file")
+	manifestCmd.Flags().String("iso-config", "", "build config file for the iso")
 	manifestCmd.Flags().String("rpmmd", "/rpmmd", "rpm metadata cache directory")
 	manifestCmd.Flags().String("target-arch", "", "build for the given target architecture (experimental)")
 	manifestCmd.Flags().StringArray("type", []string{"qcow2"}, fmt.Sprintf("image types to build [%s]", allImageTypesString()))
 	manifestCmd.Flags().Bool("local", false, "use a local container rather than a container from a registry")
+	// XXX: hide from help?
+	manifestCmd.Flags().String("experimental-root-ssh-authorized-key", "", "authorized ssh key for root as string")
```
cgwalters, line 482: Not opposed to this, but the more I've been thinking about this (you probably saw) the more I feel we need to build up the story of using blueprints/kickstarts in container builds and align that with disk image generation and not emphasize one-offs like this.

And yes I know I added it to bootc install, but the use case is different there in a way because the bootc install is really low level and I was thinking of it more like a generic low level escape hatch.

But here we can be more opinionated I think.

cgwalters:
> Hm, this will not work as is I think - centos-bootc is at booc 0.1.7 afaict but we need https://github.com/containers/bootc/pull/407 from 0.1.9 to make the ssh login work. We could explore alternatives of course to the root ssh keys …

I am working to unstick this...as you might imagine, Random Process Stuff is holding it up...

mvo5:
> > Seems [containers/bootc@b8f2b6d#diff-fe45e8cd8b566039edc4c6a4e80f355754d32078b45e465a407e22e27c39741aR649](https://github.com/containers/bootc/commit/b8f2b6dfd6785a008821ff65c503f96b06f968b6#diff-fe45e8cd8b566039edc4c6a4e80f355754d32078b45e465a407e22e27c39741aR649) should fix that. What version are we using right now in tests?
> 
> Thanks! Nice one I'm not sure what we have, maybe we should print this as part of the metadata for the tests what version of bootc is used. I will look into this, will make understanding issues like this a lot easier :)
> 
> [edit: looks like this happens with `2024-04-02T17:09:50.1335407Z _ ERROR at setup of test_image_boots[quay.io/centos-bootc/centos-bootc:stream9,raw,arm64]` so maybe just another case of a missing qemu-user syscall? I will investigate]

This is indeed another case of a missing syscall in qemu-user, I attached the fix here and will test/sent it to upstream
[0001-linux-user-Add-FITRIM-ioctl.patch.txt](https://github.com/osbuild/bootc-image-builder/files/14849152/0001-linux-user-Add-FITRIM-ioctl.patch.txt)

README.md
achilleas-k: > I don't quite understand why it's JSON and not TOML

No real (good) reason other than the initial version BIB was heavily based on https://github.com/osbuild/images/blob/main/cmd/build/main.go, which is used for test image builds and it was simpler to use the json formatting for config files.

Since we're moving towards making it more explicit that this config *is* an instance of an Image Builder blueprint, I agree that we should make it a toml to make the connection clearer.

bib/cmd/bootc-image-builder/image.go
```diff
@@ -265,8 +261,6 @@ func manifestForISO(c *ManifestConfig, rng *rand.Rand) (*manifest.Manifest, erro
 		},
 	}
 
-	img.ISOLabelTmpl = "Container-Installer-%s"
```
ondrejbudai, line 268: I think we want something like `img.ISOLabel = fmt.Sprintf("Container-Installer-%s", c.Architecture.String())` here instead. I'm not deeply familiar with ISOs, but I think we should keep the label.

mvo5:
Fwiw, this is being reworked right now so that the `--config` option stays and that user customization works as before (by mounting the deployment root and running the org.osbuild.users stage in this environment), see https://github.com/osbuild/images/compare/main...mvo5:images:bootc-disk-customize-again for the in progress PR that will be split out into smaller ones like https://github.com/osbuild/images/pull/563.

cgwalters:
> This is indeed another case of a missing syscall in qemu-user, I attached the fix here and will test/sent it to upstream
[0001-linux-user-Add-FITRIM-ioctl.patch.txt](https://github.com/osbuild/bootc-image-builder/files/14849152/0001-linux-user-Add-FITRIM-ioctl.patch.txt)

That makes sense and looks generally useful, thoguh https://github.com/containers/bootc/pull/462 is arguably what we should be doing here, right?

cgwalters:
> Fwiw, this is being reworked right now so that the --config option stays

(We covered some of this in relatime chat, but I want to record here too)
I want to apologize as I am sure it seems like I have been giving conflicting messaging here.  It's...complicated.  I really do want to make it easy and nice to bind user state into the container image.  At the same time though, a ton of people are going to enter through Anaconda and set up users and ssh keys there...and it all works by doing the same thing that the bib config is doing in writing to `/etc` and `/var` in the deployment root.

Basically I think there are potential use cases for multiple paths:
- Baking into container image
- Baking into disk image
- Fetching dynamically (cloud-init, FreeIPA, etc.)

In particular an advantage of the disk image flow is that in some cases, it's easier to "secure" the disk image than it is the container image, and having "bootstrap secrets" only in the disk images makes a lot of sense.

mvo5:
Given that we agreed we cannot drop customizations I close this one, the latest approach (that keeps customizations) is in https://github.com/osbuild/bootc-image-builder/pull/342

bib/cmd/bootc-image-builder/image.go
mvo5: Thank you! I fixed this in https://github.com/osbuild/bootc-image-builder/pull/342

PR #306, opened 2024-03-22
Update to go 1.20

cgwalters:
I think pinning on 1.19 is holding some dependency updates back, and 1.20 is in rhel9 now.


cgwalters:
Actually, we may be able to bump to 1.21 in general, which is in c9s but not (yet) in UBI.

ondrejbudai (approved):
Let's be conservative and do 1.20 until 1.21 hits stable RHEL.

PR #308, opened 2024-03-22
Update runc to 1.1.12 (CVE-2024-21626)

jeckersb:
Signed-off-by: John Eckersberg <jeckersb@redhat.com>


cgwalters (approved):
We're having a realtime chat about this and at the moment we aren't sure why dependabot didn't pick this up.  We were thinking it might have to do with the Go version.  So I think this PR is correct, but we also probably want to merge https://github.com/osbuild/bootc-image-builder/pull/306 to unblock dependabot if that's true (and make our go version not a lie).

cgwalters (approved):

PR #310, opened 2024-03-25
setup: deal with btrfs subvolumes when checking mapped storage

mvo5:
When the container storage is checked it may come from a btrfs subvolume. In /proc/self/mountinfo this looks like:
```
1364 1345 0:33 /root/var/lib/containers/storage /var/lib/containers/storage rw,relatime - btrfs /dev/mapper/luks-1356123c-c7b9-1bd7-a8bf-e13d4cbb5500 rw,seclabel,compress=zstd:1,ssd,discard=async,space_cache,subvolid=257,subvol=/root
```
as a quick fix just compare the suffix of the root instead of the full path.

Thanks for Ondrej!

[I want to write a test but I also want to unblock Ondrej]

### DIFF
--- a/bib/internal/setup/setup.go
+++ b/bib/internal/setup/setup.go
@@ -2,9 +2,11 @@ package setup
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strings"
 
 	"github.com/moby/sys/mountinfo"
 	"golang.org/x/sys/unix"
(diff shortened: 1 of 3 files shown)


bib/internal/setup/setup.go
```diff
@@ -100,23 +102,43 @@ func Validate() error {
...
-// ValidateHasContainerStorageMounted checks that the container storage
-// is mounted inside the container
-func ValidateHasContainerStorageMounted() error {
+var insideContainer = func() (bool, error) {
 	if err := exec.Command("systemd-detect-virt", "-c").Run(); err != nil {
 		if exitErr, ok := err.(*exec.ExitError); ok {
 			// not running in a container, just exit
 			if exitErr.ExitCode() == 1 {
-				return nil
+				return true, nil
```
ondrejbudai, line 110: `systemd-detect-virt -c` exits with 1 if it is NOT running in a container. Thus, I think that this function should return false here (because it's called `insideContainer`). IoW the return values are as far as I understand flipped.

bib/internal/setup/setup.go
```diff
@@ -100,23 +102,43 @@ func Validate() error {
...
+}
+
+// ValidateHasContainerStorageMounted checks that the container storage
+// is mounted inside the container
+func ValidateHasContainerStorageMounted() error {
+	inside, err := insideContainer()
+	if err != nil {
+		return err
+	}
+	if inside {
```
ondrejbudai, line 125: However, this check is actually correct I think. ;)

bib/internal/setup/setup.go
```diff
@@ -125,8 +147,10 @@ func ValidateHasContainerStorageMounted() error {
 			continue
 		}
 		containerStorageMountFound = true
-		if mnt.Root != containersStorage {
-			return fmt.Errorf("not the host /var/lib/storage/containers but %q", mnt.Root)
+		// on btrfs the contaners storage might be on a subvolume
```
ondrejbudai, line 150:
```suggestion
		// on btrfs the containers storage might be on a subvolume
```

ondrejbudai (changes requested):
Generally looks good, but I think that the return values of `insideContainer` are backwards.

bib/internal/setup/setup.go
mvo5: Thanks, I'm a Muppet. Thanks for catching it! I simplified and reuploaded.

bib/internal/setup/setup.go
mvo5: Thank you!

bib/internal/setup/setup.go
mvo5: Yeah, *cough* :) Fixed!

ondrejbudai (approved):

PR #319, opened 2024-04-01
setup: Simplify container-storage mounted check

cgwalters:
I ran into the `--local` option not working at least in my setup with the current container image (which for some reason hasn't been updated, and predates the latest tip commit which talks about btrfs)

On this current test system (MacOS + podman 5 + default podman-machine) things do work with the tip commit.

However...I don't quite understand the need to *both* try to run `systemd-detect-virt` *and* parse `/proc/self/mountinfo`.

(BTW, the logic for `insideContainer` was really confusing me because
 it only returned `true` if we were *not* in a container...I wonder
 if that was really intentional?)

Anyways, I think the goal here is just a friendly direct error message if it doesn't look like the mount is there, which we can do by just checking for the file path.  If it isn't mounted from the host then nothing will be there.


cgwalters:
OK, that one is a cross architecture build.  Hmm...those errors are weird.

> time="2024-04-02T12:23:10Z" level=error msg="Refreshing container 2193edd334ed77019ed45a393ecfbbc814c774ac03ded0142d15d3e814be923c: acquiring lock 0 for container 2193edd334ed77019ed45a393ecfbbc814c774ac03ded0142d15d3e814be923c: file exists"

It's surprising to me that we're going in a code path that accesses "containers" (i.e. running) versus just images here.

cgwalters:
> iic we have an integration test for this so the unit test removal is fine too.

Yep, several, and there are external tests too.  The new code is so simple I don't see unit testing it as making much sense.

PR #324, opened 2024-04-03
main: Wire up --log-level, add some debug logging

cgwalters:
There doesn't seem to be much usage of logrus here, but I find debug logging very useful.

### DIFF
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -133,6 +133,7 @@ func getContainerSize(imgref string) (uint64, error) {
 		return 0, fmt.Errorf("cannot parse image size: %w", err)
 	}
 
+	logrus.Debugf("container size: %v", size)
 	return size, nil
 }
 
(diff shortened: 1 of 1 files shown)


cgwalters:
(Only lightly tested locally here, I'm on a ✈️ at the moment, and the Go module fetch step ruins things there)

mvo5 (approved):
Thank you, this looks good and useful. I especially like that we info pulling image now because that tended to be a bit slow and it looked like bib was hanging.

cgwalters:
> I especially like that we info pulling image now because that tended to be a bit slow and it looked like bib was hanging.

Well that didn't get printed as we were only logging at error level, not info.  I changed it to be an unconditional print for now.

cgwalters:
Looks like testing farm disk space issues again
> processing tar file(write /big-file3: no space left on device): exit status 1

bib/cmd/bootc-image-builder/main.go
```diff
@@ -139,10 +140,13 @@ func makeManifest(c *ManifestConfig, cacheRoot string) (manifest.OSBuildManifest
 	// to using containers storage in all code paths happened.
 	// We might want to change this behaviour in the future to match podman.
 	if !c.Local {
+		fmt.Printf("Pulling image %s (arch=%s)\n", c.Imgref, c.Architecture.String())
```
mvo5, line 143: Sorry, I just realized (after looking at the failing tests) this has the same issue as: https://github.com/osbuild/bootc-image-builder/pull/286#discussion_r1530421631 - when we use "--manifest" we expect a pure json output. 

So I guess we need to either use stderr (which is weird as it's not an error) or use logrus and in manifest mode set the level to error but in build mode to info (or something else I have not thought about :)

cgwalters:
(I thought I'd adjusted the `fmt.Printf`, but now finally fixed)

cgwalters:
This one isn't critical but could probably go in, right?

mvo5 (dismissed):
I like it! Thanks for this PR! (I will add a small smoke test for it tomorrow)

mvo5 (approved):
This :+1: I repushed to fix trivial merge conflicts.

PR #325, opened 2024-04-03
bib: extract handleAWSFlags() helper

mvo5:
cmdBuild is already relatively long so extract the helper to handle the AWS cmdline

This also move to use logrus in the helper, I am sitting on the fence a bit if this should be a info or a debug, it seems if this is a slow check, maybe Info() is better as it is visible by default (AIUI).

### DIFF
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -308,7 +308,7 @@ func handleAWSFlags(cmd *cobra.Command) (upload bool, err error) {
 		return false, err
 	}
 
-	fmt.Println("Checking AWS region access...")
+	logrus.Info("Checking AWS region access...")
 	regions, err := client.Regions()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


bib/cmd/bootc-image-builder/main.go
```diff
@@ -290,6 +290,55 @@ func cmdManifest(cmd *cobra.Command, args []string) error {
...
+		return false, fmt.Errorf("aws flags set for non-ami image type (type is set to %s)", strings.Join(imgTypes, ","))
+	}
+
+	// check as many permission prerequisites as possible before starting
+	client, err := awscloud.NewDefault(region)
+	if err != nil {
+		return false, err
+	}
+
+	logrus.Info("Checking AWS region access...")
```
mvo5, line 311: I just noticed that we only show lgorus.rError so maybe this needs tweaking (info by default sounds reasonable to me)

ondrejbudai: Let's do it in a follow-up?

ondrejbudai (approved):
Thanks!

PR #339, opened 2024-04-09
bib: use target architecture for buildroot as well

mvo5:
The existing code is trying to fetch two containers when doing a cross-arch image. The buildroot is run in the native architecture and only the target tree is run non-native. However it turns out this is not neccessary as even doing the buildroot in non-native is fast enough (given that it's mostly I/O and all I/O is run naively via syscall translation).

So this commit removes the fetching of the native arch container for the buildroot. Note that in practise this was already the case when passing --target-arch to any container that had native build available. Here due to a bug/misfeature in the resolver library used it already resolved (incorrectly) to the target architecture and nobody noticed because it was good enough (after [0] got fixed).

[0] https://lists.nongnu.org/archive/html/qemu-devel/2024-02/msg03971.html

P.S. I did a benchmark of the centos-stream9,arm64 test on my machine:
- native buildroot: 7:44min
- emulated buildroot: 8:22min

### DIFF
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -170,28 +170,22 @@ func makeManifest(c *ManifestConfig, cacheRoot string) (manifest.OSBuildManifest
 
 	// Resolve container - the normal case is that host and target
 	// architecture are the same. However it is possible to build
-	// cross-arch images. When this is done the "build" pipeline
-	// will run with the "native" architecture of the target
-	// container and the other pipelines (usually just "image"
-	// will use the target architecture).
+	// cross-arch images by using qemu-user. Just run everything
+	// (including the build-root) with the target arch then, it
+	// is fast enough (given that it's mostly I/O and all I/O is
+	// run naively via syscall translation)
 	hostArch := arch.Current().String()
 	targetArch := c.Architecture.String()
 
-	resolverNative := container.NewResolver(hostArch)
-	resolverTarget := resolverNative
-	if hostArch != targetArch {
-		resolverTarget = container.NewResolver(targetArch)
+	var resolver *container.Resolver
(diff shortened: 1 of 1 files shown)


bib/cmd/bootc-image-builder/main.go
```diff
@@ -170,28 +170,22 @@ func makeManifest(c *ManifestConfig, cacheRoot string) (manifest.OSBuildManifest
...
-	resolverNative := container.NewResolver(hostArch)
-	resolverTarget := resolverNative
-	if hostArch != targetArch {
-		resolverTarget = container.NewResolver(targetArch)
+	var resolver *container.Resolver
+	if targetArch != "" && hostArch != targetArch {
+		resolver = container.NewResolver(targetArch)
+	} else {
+		resolver = container.NewResolver(hostArch)
 	}
```
cgwalters, line 185: Isn't this logic just equivalent to:

```
	if targetArch != "" {
		resolver = container.NewResolver(targetArch)
	} else {
		resolver = container.NewResolver(hostArch)
	}
```
If `hostArch == targetArch`, then it's the same thing I think.

cgwalters (approved):
Looks sane to me, one drive by nit

kingsleyzissou (dismissed):

kingsleyzissou:
> P.S. I did a benchmark of the centos-stream9,arm64 test on my machine:
> 
> - native buildroot: 7:44min
> - emulated buildroot: 8:22min

Neat!!

bib/cmd/bootc-image-builder/main.go
mvo5: Thanks, excellent point. I simplfied it accordingly.

kingsleyzissou (approved):

PR #350, opened 2024-04-12
Automatically pick up `/config.json` if it exists

cgwalters:
When passing a config, it's an ergonomic hit to need to provide it on *both* the `podman run` invocation *and* the arguments to the binary.

Change things so that we automatically pick up `/config.json` if it exists in the container.  This way the user only needs to give the podman argument.


achilleas-k (approved):

achilleas-k:
Had to resolve the conflict manually.  Rebased.

achilleas-k (approved):

PR #351, opened 2024-04-12
document that `customizations.kernel.append` exists

cgwalters:
I noticed this was wired up in the code, but undocumented and untested.

There are valid use cases for machine-specific kernel arguments, so let's support that here.

(What's the best place to verify that the kargs show up in the test
 VMs? )

### DIFF
--- a/README.md
+++ b/README.md
@@ -348,6 +348,21 @@ Example:
 }
 ```
 
+### Kernel Arguments (`kernel`, mapping)
+
+```json
+{
+  "blueprint": {
+    "customizations": {
+      "kernel": {
+        "append": "mitigations=auto,nosmt"
+      }
+    }
+  }
+}
+
+```
+
 ## Building
 
 To build the container locally you can run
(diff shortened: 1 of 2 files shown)


henrywang:
> (What's the best place to verify that the kargs show up in the test VMs? )
I think `sudo dmesg | grep "Kernel command line:"` can get the kernel args.

BTW, This setting will be conflicted with same configuration in Containerfile if it's configured on both side.

mvo5:
> I noticed this was wired up in the code, but undocumented and untested.
>
> There are valid use cases for machine-specific kernel arguments, so let's support that here.
> 
> (What's the best place to verify that the kargs show up in the test VMs? )

Thanks for this! I think something like the attached commit could be used to test this. I just gets /proc/cmdline from the booted system to validate the kargs. 
[0001-test-test-that-kernel-commandline-args-are-passed-co.patch.txt](https://github.com/osbuild/bootc-image-builder/files/14977011/0001-test-test-that-kernel-commandline-args-are-passed-co.patch.txt)

I will run this locally to validate that the diff works and can also push directly to this PR if/once it works.

mvo5 (approved):
Thank you! I like it. We can merge now and I can do the test (which is an excellent suggestion) in a followup (or push it directly here whatever is preferred).

[edit: given that this is green I'm actually in favor of the test as a followup]

PR #355, opened 2024-04-16
Fix passing the --tls-verify flag to podman pull

ondrejbudai:
When switching to always leveraging the local storage, I forgot to pass
the --tls-verify flag. This commit fixes it.

@mvo5 any ideas how to test this?

### DIFF
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -257,7 +257,7 @@ func manifestFromCobra(cmd *cobra.Command, args []string) ([]byte, *mTLSConfig, 
 	// to using containers storage in all code paths happened.
 	// We might want to change this behaviour in the future to match podman.
 	if !localStorage {
-		if output, err := exec.Command("podman", "pull", "--arch", buildArch.String(), imgref).CombinedOutput(); err != nil {
+		if output, err := exec.Command("podman", "pull", "--arch", buildArch.String(), fmt.Sprintf("--tls-verify=%v", tlsVerify), imgref).CombinedOutput(); err != nil {
 			return nil, nil, fmt.Errorf("failed to pull container image: %w\n%s", err, output)
 		}
 	}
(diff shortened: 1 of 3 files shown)


achilleas-k (approved):

bib/cmd/bootc-image-builder/main.go
```diff
@@ -139,9 +140,11 @@ func makeManifest(c *ManifestConfig, cacheRoot string) (manifest.OSBuildManifest
 	// to using containers storage in all code paths happened.
 	// We might want to change this behaviour in the future to match podman.
 	if !c.Local {
-		pullCmd := exec.Command("podman", "pull", "--arch", c.Architecture.String(), c.Imgref)
+		pullCmd := exec.Command("podman", "pull", "--arch", c.Architecture.String(), fmt.Sprintf("--tls-verify=%v", c.TLSVerify), c.Imgref)
```
rhatdan, line 143: Doesn't
```
pullCmd := exec.Command("podman", "pull", "--arch", c.Architecture.String(), "--tls-verify", c.TLSVerify, c.Imgref)
```
Work?

ondrejbudai: No, it didn't work for me. It treated `false` and `true` as image refs and tried to pull these.

rhatdan: Ok, this is because --tls-verify is legal. where as --tls-verify=false turns it off.

rhatdan:
LGTM

mvo5:
[..]
> @mvo5 any ideas how to test this?

Interesting one -  I outlined a possible test that works as a blackbox one. We could also do extract a `pullContainer` helper and mock execCommand() in there but go makes this not super straightforward and the attached one uses infrastructure we already have so that seems the quickest way to "success". 
[0001-test-add-tls-opt-test-as-part-of-test_opts.py.patch.txt](https://github.com/osbuild/bootc-image-builder/files/15008906/0001-test-add-tls-opt-test-as-part-of-test_opts.py.patch.txt)

The test runs in 2min on my system and is faster when not run in isolation as the expensive part is the pull of the centos bootc container.

ondrejbudai:
Rebased, added the test from Michael, thanks! :heart:

mvo5 (approved):
Thank you

PR #357, opened 2024-04-16
test: check that `bootc status` contains the right container_ref

mvo5:
We are currently not testing that when the build image has a valid `bootc status` that container the original container reference.


ondrejbudai (approved):
Thanks!

mvo5:
I close this one for now but will reopen the ssh root parts most likely to figure out what is going on there, I suspect that there is a real bug/mismatch in how we setup the root user config.

PR #360, opened 2024-04-17
test: check that `bootc status` contains the right container_ref

mvo5:
We are currently not testing that when the build image has a valid `bootc status` that container the original container reference.

Alternative version to https://github.com/osbuild/bootc-image-builder/pull/357 that works without creating a root user which seems broken right now on fedora but that needs to get investigated later.

### DIFF
--- a/test/test_build.py
+++ b/test/test_build.py
@@ -36,6 +36,7 @@ class ImageBuildResult(NamedTuple):
     img_type: str
     img_path: str
     img_arch: str
+    container_ref: str
     username: str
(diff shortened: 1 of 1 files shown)


ondrejbudai (dismissed):
Thanks for figuring out a workaround! :heart:

achilleas-k (approved):

ondrejbudai (dismissed):
Still good. :)

ondrejbudai (approved):

PR #361, opened 2024-04-17
main: decorate more errors with context and avoid panics

mvo5:
We recently had a bugreport where a panic had very little error context:
```
Generating manifest manifest-iso.json
panic: exec: no command
```
this prompted this commit so that:
a) we do add more context to our errors
b) avoid panicing when we can also return errors

This may lead to some stuttering `cannot do foo: cannot generate foo: cannot make foo a thing` but we can tweak and more context right now is probably better than less (ideally we would have tests that cover this and help us figure out how to fine tune it).


achilleas-k (approved):
Agreed.  Repetition is better than no info.

ondrejbudai (approved):

PR #363, opened 2024-04-18
test: test root user setup

mvo5:
A continuation of https://github.com/osbuild/bootc-image-builder/pull/357 - the root user is used to do a `bootc status` inside the booted container to ensure the setup is correct. This is nicer then the previous `sudo` approach and it has the added benefit of testing the root user creation/setup during the build too.

### DIFF
--- a/test/test_build.py
+++ b/test/test_build.py
@@ -124,7 +124,7 @@ def build_images(shared_tmpdir, build_container, request, force_aws_upload):
 
     username = "test"
     password = "password"
-    kargs = "user.sometestkarg=sometestvalue"
+    kargs = "systemd.journald.forward_to_console=1"
 
     # params can be long and the qmp socket (that has a limit of 100ish
     # AF_UNIX) is derived from the path
--- a/test/test_build.py
+++ b/test/test_build.py
@@ -39,6 +39,7 @@ class ImageBuildResult(NamedTuple):
     container_ref: str
     username: str
     password: str
+    ssh_keyfile_private_path: str
     kargs: str
     bib_output: str
     journal_output: str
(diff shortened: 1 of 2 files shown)


bib/go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/aws/aws-sdk-go v1.52.1
 	github.com/cheggaaa/pb/v3 v3.1.5
 	github.com/google/uuid v1.6.0
-	github.com/osbuild/images v0.58.0
+	github.com/osbuild/images v0.58.1-0.20240430053854-ddf3ea9f0280
```
achilleas-k, line 9: Is this an untagged commit ID?  I did a v0.59 release if you'd rather pull that in.

achilleas-k (approved):
Splendid.  Minor comment but no biggie.

bib/go.mod
mvo5: Sure thing! Fixed now :)

achilleas-k (approved):

achilleas-k (changes requested):
Almost merged.  Missed this `fixup` commit.

mvo5:
> Almost merged. Missed this `fixup` commit.

I removed the fixup commit now and fixed the bug in the PR that prevented the tests from working - it turns out we wrote invalid kickstart files because the keyfile read from disk contained a (not stripped) newline at the end and with the kickstart file looked somehting like:
```
sshkey --username foo "ssh-key
"
``` 
I opened https://github.com/osbuild/osbuild/pull/1772 with one idea

achilleas-k (approved):

PR #365, opened 2024-04-18
bib: run go unittest in GH actions and fix them

mvo5:
Previously the `0` value of the images `arch.Architecture` iota
was of type `ARCH_AARCH64`. However with
https://github.com/osbuild/images/commit/05355326188f3d2ad49003784d681524eba466ce
this is no longer the case an the zero type is now a proper
`UNSET` type. And the tests now fail because they did not set
the architecture.

This PR fixes this by setting the architecture and also adds them to the GH action

### DIFF
--- a/bib/cmd/bootc-image-builder/main_test.go
+++ b/bib/cmd/bootc-image-builder/main_test.go
@@ -10,11 +10,13 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
-	main "github.com/osbuild/bootc-image-builder/bib/cmd/bootc-image-builder"
+	"github.com/osbuild/images/pkg/arch"
 	"github.com/osbuild/images/pkg/blueprint"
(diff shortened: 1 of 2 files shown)


bib/cmd/bootc-image-builder/main_test.go
```diff
@@ -62,16 +64,20 @@ type manifestTestCase struct {
 }
 
 func getBaseConfig() *main.ManifestConfig {
-	return &main.ManifestConfig{Imgref: "testempty"}
+	return &main.ManifestConfig{
+		Architecture: arch.Current(),
```
ondrejbudai, line 68: Hmm, this makes the tests behave differently on each arch... Not sure if that's conceptually OK, or not. :thinking:

mvo5: I have no strong opinion here, I moved to a static arch now.

ondrejbudai (approved):

PR #368, opened 2024-04-19
image: enable `org.fedoraproject.Anaconda.Modules.Users` if needed

mvo5:
It turns out that on centos9 users are not added when specified in kickstart files. This commit should fix this and users are added. I did a manual validation but this will be part of the test for https://github.com/osbuild/bootc-image-builder/pull/338 (the test in 338 still not works, it seems the machine has no network(?) but I guess baby steps :)

P.S. No test as this really needs to be in images and I will add a test there (and then this needs reverting).

### DIFF
--- a/bib/cmd/bootc-image-builder/image.go
+++ b/bib/cmd/bootc-image-builder/image.go
@@ -285,6 +285,12 @@ func manifestForISO(c *ManifestConfig, rng *rand.Rand) (*manifest.Manifest, erro
 
 	img.Users = users.UsersFromBP(customizations.GetUsers())
 	img.Groups = users.GroupsFromBP(customizations.GetGroups())
+	// XXX: this should really be done by images, the consumer should not
+	// need to know these details. so once images is fixed drop it here
+	// again.
+	if len(img.Users) > 0 || len(img.Groups) > 0 {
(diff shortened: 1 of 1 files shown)


bib/cmd/bootc-image-builder/image.go
```diff
@@ -285,6 +285,12 @@ func manifestForISO(c *ManifestConfig, rng *rand.Rand) (*manifest.Manifest, erro
 
 	img.Users = users.UsersFromBP(customizations.GetUsers())
 	img.Groups = users.GroupsFromBP(customizations.GetGroups())
+	// XXX: this should really be done by images, the consumer should not
+	// need to know these details. so once images is fixed drop it here
+	// again.
```
achilleas-k, line 290: Agreed, but good quick fix for now.

achilleas-k (approved):

achilleas-k:
> (the test in 338 still not works, it seems the machine has no network(?) but I guess baby steps :)

Is this the same issue?  Network module not enabled?

mvo5:
> > (the test in 338 still not works, it seems the machine has no network(?) but I guess baby steps :)
> 
> Is this the same issue? Network module not enabled?

The network on boot is fixed with https://github.com/osbuild/images/pull/616 - we do need to find out still (with the upstream anaconda team) if they can just leave the default networking to the bootc container, it seems to have it's own policy for networking right now.

PR #372, opened 2024-04-22
github: put `/var/lib/containers` on `/mnt` via bind mount

mvo5:
On GH runners /mnt has 70G free space and because we ran out of diskspace in https://github.com/osbuild/bootc-image-builder/pull/338 in what appears the container storage move the container storage via a bind mount to `/mnt/var/lib/containers`.


ondrejbudai (approved):
Ugh, their setup is imho quite stupid.

Issue #376, opened 2024-04-22
build osbuild from git and vendor it here


cgwalters:
Instead of `dnf install osbuild.rpm`, we can consider using it as e.g. a git submodule here.

The current situation is a mess because osbuild/images depends on `osbuild.rpm`, and we vendor that Go code, but the RPM floats.  This leads to "lifecycle binding" problems.

Any objections?

achilleas-k:
Pinning a specific version or commit ID for osbuild isn't a bad idea.  We could revert it in the long run when things stabilise a bit (if they do and if we want to).

I don't know about submodules, we could just have a file with a commit ID that we update when we need to, but I'm fine with submodules too since it makes it easier to navigate to the target commit when browsing code.

cgwalters:
PR in https://github.com/osbuild/bootc-image-builder/pull/378 for feedback - only compile (build) tested.

achilleas-k:
Do we still want this?

PR #378, opened 2024-04-22
Add osbuild as a git submodule

cgwalters:
The current situation is a mess because osbuild/images depends on osbuild.rpm, and we vendor that Go code, but the RPM floats. This leads to "lifecycle binding" problems.

Closes: https://github.com/osbuild/bootc-image-builder/issues/376


cgwalters:
TIL the integration tests build the container image inside pytest.  We should clearly be inheriting from the `build` action and passing the container...or maybe using this pattern https://docs.docker.com/build/ci/github-actions/local-registry/
(Have never tried it, though of relevant note is that Kubernetes Prow does this by default, each CI job gets its own kube namespace with registry access)

cgwalters:
OK, the Konflux build worked too on amd64, but just flaked on arm64 due to provisioner issues we're seeing elsewhere.

achilleas-k:
> TIL the integration tests build the container image inside pytest. We should clearly be inheriting from the `build` action and passing the container

A nice thing about the pytest setup is that simply running `pytest` or `pytest -k some_test_name_pattern` will build the container from source locally and run a test without needing any CI scaffolding.

achilleas-k (dismissed):

cgwalters:
I think it may be that testing farm is not doing a recursive clone?  No idea how to fix that, anyone know?

cgwalters:
I tried adding a recursive clone into the tests

achilleas-k:
How does this interact with the TF tests?
https://github.com/osbuild/bootc-image-builder/blob/16be9f7355030d8f2bea183596d850487340ea8e/.github/workflows/testingfarm.yml#L46-L59

There's a checkout action before the test runs, so is that used as the source?  Should that have `submodules: True`' enabled?

plans/all.fmf
```diff
@@ -26,6 +26,8 @@ prepare:
 execute:
   how: tmt
   script: |
+    # Testing farm doesn't do a recursive clone
```
henrywang, line 29: The tmt run all scripts in plan on a random folder, not in repo folder. We can remove this command to avoid error
```
                    # Testing farm doesn't do a recursive clone
                    git submodule update --init --recursive
                    pip install --user -r test/requirements.txt
                    pytest --force-aws-upload
12:23:01         environment: None
12:23:01                 out: fatal: not a git repository (or any of the parent directories): .git
12:23:01                 out: Shared connection to 10.0.133.33 closed.
12:23:01         Command returned '128'.
```

henrywang:
From what we use in https://github.com/virt-s1/bootc-workflow-test, testing farm support submodules clone.
Is that possible the pytest does not work with git submodules because podman build is triggered by pytest.

achilleas-k:
I don't think pytest cares about git or submodules, as long as the files are there.

ondrejbudai:
We still want this. I rebased and fixed conflicts. :)

PR #385, opened 2024-04-24
container: fix crash when `Container.New()` defer() is run

mvo5:
This commit fixes a nil pointer exception when `Container.New()` needs to run the defer that calls c.Stop(). The semantics of named returns are subtle here. The local var `c` is non-nil. However when `New()` returns with `return nil, err` (e.g. in line 64+) then `c` becomes `nil` (because the of return nil). Which means the defer now tries to use a `nil` pointer and crashes.

The fix is to either use `naked` returns everywhere or just remove the named returns. The later seems more in line with what we (and the go community) generally do.

A regression test is added too.

### DIFF
--- a/bib/internal/container/container.go
+++ b/bib/internal/container/container.go
@@ -20,7 +20,7 @@ type Container struct {
 // NB:
 // - --net host is used to make networking work in a nested container
 // - /run/secrets is mounted from the host to make sure RHSM credentials are available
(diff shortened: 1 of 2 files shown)


ondrejbudai (approved):
I told you I don't like named returns, because their semantics is confusing. :D Great catch, thank you very much!

PR #386, opened 2024-04-24
README.md: target c9s because ELN is more aggressive and has bugs

mvo5:
A while ago there was PR#17 that emphasised centos stream9 instead of fedora:eln. We could not take it back then because the iso installer was not quite ready for centos and there were hardcoded EFI names. But now everything is in place so this cherry picks the original changes from Colin.

See also https://github.com/osbuild/bootc-image-builder/pull/17

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,10 +23,10 @@ $ podman machine start
 
 ## 🚀 Examples
 
-The following example builds a [Fedora ELN](https://docs.fedoraproject.org/en-US/eln/) bootable container into a QCOW2 image for the architecture you're running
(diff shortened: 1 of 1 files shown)


cgwalters (approved):

ondrejbudai (approved):
Thanks!

(I opened https://github.com/osbuild/bootc-image-builder/issues/387 to address the CI on PRs like this one.)

cgwalters:
OK so for the GHA issues, looks like it's https://github.com/orgs/community/discussions/120966 and

PR #392, opened 2024-04-25
bib: allow blueprints in toml as well

mvo5:
This commit adds support to load blueprints from toml as well.

[draft as tests are missing]

### DIFF
--- a/bib/internal/buildconfig/config.go
+++ b/bib/internal/buildconfig/config.go
@@ -1,0 +1,85 @@
+package buildconfig
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+
+	"github.com/pelletier/go-toml"
+
+	"github.com/osbuild/images/pkg/blueprint"
+)
+
+type BuildConfig struct {
+	Blueprint *blueprint.Blueprint `json:"blueprint,omitempty" toml:"blueprint"`
+}
+
+// configRootDir is only overriden in tests
+var configRootDir = "/"
+
+func decodeJsonBuildConfig(r io.Reader, what string) (*BuildConfig, error) {
+	dec := json.NewDecoder(r)
+	dec.DisallowUnknownFields()
+
+	var conf BuildConfig
+	if err := dec.Decode(&conf); err != nil {
+		return nil, fmt.Errorf("cannot decode %q: %w", what, err)
+	}
+	if dec.More() {
+		return nil, fmt.Errorf("multiple configuration objects or extra data found in %q", what)
+	}
+	return &conf, nil
+}
+
+func decodeTomlBuildConfig(r io.Reader, what string) (*BuildConfig, error) {
+	dec := toml.NewDecoder(r)
+
+	var conf BuildConfig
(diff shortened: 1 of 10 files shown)


achilleas-k:
I expressed a preference for toml to make life easier for adding kickstart files (see https://github.com/osbuild/images/pull/631).  What I'd like to see, either in this PR or a follow-up:
- [ ] Automatic detection of `/config.toml` like we do for `/config.json`.
- [ ] Prioritise toml in docs (downstream docs should be aware of this too).
- [ ] Move the config one level down, so the blueprint is the top level object, just like in osbuild-composer (this is optional; might be good to have the blueprint as a second level if we want to add BIB specific config options that don't come from the osbuild-composer blueprint).

henrywang:
I have one suggestion on this feature. It will be wonderful to align this toml configuration file with toml file used by osbuild-composer. So the osbuild-composer [Blueprint Refreence](https://osbuild.org/docs/user-guide/blueprint-reference) can be re-used as a reference. Thanks!

achilleas-k:
> I have one suggestion on this feature. It will be wonderful to align this toml configuration file with toml file used by osbuild-composer. So the osbuild-composer [Blueprint Refreence](https://osbuild.org/docs/user-guide/blueprint-reference) can be re-used as a reference. Thanks!

Since we use a very limited number of options here, it might be better to have a dedicated page for BIB.  Or maybe we could add a third bubble colour for `bootc-image-builder` like we have for on-prem vs console.

mvo5:
> > I have one suggestion on this feature. It will be wonderful to align this toml configuration file with toml file used by osbuild-composer. So the osbuild-composer [Blueprint Refreence](https://osbuild.org/docs/user-guide/blueprint-reference) can be re-used as a reference. Thanks!
> 
> Since we use a very limited number of options here, it might be better to have a dedicated page for BIB. Or maybe we could add a third bubble colour for `bootc-image-builder` like we have for on-prem vs console.

I have a branch ready for this in https://github.com/osbuild/bootc-image-builder/compare/main...mvo5:use-toml-blueprints-compat?expand=1 - it also keeps the legacy format compatiblity for now

bib/internal/buildconfig/config.go
```diff
@@ -0,0 +1,85 @@
...
+	"os"
+	"path/filepath"
+
+	"github.com/pelletier/go-toml"
+
+	"github.com/osbuild/images/pkg/blueprint"
+)
+
+type BuildConfig struct {
+	Blueprint *blueprint.Blueprint `json:"blueprint,omitempty" toml:"blueprint"`
```
achilleas-k, line 16: For a follow-up:
Perhaps we should import the substructs of the blueprint we support in BIB and let the structure validate and error out on the features we don't support.

Though that would make for bad error messaging probably.  It might be easier to just check for things we don't support and write a more useful error message that points to docs with a list of the things we do support.

achilleas-k (approved):
Excellent.  LGTM!

bib/internal/buildconfig/config.go
mvo5: Sounds good! Added to the todo.

PR #395, opened 2024-04-26
buildconfig: make buildconfig blueprint compatible and  support legacy

mvo5:
Make the buildconfig a blueprint but keep support for the legacy
format for now.

[build on top of https://github.com/osbuild/bootc-image-builder/pull/392]

### DIFF
--- a/bib/internal/buildconfig/config.go
+++ b/bib/internal/buildconfig/config.go
@@ -1,6 +1,7 @@
 package buildconfig
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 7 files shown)


bib/internal/buildconfig/config.go
```diff
@@ -1,26 +1,45 @@
...
 	"github.com/osbuild/images/pkg/blueprint"
 )
 
-type BuildConfig struct {
-	Blueprint *blueprint.Blueprint `json:"blueprint,omitempty" toml:"blueprint"`
+type legacyBuildConfig struct {
+	Blueprint *json.RawMessage `json:"blueprint"`
 }
 
+type BuildConfig blueprint.Blueprint
```
mvo5, line 22: We could just use blueprint.Blueprint everywhere instead of build config, it will make the diff a bit bigger this is why I went with the type alias but I have no super strong opinions here, either way is fine for me.

bib/internal/buildconfig/config.go
```diff
@@ -1,26 +1,48 @@
...
+	}
+
+	// support for legacy json before 2024/05
+	var legacyBC legacyBuildConfig
+	if err := json.Unmarshal(content, &legacyBC); err == nil {
+		if legacyBC.Blueprint != nil {
+			logrus.Warningf("Using legacy config")
+			content = *legacyBC.Blueprint
+		}
+	}
```
achilleas-k, line 43: This is great, thank you!

achilleas-k (approved):
Thanks you.  I really like this.

PR #397, opened 2024-04-26
bib: tweak target arch check

mvo5:
In the more complex world of the manifest lists the native target arch check of bib does no longer work. Rely on the resolver to give us the right architecture.

Thanks to Charlie Drage for reporting the issue.

See also https://github.com/osbuild/images/pull/595 and https://github.com/osbuild/images/pull/635

### DIFF
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -75,25 +75,6 @@ func canChownInPath(path string) (bool, error) {
 	return checkFile.Chown(osGetuid(), osGetgid()) == nil, nil
 }
 
-// getContainerArch returns the architecture of an already pulled container image
-func getContainerArch(imgref string) (cntArch arch.Arch, err error) {
-	outputB, err := exec.Command("podman", "image", "inspect", imgref, "--format", "{{.Architecture}}").Output()
-	if err != nil {
-		return 0, fmt.Errorf("failed inspect image for architecture: %w", util.OutputErr(err))
-	}
-	output := strings.TrimSpace(string(outputB))
-
-	// TODO: make images:arch.FromString() return an error
-	defer func() {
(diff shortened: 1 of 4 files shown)


kingsleyzissou (dismissed):
🙏

cdrage:
I can confirm that using the latest git pull of https://github.com/osbuild/images and building this, I was able to test this against both ARM64 and AMD64 and manifests are now able to resolve correctly.

Ex. using:

```sh
podman run \
    --rm \
    -it \
    --privileged \
    --pull=newer \
    -v $(pwd)/output/amd64:/output \
    -v /var/lib/containers/storage:/var/lib/containers/storage \
    localhost/bib-manifest-test:latest \
    --type raw --target-arch amd64 \
    --local \
    localhost/manifest-test
    ```
    
I was able to use the amd64 image from within the manifest no problem now! No more erroring out.

kingsleyzissou:
Ah I see why the test is failing:
- we were erroring out far earlier before
- we need a bootable container now because `internal/containers/containers` tries to mount the container and fails

mvo5:
> * we need a bootable container now because `internal/containers/containers` tries to mount the container and fails

Thank you so much! I merged your commit and pushed to this PR, fingers crossed!

kingsleyzissou:
> > * we need a bootable container now because `internal/containers/containers` tries to mount the container and fails
> 
> Thank you so much! I merged your commit and pushed to this PR, fingers crossed!

You're welcome! Thanks for your help, it passed here:
https://github.com/osbuild/bootc-image-builder/actions/runs/8853346770/job/24313988363?pr=399
So finger's crossed

kingsleyzissou (approved):
:heart:

cdrage:
image

Thank you all for all these quick fixes 🚀

PR #400, opened 2024-04-27
Provide better error for containers without a rootfs default

cgwalters:
For the generic Fedora base image the current plan is we won't include a default rootfs; that will be up to derived containers.

Output a slightly more useful error message in this case.

### DIFF
--- a/bib/internal/container/container.go
+++ b/bib/internal/container/container.go
@@ -165,6 +165,9 @@ func (c *Container) RootfsType() (string, error) {
 	// bcachefs
 	supportedFS := []string{"ext4", "xfs"}
 
+	if fsType == "" {
+		return "", fmt.Errorf("container does not include a default root filesystem type")
+	}
 	if !slices.Contains(supportedFS, fsType) {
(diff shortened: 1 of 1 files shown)


bib/internal/container/container.go
```diff
@@ -165,6 +165,9 @@ func (c *Container) RootfsType() (string, error) {
 	// bcachefs
 	supportedFS := []string{"ext4", "xfs"}
 
+	if fsType == "" {
+		return "", fmt.Errorf("container does not include a default root filesystem type")
```
mvo5, line 169: Thanks! I guess there is currently no way to also include some advise for the user what do about this or what it means? Something like `container does not include a default root filesystem type, please use --rootfs ext4|xfs to select` (which of course we do not have right now). Or `container does not include a default root filesystem type, please use the --config to select one` (which needs https://github.com/osbuild/bootc-image-builder/pull/124).

So maybe a TODO here that the user should be able to select a rootfs here? @ondrejbudai it seems we should try to get 124 in, wdyt?

mvo5 (approved):
Thank you! I'm approving as this nice! We (bib team) can deal with the (potential) followup work

PR #409, opened 2024-04-30
Enable testing on F40

ondrejbudai:
I want to see what CI thinks about the new F40-based images. :)


cgwalters:
Yeah, looks like we'll need to bridge the rootfs thing here.  And tmt ENOSPC.

Issue #401, opened 2024-04-28
testing farm ENOSPC again


cgwalters:
Testing Farm is dying on ENOSPC again.

There's lots of potential fixes here; 

- Just get systems provisioned with larger disks (incidentally we should dig in and ensure they're using efficient ephemeral storage and not stock EBS etc)
- "shard" the tests (a good description is [here](https://nexte.st/book/partitioning))
- Ensure we're flushing state between tests (e.g. are we leaking generated disk images?)

ondrejbudai:
Debugging in #414

PR #414, opened 2024-05-02
tmt: fix testing farm test execution

mvo5:
Testing farm was broken for a while, it seems we got the wrong disksize for unkown reasons. This is now fixed by switching to a different disk request syntax.

Closes: https://github.com/osbuild/bootc-image-builder/issues/401

### DIFF
--- a/plans/all.fmf
+++ b/plans/all.fmf
@@ -30,7 +30,7 @@ execute:
     pip install --user -r test/requirements.txt
     echo "Run tests"
     pytest --force-aws-upload
-  duration: 2h
+  duration: 3h
 finish:
   how: shell
   script: df -h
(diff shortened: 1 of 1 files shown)


mvo5:
Fwiw, first finding is that we do not get the disksize we ask for:
```
$    df -h / /var/tmp/ || true
Filesystem      Size  Used Avail Use% Mounted on
/dev/vda2        40G  3.3G   35G   9% /
```

plans/all.fmf
```diff
@@ -5,7 +5,8 @@ provision:
   # 120 means 120GB, disk has to be specificed here, the format from:
   #   https://tmt.readthedocs.io/en/stable/spec/hardware.html#disk
   # does not work here or below under "hardware"
-  disk: 120
+  disk:
```
ondrejbudai, line 8: Oh, shouldn't this be under the hardware key? Look at the URL: https://tmt.readthedocs.io/en/stable/spec/hardware.html#disk

achilleas-k (dismissed):

achilleas-k (approved):

PR #415, opened 2024-05-02
README.md: Link upstream docs, default to --local

cgwalters:
- Link upstream docs, as there's a lot more there around things like building container images, etc.
- Switch to using `--local` by default because it avoids foot-guns around images which require authentication, which we expect to be common in general

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,8 @@
 # bootc-image-builder
 
-A container to create disk-images from [bootc](https://github.com/containers/bootc) container inputs.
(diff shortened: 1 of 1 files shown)


achilleas-k (approved):
LGTM.

The option is likely changing with #418, but we can change the docs in that PR to match and have these in the meantime

kingsleyzissou:
Nice, I'll rebase my PR on this and add the changes to the README :)

Issue #162, opened 2024-01-29
support running as a "controller"


cgwalters:
A good pattern that the Kubernetes world has introduced is that of "spec/status" and "controllers/operators": https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md

I think what would make sense here is to teach bib how to operate as a "controller" particularly in IaaS (KubeVirt, GCP, AWS, Azure, etc.) where basically it queries the container image, and the target disk image, and runs to update the target disk image *only* if it is out of date with respect to the container.

This would require us to inject and manage metadata, etc.

PR #418, opened 2024-05-03
Replace `--local` with `--pull=[always,never,missing]` (HMS-3792)

kingsleyzissou:
This is a small POC to see how we can switch from using the `--local` flag to `--pull` which is both more descriptive and more inline with how podman behaves. This ~is a breaking change~ changes the existing behaviour, but it makes sense as it feels more intuitive and feels better from UX point-of-view.

This PR introduces 3 cases:
- always, which always pulls the container image
- missing, only pulls the container image when it is missing
- never, don't pull the image (which is the same as the previous `--local` flag)

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,8 +33,6 @@ by adding a volume-mount for the local file as well as the `--config` flag to th
 The following command will create a QCOW2 disk image. First, create `./config.toml` as described above to configure user access.
 
 ```bash
-# Ensure the image is fetched
-sudo podman pull quay.io/centos-bootc/centos-bootc:stream9
 sudo podman run \
     --rm \
     -it \
@@ -43,11 +41,17 @@ sudo podman run \
     --security-opt label=type:unconfined_t \
     -v $(pwd)/config.toml:/config.toml \
     -v $(pwd)/output:/output \
+    -v /var/lib/containers/storage:/var/lib/containers/storage \
     quay.io/centos-bootc/bootc-image-builder:latest \
     --type qcow2 \
-    --local \
+    --pull=always \ # Ensure the image is fetched
     quay.io/centos-bootc/centos-bootc:stream9
 ```
+
+The `--pull` flag is based on the behaviour of `podman-build(1)` and has the following three options:
+- `--pull=always`, always pull the remote image
+- `--pull=missing`, only pull the image if it is not in local storage
+- `--pull=never`, don't pull the image (this is functionally the same as the deprecated `--local` flag)
 
 ### Running the resulting QCOW2 file on Linux (x86_64)
 
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -585,6 +585,7 @@ func run() error {
 	manifestCmd.Flags().Bool("local", false, "use a local container rather than a container from a registry")
 	manifestCmd.Flags().String("rootfs", "", "Root filesystem type. If not given, the default configured in the source container image is used.")
 	manifestCmd.Flags().String("pull", string(pullPolicyAlways), "pull policy ['always', 'never']")
+	_ = manifestCmd.Flags().MarkDeprecated("local", "use pull=never instead")
 
 	buildCmd.Flags().AddFlagSet(manifestCmd.Flags())
 	buildCmd.Flags().String("aws-ami-name", "", "name for the AMI in AWS (only for type=ami)")
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -30,12 +30,26 @@ import (
 	"github.com/osbuild/bootc-image-builder/bib/internal/util"
 )
 
+type pullPolicy string
+
+func newPullPolicy(s string) (pullPolicy, error) {
+	switch s {
+	case "always", "never":
+		return pullPolicy(s), nil
+	default:
+		return "", fmt.Errorf("unknown pull policy: %q", s)
+	}
+}
+
 const (
 	// As a baseline heuristic we double the size of
 	// the input container to support in-place updates.
 	// This is planned to be more configurable in the
 	// future.
 	containerSizeToDiskSizeMultiplier = 2
+
+	pullPolicyAlways pullPolicy = "always"
+	pullPolicyNever  pullPolicy = "never"
 )
 
 // all possible locations for the bib's distro definitions
@@ -178,6 +192,18 @@ func manifestFromCobra(cmd *cobra.Command, args []string) ([]byte, *mTLSConfig, 
 	tlsVerify, _ := cmd.Flags().GetBool("tls-verify")
 	localStorage, _ := cmd.Flags().GetBool("local")
 	rootFs, _ := cmd.Flags().GetString("rootfs")
+	pullPolicyFlag, _ := cmd.Flags().GetString("pull")
+
+	if localStorage {
+		// don't error out since --local and
+		// --pull=never are functionally equivalent
+		pullPolicyFlag = string(pullPolicyNever)
+	}
+
+	imagePullPolicy, err := newPullPolicy(pullPolicyFlag)
+	if err != nil {
+		return nil, nil, err
+	}
 
 	if targetArch != "" && arch.FromString(targetArch) != arch.Current() {
 		// TODO: detect if binfmt_misc for target arch is
@@ -193,10 +219,11 @@ func manifestFromCobra(cmd *cobra.Command, args []string) ([]byte, *mTLSConfig, 
 	}
 	// TODO: add "target-variant", see https://github.com/osbuild/bootc-image-builder/pull/139/files#r1467591868
 
-	if localStorage {
-		if err := setup.ValidateHasContainerStorageMounted(); err != nil {
-			return nil, nil, fmt.Errorf("local storage not working, did you forget -v /var/lib/containers/storage:/var/lib/containers/storage? (%w)", err)
-		}
+	// since either pulling containers into local storage or
+	// using a container from local storage, we might want local storage
+	// to always be mounted
+	if err := setup.ValidateHasContainerStorageMounted(); err != nil {
+		return nil, nil, fmt.Errorf("local storage not working, did you forget -v /var/lib/containers/storage:/var/lib/containers/storage? (%w)", err)
 	}
 
 	buildType, err := NewBuildType(imgTypes)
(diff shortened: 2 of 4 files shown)


cgwalters:
> This is a breaking change,

Is it?  Looks to me like you're just deprecating `--local` instead of breaking it, which seems fine.

kingsleyzissou:
Valid point

achilleas-k:
Yeah, this is nice.  Gives the user more control and it aligns with what we're doing internally (and how podman works) so it should be easier to understand.

bib/cmd/bootc-image-builder/main.go
```diff
@@ -30,12 +30,18 @@ import (
...
+type pullPolicy string
+
 const (
 	// As a baseline heuristic we double the size of
 	// the input container to support in-place updates.
 	// This is planned to be more configurable in the
 	// future.
 	containerSizeToDiskSizeMultiplier = 2
+
+	always  pullPolicy = "always"
```
mvo5, line 42: (nitpick) those names are quite generic, maybe worth prefixing with pullPolicy? so `pullPolicyAlways, pullPolicyNever, pullPolicyMissing` ?

bib/cmd/bootc-image-builder/main.go
```diff
@@ -178,6 +184,13 @@ func manifestFromCobra(cmd *cobra.Command, args []string) ([]byte, *mTLSConfig,
 	tlsVerify, _ := cmd.Flags().GetBool("tls-verify")
 	localStorage, _ := cmd.Flags().GetBool("local")
 	rootFs, _ := cmd.Flags().GetString("rootfs")
+	pull, _ := cmd.Flags().GetString("pull")
+
+	if localStorage {
+		// don't error out since --local and
+		// --pull=never are functionally equivalent
+		pull = string(never)
```
mvo5, line 192: Maybe could be a bit stricter with the error checking here about the inputs we got from pull=<input>? To ensure that only valid values are input? And maybe error if pull and localStorage are set to conflicting values?

As a strawman for the validation, maybe something like (but there are many ways to do this, another alternative might be https://pkg.go.dev/github.com/thediveo/enumflag#section-readme to make sure that the inputs are checked on the cobra level).
```go
index 94841f9..d6982e9 100644
@@ -32,6 +32,15 @@ import (
 
 type pullPolicy string
 
+func newPullPolicy(s string) (pullPolicy, error) {
+       switch pullPolicy(s) {
+       case pullPolicyAlways, pullPolicyNever, pullPolicyMissing:
+               return pullPolicy(s), nil
+       default:
+               return "", fmt.Errorf("unknown pull policy %q", s)
+       }
+}
+
 const (
        // As a baseline heuristic we double the size of
        // the input container to support in-place updates.
@@ -39,9 +48,9 @@ const (
        // future.
        containerSizeToDiskSizeMultiplier = 2
 
-       always  pullPolicy = "always"
-       never   pullPolicy = "never"
-       missing pullPolicy = "missing"
+       pullPolicyAlways  pullPolicy = "always"
+       pullPolicyNever   pullPolicy = "never"
+       pullPolicyMissing pullPolicy = "missing"
 )
 
 // all possible locations for the bib's distro definitions
@@ -184,12 +193,16 @@ func manifestFromCobra(cmd *cobra.Command, args []string) ([]byte, *mTLSConfig,
        tlsVerify, _ := cmd.Flags().GetBool("tls-verify")
        localStorage, _ := cmd.Flags().GetBool("local")
        rootFs, _ := cmd.Flags().GetString("rootfs")
-       pull, _ := cmd.Flags().GetString("pull")
+       pullFlag, _ := cmd.Flags().GetString("pull")
 
        if localStorage {
                // don't error out since --local and
                // --pull=never are functionally equivalent
-               pull = string(never)
+               pullFlag = string(pullPolicyNever)
+       }
+       imagePullPolicy, err := newPullPolicy(pullFlag)
+       if err != nil {
+               return nil, nil, err
        }
 
        if targetArch != "" && arch.FromString(targetArch) != arch.Current() {
@@ -248,11 +261,10 @@ func manifestFromCobra(cmd *cobra.Command, args []string) ([]byte, *mTLSConfig,
                return arch.FromString(a) == cntArch
        }
 
-       imagePullPolicy := pullPolicy(pull)
        // If --pull=always or if --pull=missing and the container image is missing,
        // pull the container into local storage. This behaviour is more in line
        // with podman
-       if imagePullPolicy == always || (imagePullPolicy == missing && !containerImageExists(imgref, cntArch)) {
+       if imagePullPolicy == pullPolicyAlways || (imagePullPolicy == pullPolicyMissing && !containerImageExists(imgref, cntArch)) {
                logrus.Infof("Pulling image %s (arch=%s)\n", imgref, cntArch)
                cmd := exec.Command(
                        "podman",
@@ -599,7 +611,7 @@ func run() error {
        manifestCmd.Flags().String("rootfs", "", "Root filesystem type. If not given, the default configured in the source container image is used.")
-       manifestCmd.Flags().String("pull", string(missing), "pull policy ['always', 'never', 'missing]")
+       manifestCmd.Flags().String("pull", string(pullPolicyMissing), "pull policy ['always', 'never', 'missing]")
        _ = manifestCmd.Flags().MarkDeprecated("local", "use pull=never instead")
 
        buildCmd.Flags().AddFlagSet(manifestCmd.Flags())
```

mvo5 (commented):
Thank you! Some quick drive-by ideas inline, I would also love to see a test but we can chat next week, the existing test_opts.py hopefully makes this straightforward (but I need to double check :)

bib/cmd/bootc-image-builder/main.go
kingsleyzissou: Yeah good idea!

bib/cmd/bootc-image-builder/main.go
kingsleyzissou: Thanks, yeah I like this. I'll update the PR :)

README.md
```diff
@@ -43,12 +41,18 @@ sudo podman run \
     --security-opt label=type:unconfined_t \
     -v $(pwd)/config.toml:/config.toml \
     -v $(pwd)/output:/output \
+    -v /var/lib/containers/storage:/var/lib/containers/storage \
     quay.io/centos-bootc/bootc-image-builder:latest \
     --type qcow2 \
-    --local \
+    --pull=always \ # Ensure the image is fetched
```
rhatdan, line 47: PLEASE make this missing not always.

README.md
```diff
@@ -43,12 +41,18 @@ sudo podman run \
...
     quay.io/centos-bootc/centos-bootc:stream9
 ```
 
+The `--pull` flag is based on the behaviour of `podman-build(1)` and has the following three options:
+- `--pull=always`, always pull the remote image
+- `--pull=missing`, only pull the image if it is not in local storage
+- `--pull=never`, don't pull the image (this is functionally the same as the deprecated `--local` flag)
```
rhatdan, line 54: Podman build also supports newer.

$ podman run --help | grep -- --pull

rhatdan: $ podman build --help | grep -- --pull 
      --pull string[="true"]                         Pull image policy ("always/true"|"missing"|"never/false"|"newer") (default "missing")

bib/cmd/bootc-image-builder/main.go
```diff
@@ -30,12 +30,27 @@ import (
 	"github.com/osbuild/bootc-image-builder/bib/internal/util"
 )
 
+type pullPolicy string
+
+func newPullPolicy(s string) (pullPolicy, error) {
+	switch s {
+	case "always", "never":
```
rhatdan, line 37: What about missing and newer?

test/test_build.py
```diff
@@ -256,7 +256,7 @@ def build_images(shared_tmpdir, build_container, request, force_aws_upload):
             *types_arg,
             *upload_args,
             *target_arch_args,
-            "--local" if local else "--local=false",
+            "--pull=never" if local else "--pull=always",
```
rhatdan, line 259: s/always/missing/g

cgwalters:
OK this topic is subtle and confusing.  I am coming around to the PoV that the *only* viable solution is to change all the documentation for bib to say:

```
podman pull <image>  # If not present
podman run quay.io/bootc-image-builder ... <image>
```

And we should *not* expose `--pull` as an argument to bib itself. Here's why:

While today bib is invoking `podman pull` - it's doing so *inside the container context itself*.  In a scenario like Podman Machine on macOS - the default thing that happens when the user types `podman` is that the pull secret is passed via remoting from the client to the Linux VM.  The pull secret is sitting out there on the user's home directory, and is not present in the "system context" in the Linux VM except via the virtiofs/9p bind mount.

Trying to replicate the effect of that remoting work from inside is of course technically possible, but I suspect is going to be quite fiddly.

Also as I've been arguing in a few contexts for a while, IMO one of the *primary* entrypoints to bib should be users building their own derived containers they used locally - for which there's no need to pull at all.  

The case of building a "remote" image of course definitely exists - in fact it's very much a "production" use case I think for e.g. having one build pipeline (Tekton, Actions, etc.) churn out container images, and then a distinct pipeline asynchronously generating disk images by polling the registry (cc https://github.com/osbuild/bootc-image-builder/issues/162 ).

But anyways there again - it's simple, the pipeline itself just does a `podman pull` of the image before invoking bib.

cgwalters:
Or to say :arrow_up: another way: all this "bib controlling pulls" superficially appears to work except when dealing with images that require authentication, in which case it's all just broken.  And we shouldn't make the authenticated case feel broken.

bib/cmd/bootc-image-builder/main.go
kingsleyzissou: Thanks good catch re missing. We still need to add more tests and figuring out how to approach newer

cgwalters (changes requested):
Marking as requested-changes unless someone can argue against the authenticated image problem.

ondrejbudai:
Just to zoom out a bit: The core problem is that we want to do authenticated pulls on the host from a container (bib). The usual pattern used in the wild is to mount the docker/podman socket inside your application container and control the host's container engine remotely. See e.g. [watchtower](https://containrrr.dev/watchtower/).

On the other hand, the socket approach isn't that common in the podman world (except for remote use-cases), it feels cumbersome to set up, and I actually don't know at all how it would work in case of podman machine.

Additionally, bib doesn't only need to pull images, but it also needs access to the image's underlying layers, which means that even if wanted to experiment with using a socket, we would still need to mount the storage inside.

I just wanted to mention sockets for the sake of completeness, but I fully acknowledge that this is not the right way.

---

I think that having `--pull=XYZ` has some benefits (I love one-liners), but I do agree that establishing the following pattern might be better long-term, as there's just one mental path for everything.

```
podman pull <image>  # If not present
podman run quay.io/bootc-image-builder ... <image>
```

achilleas-k:
FWIW, I like the simplicity of limiting the behaviour of BIB to local containers only (no pulling ever) and letting the user take care of having the containers locally, either by building them or pulling them ahead of time.  It reduces the complexity of the project as well as the mental load for users trying to wrap their heads around what BIB is doing when pulling, whether it needs authentication or not, etc.

cgwalters:
The usual pattern used in the wild is to mount the docker/podman socket inside your application container and control the host's container engine remotely. See e.g. [watchtower](https://containrrr.dev/watchtower/).

I don't think interacting with the socket helps, it gives the same semantics as just running the `podman` CLI.

Things like watchtower will generally assume that you've configured the pull secret for the container image globally for the container engine, AFAICS.

The problem is the Podman-machine case again; if you follow the default developer UX of running e.g. `podman login + podman pull` etc. on the host system, it will not work to do `podman machine ssh` and do `podman pull` from there, because the auth file is on the host and podman-in-vm doesn't know about it.

rhatdan:
I agree simplify BIB and only work on local storage.  Force the caller to pull the image first or build the image locally.

PR #423, opened 2024-05-07
Deprecate `--local` flag (HMS-3792)

kingsleyzissou:
Pulling container images before building would break in the case of authenticated images on podman machine, since the auth file lives on the host and podman machine and won't know about it.

This commit deprecates the `--local` flag and puts the requirement on the user to ensure that the container is in local storage before initiating the build.

### DIFF
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -194,7 +194,6 @@ func manifestFromCobra(cmd *cobra.Command, args []string, pbar progress.Progress
 	imgTypes, _ := cmd.Flags().GetStringArray("type")
 	rpmCacheRoot, _ := cmd.Flags().GetString("rpmmd")
 	targetArch, _ := cmd.Flags().GetString("target-arch")
-	tlsVerify, _ := cmd.Flags().GetBool("tls-verify")
 	rootFs, _ := cmd.Flags().GetString("rootfs")
 	useLibrepo, _ := cmd.Flags().GetBool("use-librepo")
 
@@ -305,7 +304,6 @@ func manifestFromCobra(cmd *cobra.Command, args []string, pbar progress.Progress
 		Config:         config,
 		ImageTypes:     imageTypes,
 		Imgref:         imgref,
-		TLSVerify:      tlsVerify,
 		RootfsMinsize:  cntSize * containerSizeToDiskSizeMultiplier,
 		DistroDefPaths: distroDefPaths,
 		SourceInfo:     sourceinfo,
@@ -653,7 +651,10 @@ func buildCobraCmdline() (*cobra.Command, error) {
 	rootCmd.AddCommand(versionCmd)
 
 	rootCmd.AddCommand(manifestCmd)
-	manifestCmd.Flags().Bool("tls-verify", true, "require HTTPS and verify certificates when contacting registries")
+	manifestCmd.Flags().Bool("tls-verify", false, "DEPRECATED: require HTTPS and verify certificates when contacting registries")
+	if err := manifestCmd.Flags().MarkHidden("tls-verify"); err != nil {
+		return nil, fmt.Errorf("cannot hide 'tls-verify' :%w", err)
+	}
 	manifestCmd.Flags().String("rpmmd", "/rpmmd", "rpm metadata cache directory")
 	manifestCmd.Flags().String("target-arch", "", "build for the given target architecture (experimental)")
 	manifestCmd.Flags().StringArray("type", []string{"qcow2"}, fmt.Sprintf("image types to build [%s]", imagetypes.Available()))
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -195,9 +195,19 @@ func manifestFromCobra(cmd *cobra.Command, args []string, pbar progress.Progress
 	rpmCacheRoot, _ := cmd.Flags().GetString("rpmmd")
 	targetArch, _ := cmd.Flags().GetString("target-arch")
 	tlsVerify, _ := cmd.Flags().GetBool("tls-verify")
-	localStorage, _ := cmd.Flags().GetBool("local")
 	rootFs, _ := cmd.Flags().GetString("rootfs")
 	useLibrepo, _ := cmd.Flags().GetBool("use-librepo")
+
+	// If --local was given, warn in the case of --local or --local=true (true is the default), error in the case of --local=false
+	if cmd.Flags().Changed("local") {
+		localStorage, _ := cmd.Flags().GetBool("local")
+		if localStorage {
+			fmt.Fprintf(os.Stderr, "WARNING: --local is now the default behavior, you can remove it from the command line\n")
+		} else {
+			return nil, nil, fmt.Errorf(`--local=false is no longer supported, remove it and make sure to pull the container before running bib:
+	sudo podman pull %s`, imgref)
+		}
+	}
 
 	if targetArch != "" && arch.FromString(targetArch) != arch.Current() {
 		// TODO: detect if binfmt_misc for target arch is
@@ -213,10 +223,8 @@ func manifestFromCobra(cmd *cobra.Command, args []string, pbar progress.Progress
 	}
 	// TODO: add "target-variant", see https://github.com/osbuild/bootc-image-builder/pull/139/files#r1467591868
 
-	if localStorage {
-		if err := setup.ValidateHasContainerStorageMounted(); err != nil {
-			return nil, nil, fmt.Errorf("local storage not working, did you forget -v /var/lib/containers/storage:/var/lib/containers/storage? (%w)", err)
-		}
+	if err := setup.ValidateHasContainerStorageMounted(); err != nil {
+		return nil, nil, fmt.Errorf("could not access container storage, did you forget -v /var/lib/containers/storage:/var/lib/containers/storage? (%w)", err)
 	}
 
 	imageTypes, err := imagetypes.New(imgTypes...)
@@ -227,27 +235,6 @@ func manifestFromCobra(cmd *cobra.Command, args []string, pbar progress.Progress
 	config, err := buildconfig.ReadWithFallback(userConfigFile)
 	if err != nil {
 		return nil, nil, fmt.Errorf("cannot read config: %w", err)
-	}
-
-	// If --local wasn't given, always pull the container.
-	// If the user mount a container storage inside bib (without --local), the code will try to pull
-	// a newer version of the container even if an older one is already present. This doesn't match
-	// how `podman run`` behaves by default, but it matches the bib's behaviour before the switch
-	// to using containers storage in all code paths happened.
-	// We might want to change this behaviour in the future to match podman.
-	if !localStorage {
-		logrus.Infof("Pulling image %s (arch=%s)\n", imgref, cntArch)
-		cmd := exec.Command("podman", "pull", "--arch", cntArch.String(), fmt.Sprintf("--tls-verify=%v", tlsVerify), imgref)
-		// podman prints progress on stderr so connect that to give
-		// better UX. But do not connect stdout as "bib manifest"
-		// needs to be purely json so any stray output there is a
-		// problem
-		cmd.Stderr = os.Stderr
-		if err := cmd.Run(); err != nil {
-			return nil, nil, fmt.Errorf("failed to pull container image: %w", util.OutputErr(err))
-		}
-	} else {
-		logrus.Debug("Using local container")
 	}
 
 	pbar.SetPulseMsgf("Manifest generation step")
@@ -670,7 +657,10 @@ func buildCobraCmdline() (*cobra.Command, error) {
 	manifestCmd.Flags().String("rpmmd", "/rpmmd", "rpm metadata cache directory")
 	manifestCmd.Flags().String("target-arch", "", "build for the given target architecture (experimental)")
 	manifestCmd.Flags().StringArray("type", []string{"qcow2"}, fmt.Sprintf("image types to build [%s]", imagetypes.Available()))
-	manifestCmd.Flags().Bool("local", false, "use a local container rather than a container from a registry")
+	manifestCmd.Flags().Bool("local", true, "DEPRECATED: --local is now the default behavior, make sure to pull the container image before running bootc-image-builder")
+	if err := manifestCmd.Flags().MarkHidden("local"); err != nil {
+		return nil, fmt.Errorf("cannot hide 'local' :%w", err)
+	}
 	manifestCmd.Flags().String("rootfs", "", "Root filesystem type. If not given, the default configured in the source container image is used.")
 	manifestCmd.Flags().Bool("use-librepo", false, "(experimenal) switch to librepo for pkg download, needs new enough osbuild")
 	// --config is only useful for developers who run bib outside
--- a/test/test_build.py
+++ b/test/test_build.py
@@ -416,6 +416,9 @@ def build_images(shared_tmpdir, build_container, request, force_aws_upload, gpg_
             ]
             cmd.extend(signed_image_args)
 
+            # Pull the signed image
+            testutil.pull_container(container_ref, tls_verify=False)
+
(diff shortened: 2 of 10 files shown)


bib/cmd/bootc-image-builder/main.go
```diff
@@ -193,10 +192,8 @@ func manifestFromCobra(cmd *cobra.Command, args []string) ([]byte, *mTLSConfig,
 	}
 	// TODO: add "target-variant", see https://github.com/osbuild/bootc-image-builder/pull/139/files#r1467591868
 
-	if localStorage {
-		if err := setup.ValidateHasContainerStorageMounted(); err != nil {
-			return nil, nil, fmt.Errorf("local storage not working, did you forget -v /var/lib/containers/storage:/var/lib/containers/storage? (%w)", err)
-		}
+	if err := setup.ValidateHasContainerStorageMounted(); err != nil {
+		return nil, nil, fmt.Errorf("local storage not working, did you forget -v /var/lib/containers/storage:/var/lib/containers/storage? (%w)", err)
```
cgwalters, line 196: Now that we're always local, how about
```suggestion
		return nil, nil, fmt.Errorf("could not access container storage, did you forget -v /var/lib/containers/storage:/var/lib/containers/storage? (%w)", err)
```

bib/cmd/bootc-image-builder/main.go
```diff
@@ -554,6 +536,10 @@ func run() error {
 	manifestCmd.Flags().StringArray("type", []string{"qcow2"}, fmt.Sprintf("image types to build [%s]", allImageTypesString()))
 	manifestCmd.Flags().Bool("local", false, "use a local container rather than a container from a registry")
 	manifestCmd.Flags().String("rootfs", "", "Root filesystem type. If not given, the default configured in the source container image is used.")
+	err := manifestCmd.Flags().MarkDeprecated("local", "please ensure the container image has already been pulled")
```
cgwalters, line 539: This emits a warning when the user passes it, right?  I feel like it's going to be annoying for people...let's just silently accept it for a while; like at least on the order of ~months until we have a chance to update the docs?

cgwalters (commented):
Overall LGTM just two nits

bib/cmd/bootc-image-builder/main.go
kingsleyzissou: Sure, I'll update in a bit. Just wanted to see if the tests all pass first in CI

test/test_build.py
```diff
@@ -79,18 +79,22 @@ def image_type_fixture(shared_tmpdir, build_container, request, force_aws_upload
     """
     container_ref, images, target_arch, local = parse_request_params(request)
 
-    if local:
-        cont_tag = "localhost/cont-base-" + "".join(random.choices(string.digits, k=12))
+    if not local:
```
mvo5, line 82: I wonder if we could avoid one `if` here by simply (re)using the `container_ref` var (which is probably a slightly better name anyway), something like:
```python
diff --git a/test/test_build.py b/test/test_build.py
index 9413949..4d0d9db 100644
--- a/test/test_build.py
+++ b/test/test_build.py
@@ -79,21 +79,18 @@ def image_type_fixture(shared_tmpdir, build_container, request, force_aws_upload
     """
     container_ref, images, target_arch, local = parse_request_params(request)
 
-    if not local:
-        cont_tag = container_ref
-    else:
-        cont_tag = "localhost/cont-base-" + \
+    if local:
+        container_ref = "localhost/cont-base-" + \
             "".join(random.choices(string.digits, k=12))
-
         # we are not cross-building local images (for now)
-        request.param = ",".join([cont_tag, images, "", "true"])
+        request.param = ",".join([cont_ref, images, "", "true"])
 
     # copy the container into containers-storage
     # for all builds
     subprocess.check_call([
         "skopeo", "copy",
         f"docker://{container_ref}",
-        f"containers-storage:[overlay@/var/lib/containers/storage+/run/containers/storage]{cont_tag}"
+        f"containers-storage:[overlay@/var/lib/containers/storage+/run/containers/storage]{container_ref}"
     ])
 
     with build_images(shared_tmpdir, build_container, request, force_aws_upload) as build_results:
```

mvo5 (commented):
Thanks for this! Looks good, one tiny idea/suggestion inline.

test/test_build.py
kingsleyzissou: Oh yeah good suggestion! Thanks

achilleas-k (dismissed):
The joy and simplicity of a mostly-red PR :)

bib/cmd/bootc-image-builder/main.go
```diff
@@ -193,10 +191,8 @@ func manifestFromCobra(cmd *cobra.Command, args []string) ([]byte, *mTLSConfig,
 	}
 	// TODO: add "target-variant", see https://github.com/osbuild/bootc-image-builder/pull/139/files#r1467591868
 
-	if localStorage {
-		if err := setup.ValidateHasContainerStorageMounted(); err != nil {
-			return nil, nil, fmt.Errorf("local storage not working, did you forget -v /var/lib/containers/storage:/var/lib/containers/storage? (%w)", err)
-		}
+	if err := setup.ValidateHasContainerStorageMounted(); err != nil {
+		return nil, nil, fmt.Errorf("could not access container storage, did you forget -v /var/lib/containers/storage:/var/lib/containers/storage? (%w)", err)
```
achilleas-k, line 195: I'm seeing this error in the failed tests.  Did you forget ... ? :laughing:

kingsleyzissou: I did forget. I decided I will just fix it tomorrow 😅😭

mvo5 (dismissed):
Thank you!

kingsleyzissou:
`flake8` failed on line length. Think it should be all good now 🤞

mvo5 (approved):

mvo5 (dismissed):

kingsleyzissou:
Okay it's finally green now. Removing the flag also broke the cross-arch integration tests :/
Simple PR but had so many knock-on effects

mvo5 (approved):
Thank you

test/test_build.py
```diff
@@ -81,17 +81,23 @@ def image_type_fixture(shared_tmpdir, build_container, request, force_aws_upload
...
-        # copy the container into containers-storage
-        subprocess.check_call([
-            "skopeo", "copy",
-            f"docker://{container_ref}",
-            f"containers-storage:[overlay@/var/lib/containers/storage+/run/containers/storage]{cont_tag}"
-        ])
+    # copy the container into containers-storage
+    # for all builds
+    arch = "amd64"
+    if target_arch == "aarch64" or target_arch == "arm64":
```
mvo5, line 92: Hm, we probably want to raise an exception here if we come accross a case when we hit an unknown arch, just to be on the safe side.

test/test_build.py
```diff
@@ -81,17 +81,23 @@ def image_type_fixture(shared_tmpdir, build_container, request, force_aws_upload
...
+        arch = "arm64"
```
mvo5, line 93: silly question, dosn't podman understand both kernel and golang arch names?

mvo5 (commented):
One more note/question but could be followup (just defensive)

test/test_build.py
rhatdan: Yes it seems to:
```
$ podman run --arch=aarch64 alpine arch
aarch64
$ podman run --arch=arm64 alpine arch
aarch64
```

rhatdan: Perhaps we should just pass down target_arch to Podman and allow it to handle it or throw an error.

test/test_build.py
kingsleyzissou: If I understand correctly, @ondrejbudai was keen to wait until we can have tagged releases before merging this. So I could potentially do the changes now

test/test_build.py
kingsleyzissou: Cool thanks, I didn't even think to just let podman handle it. Makes a lot of sense

ondrejbudai (changes requested):
I love this! :)

Marking this as request changes, because I want to first talk to the podman desktop folks if this is alright for them, and I want to figure out if we can somehow have bib releases, so users know what version they run, and thus which flags are supported and what's the default behaviour (git SHAs aren't very friendly in this case unfortunately).

test/test_build.py
kingsleyzissou: I went a bit defensive and raised an error here but passed down the target arch as is to podman

test/test_build.py
kingsleyzissou: Turns out skopeo doesn't understand both. I will change to use podman pull instead

achilleas-k:
@ondrejbudai How does this change affect BIB in Konflux?

bib/cmd/bootc-image-builder/main.go
```diff
@@ -185,10 +185,27 @@ func manifestFromCobra(cmd *cobra.Command, args []string) ([]byte, *mTLSConfig,
...
+		if localStorage {
+			fmt.Fprintf(os.Stderr, "WARNING: --local is now the default behavior, you can remove it from the command line\n")
+		} else {
+			return nil, nil, fmt.Errorf(`--local=false is no longer supported, remove it and make sure to pull the container before running bib:
+	sudo podman pull %s`, imgref)
+		}
+	}
+
+	// --local is the default now, so there is no longer a need for --tls-verify
+	if cmd.Flags().Changed("local") {
```
mvo5, line 202: silly question, why do we need to check "local" here, could we just skip that and check for only for that?

bib/cmd/bootc-image-builder/main.go
```diff
@@ -604,11 +597,11 @@ func buildCobraCmdline() (*cobra.Command, error) {
 	rootCmd.AddCommand(versionCmd)
 
 	rootCmd.AddCommand(manifestCmd)
-	manifestCmd.Flags().Bool("tls-verify", true, "require HTTPS and verify certificates when contacting registries")
+	manifestCmd.Flags().Bool("tls-verify", false, "require HTTPS and verify certificates when contacting registries")
```
mvo5, line 600: Should we also put a big "DEPRECATED" here like we did for "local" ?

test/test_manifest.py
```diff
@@ -31,6 +31,8 @@ def find_image_size_from(manifest_str):
 
 @pytest.mark.parametrize("tc", gen_testcases("manifest"))
 def test_manifest_smoke(build_container, tc):
+    testutil.pull_container(tc.container_ref, tc.target_arch)
```
mvo5, line 34: Hm, would it make sense to make this part of the build_container fixture somehow? it seems to repeat quite a bit.

mvo5 (commented):
Thanks, this looks nice! A quick first round of reviews, if this is ready now I will do a deeper dive later today too

bib/cmd/bootc-image-builder/main.go
kingsleyzissou: Ah flip dropped the commit and re-added and forgot to add this back in. Thanks!

bib/cmd/bootc-image-builder/main.go
kingsleyzissou: I'll ping @ondrejbudai here because this was his change. I think the idea was to catch when `--local=false` was set and error out on it.

test/test_manifest.py
kingsleyzissou: Yeah good idea, let me try

bib/cmd/bootc-image-builder/main.go
mvo5: The "local" case *seems* to be covered in the lines above, this is why I'm wondering. But maybe I'm just missing something, I only had a casual look at this part :)

test/test_manifest.py
kingsleyzissou: Would it maybe make sense to make a different fixture for some of these tests rather? It doesn't look like the `build_container` fixture should be used here, right?

bib/cmd/bootc-image-builder/main.go
kingsleyzissou: Typo again 🤦 this should be `tls-verify`

test/testutil.py
```diff
@@ -140,6 +140,21 @@ def create_filesystem_customizations(rootfs: str):
...
+        target_arch = platform.machine()
+
+    if target_arch not in ["x86_64", "amd64", "aarch64", "arm64", "s390x", "ppc64le"]:
+        raise RuntimeError(f"unknown host arch: {target_arch}")
+
+    output = subprocess.run([
+        "podman", "pull",
+        "--arch", target_arch,
+        container_ref,
+    ], check=False, stdout=subprocess.PIPE, stderr=subprocess.PIPE, encoding="utf8")
```
mvo5, line 154: Why do we capture the output here? It seems we are not doing anything with it. Maybe just remove the stdout/stderr/encoding here?

mvo5 (commented):
Thanks, this looks good, small suggestions inline still more nit-picky, happy to approve.

kingsleyzissou:
> Thanks, this looks good, small suggestions inline still more nit-picky, happy to approve.

Thanks! I think the changes are good to do here, especially hiding the flags. So I'll do that and ping you for a review

test/testutil.py
```diff
@@ -140,6 +140,20 @@ def create_filesystem_customizations(rootfs: str):
...
 
 
+def pull_container(container_ref, target_arch=""):
+    if target_arch == "":
+        target_arch = platform.machine()
+
+    if target_arch not in ["x86_64", "amd64", "aarch64", "arm64", "s390x", "ppc64le"]:
+        raise RuntimeError(f"unknown host arch: {target_arch}")
+
+    output = subprocess.run([
```
mvo5, line 150: (super nitpick) now that we use `check=True` and not using the `output` anymore we might as well remove the `output = ` bit :)

mvo5 (approved):
Thank you

test/testutil.py
kingsleyzissou: The tests complained so I had to remove that :p

achilleas-k (approved):

ondrejbudai (dismissed):

ondrejbudai:
Thanks! :+1:

achilleas-k:
Merge conflicts

jpace121:
Hi! So I'm currently using bootc-image-builder and am purposely not using --local, and being forced to use --local would break the tool for me.

More specifically, on my build computer I set the `graphroot` in storage.conf to a directory on a separate very large SSD from /, since the SSD I use for / is somewhat small. This means I don't have any images in `/var/lib/containers/storage`, and I don't want to; I don't have enough storage there. bootc-image-builder hard codes looking in `/var/lib/containers/storage` when looking for/building manifests though. To workaround this, I tried mounting the host's graphroot in the bootc-image-builder container under `/var/lib/containers/storage`, but that didn't work because I guess there is some db in that directory that is aware of what directory its in, and when the directories don't line up, podman get grumpy.

If you remove the local option, would you be willing to also find the places where the graphroot is assumed to be `/var/lib/containers/storage` and provide an option to override it?

achilleas-k:
@jpace121 I think your use case can be solved by mounting both into the bootc-image-builder container.
```
podman run -v /var/lib/containers:/var/lib/containers -v /path/to/large/ssd/container/storage:/path/to/large/ssd/container/storage ...
```

Can you see if that works?

The `/var/lib/containers` mount should make BIB pick up the `storage.conf` and point it to the large container storage, which will be at the expected location inside the container.

> If you remove the local option, would you be willing to also find the places where the graphroot is assumed to be `/var/lib/containers/storage` and provide an option to override it?

If the path that graphroot points to isn't mounted in the container, there's not much we can do from inside the container itself to reach it automatically.  Unless I'm misunderstanding the suggestion.

mvo5 (approved):

achilleas-k (approved):

achilleas-k (approved):

kingsleyzissou:
One or two more of the tests needed to pull the containers first, I updated Ondrej's commit with it.

mvo5 (approved):
Thank you

achilleas-k (approved):

PR #442, opened 2024-05-19
Add missing volume mount to README

sshnaidm:
The current command fails if no mounted containers storage, add podman mount to fix it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ sudo podman run \
     --security-opt label=type:unconfined_t \
     -v $(pwd)/config.toml:/config.toml \
     -v $(pwd)/output:/output \
+    -v /var/lib/containers/storage:/var/lib/containers/storage \
     quay.io/centos-bootc/bootc-image-builder:latest \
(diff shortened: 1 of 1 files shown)


p5:
Just a thought, and this could be completely wrong, but could you replace that /var mount with the user's rootless Podman storage directory, thus building in a rootful `podman run` command using the user's rootless Podman image?

sshnaidm:
> Just a thought, and this could be completely wrong, but could you replace that /var mount with the user's rootless Podman storage directory, thus building in a rootful `podman run` command using the user's rootless Podman image?

AFAIK it doesn't support rootless podman currently.

achilleas-k:
> Just a thought, and this could be completely wrong, but could you replace that /var mount with the user's rootless Podman storage directory, thus building in a rootful `podman run` command using the user's rootless Podman image?

I haven't tried this but I would worry about root taking ownership of files in the user's container store if it writes anything to it.  Easy to test though.

Issue #387, opened 2024-04-24
Don't run expensive tests PRs that change only README.md


ondrejbudai:
Integration tests take 40 minutes and they provide absolutely no value for PRs changing only the README.

PR #479, opened 2024-06-15
Feature/no integration test readme

bearrito:
I use https://nektosact.com/ to test github actions locally, but getting that to handle pull requests like this isn't working.I'm digging into that. I can't run actions in my personal account, so I think I need to test here. 

What we could do. There might be a simpler way.
1. Create a new branch off `main`, named `fake-main` 
2. Target this PR against that. Merge it.
3. Target further test prs against `fake-main`
4. I created a new file `CANARY.md` so that I never have to touch `README.md`
5. I removed the integration tests for now so if the conditional job fails we aren't stuck waiting on that. That will be added back in.

This can be useful - https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

Closes: https://github.com/osbuild/bootc-image-builder/issues/387


mvo5:
Thanks for your PR and sorry for the slow reply (we are usually faster but devconf happend :)

I agree we need a `HACKING.md` or similar - especially since one of our goals have been to make running all tests very easy to run locally. If you have an up-to-date podman (4.x) and some test dependencies (like qemu, python3-paramiko, ...) it should be enough to run:
```
$ sudo pytest -s -vv
```
in the rootdir and it will run everything locally with reasonable output to follow what is happening.

mvo5:
I created https://github.com/osbuild/bootc-image-builder/pull/482 now with a very simple (but hopefully useful) start of a `HACKING.md`

PR #497, opened 2024-06-25
Add support for ppc64le architecture

sacsant:
bootc-image-builder lacks support for ppc64le.
This two patch series adds such support.
 - Update manifestForDiskImage to include support for ppc64le architecture
 - Update partition table map with ppc64le specific data

Signed-off-by: Sachin Sant <sachinp@linux.ibm.com>

### DIFF
--- a/bib/cmd/bootc-image-builder/partition_tables.go
+++ b/bib/cmd/bootc-image-builder/partition_tables.go
@@ -150,4 +150,41 @@ var partitionTables = distro.BasePartitionTableMap{
 			},
 		},
 	},
+	arch.ARCH_PPC64LE.String(): disk.PartitionTable{
+		UUID: "D209C89E-EA5E-4FBD-B161-B461CCE297E0",
+		Type: "gpt",
+		Partitions: []disk.Partition{
+			{
+				Size:     4 * MebiByte,
+				Type:     disk.PRePartitionGUID,
+				Bootable: true,
+			},
+			{
+				Size: 500 * MebiByte,
+				Type: disk.FilesystemDataGUID,
+				UUID: disk.FilesystemDataUUID,
+				Payload: &disk.Filesystem{
+					Type:         "ext4",
+					Mountpoint:   "/boot",
+					Label:        "boot",
+					FSTabOptions: BootOptions,
+					FSTabFreq:    1,
+					FSTabPassNo:  2,
+				},
+			},
+			{
+				Size: 2 * GibiByte,
+				Type: disk.FilesystemDataGUID,
+				UUID: disk.FilesystemDataUUID,
+				Payload: &disk.Filesystem{
+					Type:         "ext4",
+					Label:        "root",
+					Mountpoint:   "/",
+					FSTabOptions: RootOptions,
+					FSTabFreq:    1,
+					FSTabPassNo:  1,
+				},
+			},
+		},
+	},
 }
--- a/bib/cmd/bootc-image-builder/image.go
+++ b/bib/cmd/bootc-image-builder/image.go
@@ -145,6 +145,13 @@ func manifestForDiskImage(c *ManifestConfig, rng *rand.Rand) (*manifest.Manifest
 			},
 			Zipl: true,
 		}
+	case arch.ARCH_PPC64LE:
+		img.Platform = &platform.PPC64LE{
+			BasePlatform: platform.BasePlatform{
+				QCOW2Compat: "1.1",
+			},
+			BIOS: true,
+		}
 	}
 
 	if kopts := customizations.GetKernel(); kopts != nil && kopts.Append != "" {


mvo5:
Thank you for this PR! It seems we need to do a bit of work on our side first before this works. I tried the following to test it:
```
diff --git a/test/test_build.py b/test/test_build.py
index da27fbc..337b51f 100644
--- a/test/test_build.py
+++ b/test/test_build.py
@@ -285,6 +285,7 @@ def build_images(shared_tmpdir, build_container, request, force_aws_upload):
             *upload_args,
             *target_arch_args,
             "--local" if local else "--local=false",
+            "--rootfs", "ext4",
         ])
 
         # print the build command for easier tracing
diff --git a/test/testcases.py b/test/testcases.py
index e5b1f6a..81335f3 100644
--- a/test/testcases.py
+++ b/test/testcases.py
@@ -52,6 +52,8 @@ def gen_testcases(what):
             # todo: add fedora:eln
             test_cases.append(
                 f'{CONTAINERS_TO_TEST["centos"]},raw,arm64')
+            test_cases.append(
+                'quay.io/sacsant/fedora-bootc:latest,raw,ppc64le')
         elif platform.machine() == "arm64":
             # TODO: add arm64->x86_64 cross build test too
             pass
diff --git a/test/vm.py b/test/vm.py
index 1bb21f5..2297ed7 100644
--- a/test/vm.py
+++ b/test/vm.py
@@ -134,6 +134,11 @@ class QEMU(VM):
             ]
             if use_ovmf:
                 qemu_cmdline.extend(["-bios", find_ovmf()])
+        elif self._arch in ("ppc64le"):
+            qemu_cmdline = [
+                "qemu-system-ppc64le",
+                "-M", "pseries",
+            ]
         else:
             raise ValueError(f"unsupported architecture {self._arch}")
 
```
and when running the test with:
```
$ sudo pytest  -s -vv ./test/test_build.py::test_image_boots[quay.io/sacsant/fedora-bootc:latest,raw,ppc64le]
...
panic: required mounts for bootupd stage [/boot/efi] missing

goroutine 1 [running]:
github.com/osbuild/images/pkg/manifest.(*RawOSTreeImage).addBootupdStage(0xc00052a7e0, 0xc0007d0938)
	/root/go/pkg/mod/github.com/osbuild/images@v0.61.0/pkg/manifest/raw_ostree.go:127 +0x2a5
....
```
so our "images" code is not flexible enough right now to handle a scenario without a /boot/efi .

mvo5:
Wonderful, it seems like https://github.com/osbuild/images/pull/758 has landed so the panic() I saw in the previous run is fixed.

sacsant:
> Wonderful, it seems like [osbuild/images#758](https://github.com/osbuild/images/pull/758) has landed so the panic() I saw in the previous run is fixed.

Yes, yoheiueda mentioned about this PR. Some more changes related to bootupd is also required. Working on them.

cgwalters:
> Some more changes related to bootupd is also required. Working on them.

Currently we do have bootupd enabled on Fedora CoreOS derivatives for ppc64le as far as I know; there were bootupd specific interactions you were seeing with bib here?

sacsant:
> > Some more changes related to bootupd is also required. Working on them.
> 
> Currently we do have bootupd enabled on Fedora CoreOS derivatives for ppc64le as far as I know; there were bootupd specific interactions you were seeing with bib here?

The last missing piece from bootc perspective is bootupd related code for bootloader install.
From bootc/lib/src/install.rs:1210
crate::bootloader::install_via_bootupd(&rootfs.device, &rootfs.rootfs, &state.config_opts)?;

rootfs.device passed to install_via_bootupd points to base device/disk (e.g /dev/sdb ). For ppc64le grub2-install requires PReP boot partition which is typically the first partition on device (eg /dev/sdb1) so instead of base device pass the prep-boot partition value. Something similar to https://github.com/coreos/bootupd/blob/6a4fc6f8758ff18432d5a43995b8f53ad79d4419/src/bios.rs#L42

bib/cmd/bootc-image-builder/partition_tables.go
```diff
@@ -118,4 +118,39 @@ var partitionTables = distro.BasePartitionTableMap{
 			},
 		},
 	},
+        arch.ARCH_PPC64LE.String(): disk.PartitionTable{
+		UUID: "0x14fc63d2",
+		Type: "dos",
```
cgwalters, line 123: I am pretty sure this should use GPT, right?

sacsant: Both `dos` and `gpt` are supported types. So yes, this can be changed to GPT (and the partition Type values 41 and 83 to 9E1A2D38-C612-4316-AA26-8B49521E5A8B & 0FC63DAF-8483-4772-8E79-3D69D8477DE4 respectively).

Let me know which is preferred?

cgwalters: dos is a very legacy format though, there's no reason to use it.

sacsant: Have pushed the updated code which uses gpt instead of dos format.

bib/cmd/bootc-image-builder/partition_tables.go
```diff
@@ -118,4 +118,39 @@ var partitionTables = distro.BasePartitionTableMap{
 			},
 		},
 	},
+        arch.ARCH_PPC64LE.String(): disk.PartitionTable{
+		UUID: "0x14fc63d2",
```
cgwalters, line 122: Hmm, I think that UUID format only makes sense for dos types...although, this has me wondering where the UUID referenced in the x86_64 version `D209C89E-EA5E-4FBD-B161-B461CCE297E0` came from?  Maybe we should use that?

cgwalters (approved):
Looks sane to me offhand.

Though I just now realized...you presumably copied the `dos` partition table layout from the aarch64 one...but why :scream: did someone pick dos format for aarch64? We need to follow that up separately.

(Also, again not a new problem, but this code rather desperately needs some documented `const` for some of these values...but again, can be a followup)

cgwalters:
> but why 😱 did someone pick dos format for aarch64? We need to follow that up separately.

:arrow_right: https://github.com/osbuild/bootc-image-builder/issues/551

sacsant:
> Looks sane to me offhand.
> 
> Though I just now realized...you presumably copied the `dos` partition table layout from the aarch64 one...but why 😱 did someone pick dos format for aarch64? We need to follow that up separately.
> 

I followed the code in images repository https://github.com/osbuild/images/blob/main/pkg/distro/fedora/partition_tables.go#L109

> (Also, again not a new problem, but this code rather desperately needs some documented `const` for some of these values...but again, can be a followup)

bib/cmd/bootc-image-builder/partition_tables.go
sacsant: Thanks. I have pushed an updated commit to use this UUID as well as use disk.PRePartitionGUID and other constants instead of hardcoding the Type values.

sacsant:
Rebased the branch. 
Let me know if there are any review comments on this PR.

cgwalters (approved):

cgwalters (approved):

PR #521, opened 2024-07-04
bib: when doing a `podman pull` connect stderr

mvo5:
When we run `podman pull` there is a *long* pause in the output. With that `bib` appears hanging. But `podman` prints progress on stderr so connect that give us a better user experience without polluting the `stdout` fd which needs to be pure `json` for
`bootc-image-builder manifest`.

This is a first step towards a more user friendly output from `bib`. Eventually we should probably wrap the podman progress into our own progress bar, there is an issue in podman [0] with some discussion how to do that but it requires a bunch more work.

[0] https://github.com/containers/podman-desktop/issues/152

[A different take to https://github.com/osbuild/bootc-image-builder/pull/286]


ondrejbudai (approved):
:heart:

PR #523, opened 2024-07-04
main: allow running outside of a container

mvo5:
It can be useful to run `bootc-image-builder` outside a container, e.g. during local development. Let's allow it but warn the user that it's not a supported configuration.

[my main use-case is to run the experimental progress bar as it's hard to automatically test]


ondrejbudai (approved):
I see no harm here. :)

PR #525, opened 2024-07-04
main,osbuildprogress: add `--progress=text` support (COMPOSER-2394)

mvo5:
This adds a new `bootc-image-builder` mode that makes use of the
osbuild json-seq progress information. It will display some
user friendly progress information. It can be tested via:
```sh
$ cd bib/cmd/bootc-image-builder
$ echo '{}' > config.json
$ go build && sudo ./bootc-image-builder build --progress=text --store /var/tmp/osbuild-store --config config.json quay.io/centos-bootc/centos-bootc:stream9
```
It looks like this right now:
```
Building manifest-qcow2.json
Building [|]                                                                    
[4 / 8] step: image [------------------------->_________________________] 50.00%
[7 / 11] module: bootc.install-to-filesystem [---------------->_________] 63.64%
last msg: Starting module org.osbuild.bootc.install-to-filesystem              
```

It uses `chegggaaa/pb/v3`  - but there are still open questions about the message output and error handling.

Draft because:
- error handling is very sketchy, if something in the build fails the user gets a "failed" message but no context, we need to either make sure osbuild progress has a concept of an error and dump the entire output of the stage into this (as errors details tend to be inside the failed stage (this needs some osbuild work)
- ~no syncronisation between cmd.Wait() and the progress reader reading output, this also needs to get fixed via e.g. a waitgroup so that both the cmd needs to finish and the stdout read gets an EOF before we continue as otherwise the last progress messages may get lost~
- tests (obviously) but that is tricky for progress, definitely some of the failure handling
- ~osbuild currently prints random non-json at the end of a run, we need to fix that or deal with it gracefully~
- Naming of things like pipelines/source and stages
- ~Needs https://github.com/osbuild/osbuild/pull/1826 for correct reporting~
6. ... ?

### DIFF
--- a/bib/internal/osbuildprogress/progress.go
+++ b/bib/internal/osbuildprogress/progress.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"os/exec"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/cheggaaa/pb/v3"
@@ -45,16 +46,18 @@ type OsbuildJsonProgress struct {
 	Message string `json:"message"`
 }
 
-func scanJsonSeq(r io.Reader, ch chan OsbuildJsonProgress, errCh chan error) {
-	var progress OsbuildJsonProgress
+func scanJsonSeq(wg *sync.WaitGroup, r io.Reader, ch chan OsbuildJsonProgress, errCh chan error) {
+	wg.Add(1)
+	defer wg.Done()
 
+	var progress OsbuildJsonProgress
 	scanner := bufio.NewScanner(r)
 	for scanner.Scan() {
 		// XXX: use a proper jsonseq reader?
 		line := scanner.Bytes()
 		line = bytes.Trim(line, "\x1e")
 		if err := json.Unmarshal(line, &progress); err != nil {
-			// XXX: provide an invalid lines chan
+			// XXX: provide an invalid lines chan here?
 			errCh <- err
 			continue
 		}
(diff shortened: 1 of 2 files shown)


schuellerf:
As you already mentioned in the comment, I also think that the progress (at least the object itself) should have indefinite levels to be generic.

And the progress should definitely include "Generating manifest"
So basically the progress should start right after entering the application and should not only cover some main parts. (Those should be some sub-progress IMHO)

schuellerf:
The way I think it should work is similar to a "logger" it should always exist and just be handed over (directly or indirectly/globally)

bib/internal/osbuildprogress/progress.go
```diff
@@ -0,0 +1,203 @@
...
+		Pipeline struct {
+			Name  string `json:"name"`
+			Stage struct {
+				Name string `json:"name"`
+				ID   string `json:"id"`
+			} `json:"stage"`
+			ID string `json:"id"`
+		} `json:"pipeline"`
+	} `json:"context"`
+	Progress struct {
```
mvo5, line 31: We need to add "build_result" here if/once https://github.com/osbuild/osbuild/pull/1831 is merged

mvo5:
> As you already mentioned in the comment, I also think that the progress (at least the object itself) should have indefinite levels to be generic.
> 
> And the progress should definitely include "Generating manifest" So basically the progress should start right after entering the application and should not only cover some main parts. (Those should be some sub-progress IMHO)

Thanks, yeah, that is excellent feedback. This version was a draft and mostly to explore the space. I created https://github.com/osbuild/bootc-image-builder/pull/743 and https://github.com/osbuild/images/pull/1047 now which are cleaned up versions based on this PR and address your points. I will close this one in favor of the other two.

PR #526, opened 2024-07-04
test: rework how test cases are represented/generated

mvo5:
The old way of generating test cases was via strings that would use a `,` to differenciate between various parameters like container_ref or image_type.

With the adding of the rootfs this becomes unwieldy (arguable it was unwieldy already and now it's unentable). This commit introduces a proper "TestCase" class that makes generating and reading the test cases a lot eaiser (IMHO).

Note that this was prompted because of PR#439 where we want to make the rootfs tests on fedora "btrfs".

[draft until tests are green, there is a lot of churn in here :)]

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -1,5 +1,7 @@
 import pytest
 
+from testcases import TestCase
+
 
 def pytest_addoption(parser):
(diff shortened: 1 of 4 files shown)


achilleas-k (approved):

achilleas-k:
This is really cool.  I love it!

PR #547, opened 2024-07-17
Enable new Anaconda modules: Users, Services, and Security (RHEL-49499)

achilleas-k:
**bib: enable Anaconda Users module for ISO**

We generally only enable the Users module in ISOs when a user is
configured because we want the installation to be fully automatic /
unattended and, in the general case, Anaconda will block and wait for
user input if no users are configured.  However, since we always add
`rootpw --lock` to our kickstart file, this satisfies Anaconda's user
configuration rule and the installation proceeds without blocking.

Enabling the Users module unconditionally allows users to include user
configuration when they inject their own kickstart files.

---

**bib: enable Anaconda Services and Security modules**

These modules don't affect any existing functionality but enable more
kickstart verbs for users that want to use them.

---


ondrejbudai (approved):
Thanks! :)

PR #554, opened 2024-07-19
include subscription-manager to fix integration with Konflux

ondrejbudai:
Konflux mounts in /etc/pki/entitlement instead of /run/secrets. This is not how we intended bib to work, but it works if subscription-manager is in bib. Include it temporarily, before we find a better long-term solution.

See https://github.com/konflux-ci/build-definitions/blob/f3ac40bbc0230eccb8d98a4d54dabd55a4943c5d/task/build-vm-image/0.1/build-vm-image.yaml#L198

### DIFF
--- a/package-requires.txt
+++ b/package-requires.txt
@@ -12,3 +12,9 @@ qemu-img
 
 # rpm-ostree wants these for packages
 selinux-policy-targeted distribution-gpg-keys
+
+# Konflux mounts in /etc/pki/entitlement instead of /run/secrets.
+# This is not how we intended bib to work, but it works if subscription-manager is in bib.
+# Include it temporarily, before we find a better long-term solution.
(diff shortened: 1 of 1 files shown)


cgwalters:
xref https://gitlab.com/fedora/bootc/base-images/-/merge_requests/7 where we only added subman in c9s.

(This also relates to https://gitlab.com/fedora/bootc/tracker/-/issues/2 where bib would become more distro-bound)

> Konflux mounts in /etc/pki/entitlement instead of /run/secrets. 

Hmm...I'd believe it if you said you were sure that's the problem, but I suspect the real problem is more like https://github.com/openshift/os/pull/876

ondrejbudai:
> Hmm...I'd believe it if you said you were sure that's the problem, but I suspect the real problem is more like [openshift/os#876](https://github.com/openshift/os/pull/876) - basically the container image needs the symlinks. Adding subscription-manager happens to do that.

We need `/etc/rhsm` in this case, so full `subscription-manager`.

So far it seems that this unblocked the broken pipeline.

cgwalters (approved):

PR #565, opened 2024-07-26
Use repos from container for depsolving

bcl:


bib/cmd/bootc-image-builder/image.go
```diff
@@ -57,8 +57,8 @@ type ManifestConfig struct {
 	// Extracted information about the source container image
 	SourceInfo *source.Info
 
-	// Command to run the depsolver
-	DepsolverCmd []string
+	// Path to /etc tree use for osbuild-depsolve-dnf
```
mvo5, line 60: (super nitpick) when reading this I wondered if this is the path to <target>/etc or <target>, so maybe worth tweking the comment slightly to something like: `// Path to the tree that contains the /etc/ to use for osbuild-depsolve-dnf` or (maybe too different): `// Path to the target container used by osbuild-depsolve-dnf for depsolving` or something.

bib/internal/container/container.go
```diff
@@ -83,9 +83,14 @@ func (c *Container) Stop() error {
 	if output, err := exec.Command("podman", "stop", c.id).CombinedOutput(); err != nil {
 		return fmt.Errorf("stopping %s container failed: %w\noutput:\n%s", c.id, err, output)
 	}
-	// when the container is stopped by podman it will not honor the "--rm"
-	// that was passed in `New()` so manually remove the container here.
+	// when the container is stopped by podman it may not honor the "--rm"
+	// that was passed in `New()` so manually remove the container here if it is still available
```
mvo5, line 87: Thanks for finding and fixing this! I wonder if it could be it's own PR though? Partly because it's orthogonal to the PR title, partly because I would love to poke at this a bit more and see if there is a way that avoids reading the error message (I don't think podman does i18n for errors but it might some day). But maybe I'm overthinking this. Alternatively we can of course keep as is and just do followups.

mvo5 (commented):
I really like this approach! Some small suggestions/questions inline though.

bib/internal/container/container.go
```diff
@@ -83,9 +83,14 @@ func (c *Container) Stop() error {
...
 	if output, err := exec.Command("podman", "rm", c.id).CombinedOutput(); err != nil {
+		// This is ok, container has already been removed
+		if strings.Contains(string(output), "no container with ID or name") {
+			return nil
+		}
```
achilleas-k, line 92: I think we can use `--ignore` here.

From podman-rm(1)
```
   --ignore, -i
       Ignore errors when specified containers are not in the container store.  A user might have decided to manually remove a container which leads to a failure during the ExecStop directive of a systemd service referencing that container.
```

bib/internal/container/container.go
bcl: I'd rather keep it as part of this since it lets the tests pass, at least locally for me on Fedora 39.

bib/internal/container/container.go
bcl: Looks like that works. Tests are failing for different reasons now, looks like commit af8321c0ed307ec08c360d06b7e8c9e9c42ac3f2 needs to adjust some other things.

mvo5 (approved):
Thank you

bib/internal/container/container.go
mvo5: @bcl Thanks for raising this - I opened https://github.com/osbuild/bootc-image-builder/pull/570 to fix this

achilleas-k (approved):

PR #567, opened 2024-07-29
test: extract `testutil.podman_run_common`

mvo5:
This consolidates the duplication from `test_{build,manifest}` when running podman. All common arguments are now in a single place which also means that the host container storage is mapped consistently which should speed up the tests.


cgwalters (approved):
Nice!

achilleas-k (approved):

PR #568, opened 2024-07-29
bib: drop hardcoding disk UUID

mvo5:
The current parition_tables.go hardcodes the UUID for the PartitionTable. This is undesired. Drop it so that images generates it automatically and randomly for us.

A similar change is probably wanted for images itself and the parition UUIDs.

[draft as this needs discussion and at lest one test

Also volumeid for vfat needs fixing


mvo5:
Setting to draft for now for the same reasons as https://github.com/osbuild/images/pull/823

PR #573, opened 2024-07-31
bib: support multiple distro versions in a distrodef

mvo5:
bib: support multiple distro versions in a distrodef

This commit adds support to support multiple versions of a distro in
a distrodef. This is required so that centos-10 can used as it has a
different packageset compared to centos-9.

----

defs: add centos-10/rhel-10

The package set for {centos,rhel}-10 needs tweaking, x11 got
dropped.

Fixes: [RHEL-50741](https://issues.redhat.com/browse/RHEL-50741)

---

Fwiw, I was thinking about an alternative approach that would fallback to the closest version number if a distrodef is missing. But it seems like a bit overkill and this is a reasonable starting point. Also otk will hopefully make sharing this easier.


ondrejbudai:
Looks good, but I think we should implement the version fallback in this PR, I'm pretty sure someone out there is already using bib with Fedora rawhide, and thus `VERSION_ID` == `41`.

mvo5:
> Looks good, but I think we should implement the version fallback in this PR, I'm pretty sure someone out there is already using bib with Fedora rawhide, and thus `VERSION_ID` == `41`.

A simple fix would be a symlink but I'm happy to also do the fallback.

mvo5:
I added a symlink to not regression on rawhide and start working on the fallback now.

ondrejbudai (approved):
Yeah, that works too. :)

achilleas-k (approved):

PR #574, opened 2024-07-31
distrodef: add fallback code when extact distro ver is not  found

mvo5:
This adds a fallback when no extact match for a distrodef is found.
The rules are:
- exact match (e.g. fedora-40) wins
- highest version for the given distro wins otherwise, e.g.
  fedora-42 is returned when asking for fedora-44 and [fedora-39 fedora-42]
  are available
- it is an error if the requested version is smaller than the availalble
  version and there is no exact match. E.g.
  an error is raised if asking for fedora-38 and [fedora-39 fedora-42]
  are available

Build on top of https://github.com/osbuild/bootc-image-builder/pull/573

### DIFF
--- a/bib/data/defs/fedora-41.yaml
+++ b/bib/data/defs/fedora-41.yaml
@@ -1,1 +1,0 @@
-fedora-40.yaml
--- a/bib/internal/distrodef/distrodef.go
+++ b/bib/internal/distrodef/distrodef.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"golang.org/x/exp/maps"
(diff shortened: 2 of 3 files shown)


bib/data/defs/centos-10.yaml
```diff
@@ -0,0 +1,94 @@
+anaconda-iso:
```
cgwalters, line 1: I know this is not a new thing but I will just reiterate https://github.com/osbuild/bootc-image-builder/pull/58#discussion_r1443292447

cgwalters (commented):
Not an expert in this code, but FTR looks sane to me!

bib/data/defs/centos-10.yaml
mvo5: Uh, sorry, yeah, I really need to look into your suggestion here :(

achilleas-k (approved):

ondrejbudai (approved):
Splendid, thanks! :heart:

PR #581, opened 2024-08-01
Skip rootfs handling when building ISOs

### LINKED ISSUE #494: Manifest generation for ISO should ignore missing rootfs type
https://github.com/osbuild/bootc-image-builder/blob/61863ff3c39da79716b0a94945b153e58a138ca6/bib/cmd/bootc-image-builder/main.go#L244-L252

When generating a manifest for an ISO, if the base container doesn't specify a rootfs and it's also not set on the command line, the generation fails.  It probably shouldn't because we don't do anything with it when building an ISO.

Came across this while doing some ISO tests for #438.  I also realised we don't have a quick smoke test for ISO manifest generation (we only do disks in `test_manifest_smoke()`, so I'll add that too.

achilleas-k:
The ISO build doesn't require the rootfs configuration from the image,
so there is no need to read it from the container nor to require the
flag when it is missing.

Closes #494.

### DIFF
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -262,27 +262,29 @@ func manifestFromCobra(cmd *cobra.Command, args []string) ([]byte, *mTLSConfig, 
 	}()
 
 	var rootfsType string
-	if rootFs != "" {
-		rootfsType = rootFs
-	} else {
-		rootfsType, err = container.DefaultRootfsType()
-		if err != nil {
-			return nil, nil, fmt.Errorf("cannot get rootfs type for container: %w", err)
+	if buildType != BuildTypeISO {
+		if rootFs != "" {
+			rootfsType = rootFs
+		} else {
+			rootfsType, err = container.DefaultRootfsType()
+			if err != nil {
+				return nil, nil, fmt.Errorf("cannot get rootfs type for container: %w", err)
+			}
+			if rootfsType == "" {
+				return nil, nil, fmt.Errorf(`no default root filesystem type specified in container, please use "--rootfs" to set manually`)
+			}
 		}
-		if rootfsType == "" {
-			return nil, nil, fmt.Errorf(`no default root filesystem type specified in container, please use "--rootfs" to set manually`)
-		}
-	}
 
-	// TODO: on a cross arch build we need to be conservative,
-	// i.e.  we can only use the default ext4 because if xfs is
-	// select we run into the issue that mkfs.xfs calls
-	// "ioctl(BLKBSZSET)" which is missing in qemu-user, once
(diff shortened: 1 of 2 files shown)


cgwalters (approved):

bib/cmd/bootc-image-builder/main.go
```diff
@@ -262,27 +262,29 @@ func manifestFromCobra(cmd *cobra.Command, args []string) ([]byte, *mTLSConfig,
...
-	if rootFs != "" {
-		rootfsType = rootFs
-	} else {
-		rootfsType, err = container.DefaultRootfsType()
-		if err != nil {
-			return nil, nil, fmt.Errorf("cannot get rootfs type for container: %w", err)
-		}
-		if rootfsType == "" {
-			return nil, nil, fmt.Errorf(`no default root filesystem type specified in container, please use "--rootfs" to set manually`)
+	if buildType != BuildTypeISO {
```
mvo5, line 265: (super nitpick) this is already quite nested, maybe worthwhile to extract into a helper because then much of the nesting can just become early returns? something like:
```go
func getRootfsType(container *podman_container.Container, cntArch arch.Arch, buildType BuildType, userProvidedRootfs string) (string, error) {
	if buildType == BuildTypeISO {
		return "", nil
	}
	if userProvidedRootfs != "" {
		return userProvidedRootfs, nil
	}
	rootfsType, err := container.DefaultRootfsType()
	if err != nil {
		return "", fmt.Errorf("cannot get rootfs type for container: %w", err)
	}
	if rootfsType == "" {
		return "", fmt.Errorf(`no default root filesystem type specified in container, please use "--rootfs" to set manually`)
	}

	// TODO: on a cross arch build we need to be conservative,
	// i.e.  we can only use the default ext4 because if xfs is
	// select we run into the issue that mkfs.xfs calls
	// "ioctl(BLKBSZSET)" which is missing in qemu-user, once
	// https://www.mail-archive.com/qemu-devel@nongnu.org/msg1037409.html
	// is merged we can remove the following code
	if cntArch != arch.Current() && rootfsType != "ext4" {
		logrus.Warningf("container prefered root filesystem %q cannot be used during cross arch build", rootfsType)
		return "", nil
	}

	return rootfsType, nil
}
```
diff: [pr581.diff.txt](https://github.com/user-attachments/files/16468640/pr581.diff.txt)

But the wins are not that huge as it takes a bit too many random parameters for my taste. So a bit up to you :)

achilleas-k: We could do some cleanup later, sure.

mvo5: I can take care of this, sorry that I was so late to his

Issue #551, opened 2024-07-18
aarch64 uses dos format partitions


cgwalters:
https://github.com/osbuild/bootc-image-builder/blame/babb61b0112a7b55b4d1137f200074a86ff25f61/bib/cmd/bootc-image-builder/partition_tables.go#L79

Why?? git annotate points to https://github.com/osbuild/bootc-image-builder/commit/344a1c4d0aebe6502f6a56faf188839cc0151184 cc @achilleas-k

cgwalters:
While we're here, I think we should fix ppc64le and s390x in osbuild to use gpt.

For reference it looks like the coreos pipeline is doing this, e.g. https://github.com/coreos/coreos-assembler/blob/aea950cfadfae9eae9b92c6bb22bfaafe84c092c/src/osbuild-manifests/coreos.osbuild.s390x.mpp.yaml#L59

cgwalters:
@yoheiueda (in case you know)...is there any reason not to use gpt for s390x? AFAICS it seems today Anaconda in RHEL9.4 defaults to dos...which, just skimming the code looks like it's probably this https://github.com/rhinstaller/anaconda/blob/19df67bad509b4cb65ff23c24a7d54ee586b5e68/pyanaconda/modules/storage/initialization.py#L95 just because anaconda treats zipl as "mbr"?

And I'm guessing that the osbuild Go reimplementation of this just inherited that logic. But ISTM that we should really change Anaconda too to just default to GPT unconditionally nowadays. I guess there's maybe a vague concern about old x86_64 BIOS systems, but that's clearly not relevant for s390x.

(Random aside, searching for "anaconda gpt" is now dominated by AI-related search results because "gpt" acquired a much more popular expansion in recent years, and "anaconda" was previously also a separate python installer...)

yoheiueda:
@cgwalters zipl supports GPT, so I can't come up with any ideas to use dos partition table format.

achilleas-k:
> https://github.com/osbuild/bootc-image-builder/blame/babb61b0112a7b55b4d1137f200074a86ff25f61/bib/cmd/bootc-image-builder/partition_tables.go#L79
> 
> Why?? git annotate points to [344a1c4](https://github.com/osbuild/bootc-image-builder/commit/344a1c4d0aebe6502f6a56faf188839cc0151184) cc @achilleas-k

We should switch to gpt wherever possible, I agree.  For some history about this:

[...]

PR #582, opened 2024-08-01
partitions: Switch aarch64 to GPT, fully deduplicate

cgwalters:
Closes: https://github.com/osbuild/bootc-image-builder/issues/551

There was a fair amount of copy-paste even when we just had aarch64+x86_64, the addition of duplicate bootfs/rootfs definitions for ppc64le+s390x is a strong motivation to finally define shared constants.

I also factored out a shared `const` with some comments for the disk UUID that will save someone some time wondering about it later (this textually but not logically conflicts with https://github.com/osbuild/bootc-image-builder/pull/568 but I think this makes sense to land first).

### DIFF
--- a/bib/cmd/bootc-image-builder/partition_tables.go
+++ b/bib/cmd/bootc-image-builder/partition_tables.go
@@ -21,9 +21,69 @@ const (
 	RootOptions = "ro"
 )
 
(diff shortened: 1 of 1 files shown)


mvo5 (approved):
Thank you! This is really nice

PR #597, opened 2024-08-07
docs: Authenticate to private repo using auth.json mount

No9:
Not strictly a bootc-image-builder volume but I couldn't see in the docs how to use a private repo so came up with this approach and thought I'd capture it in case others need it.


No9:
This is incorrect - The docs here outline how credentials should be handled. 
https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/using_image_mode_for_rhel_to_build_deploy_and_manage_operating_systems/managing-users-groups-ssh-key-and-secrets-in-image-mode-for-rhel#injecting-secrets-to-a-image-mode-for-rhel_managing-users-groups-ssh-key-and-secrets-in-image-mode-for-rhel

Issue #579, opened 2024-08-01
Support custom mountpoints


achilleas-k:
Opening this issue to track support for custom mountpoints.

@mvo5 described the issue with custom mountpoints in 06e1b2a67abea54425ea4d36cb64a5d2d988af1e.

Short version:
`bootc` needs an empty root tree to install to when running `bootc install to-filesystem`.  With our current pipelines, when we build an image, we format the disk with all the partitions, mount every mountpoint to its location under a root tree, and call `bootc install` to put its files in the fully mounted root tree, which will be non-empty if it contains directories for custom mountpoints.

I tested the idea in the commit message
> After "install-to-filesystem" ran we need a "org.osbuild.mkdir" stage for the extra mount points that also only mounts the "essential" mounts.
and it works as expected, with some caveats:

1. Custom mountpoints cannot be created on `/` (the deployed root, not the physical disk root), since after `bootc install`, it is marked immutable.
2. Custom mountpoints cannot contain any data on first boot.
3. If a custom mountpoint is configured for a path that exists in the base image and is non-empty, the mountpoint will cover the data.

For example, I'm considering the following scenario:
```Dockerfile
FROM quay.io/centos-bootc/centos-bootc:stream9

RUN mkdir -p /opt/myapplication/log
RUN date > /opt/myapplication/log/build-time
```
```toml
[[customizations.filesystem]]
mountpoint = "/opt/myapplication/log"
minsize = "20 GiB"
```

With our proposed solution, bootc will create a filesystem that contains `/opt/myapplication/log/build-time`, but on boot the path `/opt/myapplication/log` will be shadowed by the new mountpoint.  If there's no way to support this scenario, we should probably inspect the image (which we already do when preparing the manifest) and error out.

Questions (cc @cgwalters):
- I'm told we have users that would like to create mountpoints on `/`.  Can we do this?  If we flip the immutable flag, create a directory for the mountpoint under `/`, and flip it back on, will that be enough or will there be unwanted side effects?
- Do we want to support a scenario where a base image contains data destined for a custom mountpoint?  I'm thinking of scenarios where a base container image is created with some data under, say, `/data`, and the disk image is meant to have a separate partition for it.  Is this even possible?

cgwalters:
Aside but an important one for me: It'd be really, really, really, really nice if we can share code with Anaconda too. All of these things need to be supported there too.

---

[...]

PR #601, opened 2024-08-09
Support custom mountpoints under /var (COMPOSER-2301)

achilleas-k:
This PR adds initial support for specifying custom mountpoints while creating a disk image.  We define a new mountpoint policy internal to BIB, separate and more restrictive than the ostree mountpoint policies in osbuild/images.  The rules are:
1. `/` and `/boot` are supported for modifying their size.
2. Custom mountpoints under `/var` that do not become symlinks to other paths are supported.
3. `/var` itself is not.

The combination of 2. and 3. isn't directly supported in our policy checkers, so instead the policy in this PR initially allows for `/var` and its subdirectories, but then makes an explicit check that `/var` itself isn't in the custom filesystems in the blueprint.

Btrfs subvolumes are not supported because they are mounted with the parent disk at the location where they are created, so it is not possible to run `bootc install to-filesystem` with an empty root fileystem (see also https://github.com/osbuild/images/pull/846).  Since mountpoints cannot be created at paths that will contain data on or before first boot, a workaround for this would be to create subvolumes on first boot (manually or otherwise).  We could enable this with firstboot scripts if necessary.

The discussion in #579 should provide further context for the limitations.


bib/cmd/bootc-image-builder/image.go
```diff
@@ -77,20 +75,74 @@ func Manifest(c *ManifestConfig) (*manifest.Manifest, error) {
...
+	// ostree mountpoint policy defined in osbuild/images. It only allows /
+	// (for sizing the root partition) and custom mountpoints under /var but
+	// not /var itself.
+
+	// Since our policy library doesn't support denying a path while allowing
+	// its subpaths (only the opposite), we augment the standard policy check
+	// with a simple search through the custom mountpoints to deny /var
+	// specifically.
+	mountpointPolicy = pathpolicy.NewPathPolicies(map[string]pathpolicy.PathPolicy{
+		// allow all existing mountpoints to support size customizations
```
mvo5, line 89: Maybe: `// allow all existing mountpoints (but no subdirs) to support size customizations`

bib/cmd/bootc-image-builder/image.go
```diff
@@ -77,20 +75,74 @@ func Manifest(c *ManifestConfig) (*manifest.Manifest, error) {
...
+		"/":     {Deny: false, Exact: true},
+		"/boot": {Deny: false, Exact: true},
+
+		// /var is not allowed, but we need to allow any subdirectories that
+		// are not denied below, so we allow it initially and then check it
+		// separately
```
mvo5, line 95: maybe `// separately in the code below`

bib/cmd/bootc-image-builder/image.go
```diff
@@ -77,20 +75,74 @@ func Manifest(c *ManifestConfig) (*manifest.Manifest, error) {
...
+		"/var/mail":     {Deny: true}, // symlink to spool/mail
+		"/var/mnt":      {Deny: true},
+		"/var/roothome": {Deny: true},
+		"/var/run":      {Deny: true}, // symlink to ../run which is on tmpfs
+		"/var/srv":      {Deny: true},
+		"/var/usrlocal": {Deny: true},
+	})
+
+	mountpointMinimalPolicy = pathpolicy.NewPathPolicies(map[string]pathpolicy.PathPolicy{
+		// allow all existing mountpoints to support size customizations
```
mvo5, line 110: maybe `// allow all existing mountpoints (but no subdirs) to support size customizations`

bib/cmd/bootc-image-builder/image.go
```diff
@@ -77,20 +75,74 @@ func Manifest(c *ManifestConfig) (*manifest.Manifest, error) {
...
+		"/":     {Deny: false, Exact: true},
+		"/boot": {Deny: false, Exact: true},
+	})
+)
+
+func checkMountpoints(filesystems []blueprint.FilesystemCustomization, policy *pathpolicy.PathTrie) error {
+	invalid := make([]string, 0)
+	for _, fs := range filesystems {
+		if err := policy.Check(fs.Mountpoint); err != nil {
+			invalid = append(invalid, fs.Mountpoint)
```
mvo5, line 120: It seems this check is discarding some information we get from the policy.Check() error that might be important to users and could lead to slightly misleading situations. Consider a customization that has `/boot/` instead of `/boot`. This will result in the following error (to the user): `The following custom mountpoints are not supported ["/boot/"]`. However the error from the underlying error from `Check()` is: `path must be canonical`.

Maybe instead (and tweak once https://github.com/osbuild/images/pull/853 is merged)?
```go
diff --git a/bib/cmd/bootc-image-builder/image.go b/bib/cmd/bootc-image-builder/image.go
index cc8be76..63a31f1 100644
--- a/bib/cmd/bootc-image-builder/image.go
+++ b/bib/cmd/bootc-image-builder/image.go
@@ -2,6 +2,7 @@ package main
 
 import (
        cryptorand "crypto/rand"
+       "errors"
        "fmt"
        "math"
        "math/big"
@@ -114,17 +115,17 @@ var (
 )
 
 func checkMountpoints(filesystems []blueprint.FilesystemCustomization, policy *pathpolicy.PathTrie) error {
-       invalid := make([]string, 0)
+       var errs []error
        for _, fs := range filesystems {
                if err := policy.Check(fs.Mountpoint); err != nil {
-                       invalid = append(invalid, fs.Mountpoint)
+                       errs = append(errs, err)
                }
                if fs.Mountpoint == "/var" {
-                       invalid = append(invalid, "/var")
+                       errs = append(errs, fmt.Errorf("path '/var' is not allowed"))
                }
        }
-       if len(invalid) > 0 {
-               return fmt.Errorf("The following custom mountpoints are not supported %+q", invalid)
+       if len(errs) > 0 {
+               return fmt.Errorf("The following errors occurred while setting up custom mountpoints:\n%w", errors.Join(errs...))
        }
        return nil
 }
diff --git a/bib/cmd/bootc-image-builder/image_test.go b/bib/cmd/bootc-image-builder/image_test.go
index 6dace68..6ff74d9 100644
--- a/bib/cmd/bootc-image-builder/image_test.go
+++ b/bib/cmd/bootc-image-builder/image_test.go
@@ -105,7 +105,7 @@ func TestCheckFilesystemCustomizationsValidates(t *testing.T) {
                                {Mountpoint: "/ostree"},
                        },
                        ptmode:      disk.RawPartitioningMode,
-                       expectedErr: `The following custom mountpoints are not supported ["/ostree"]`,
+                       expectedErr: "The following errors occurred while setting up custom mountpoints:\npath '/ostree ' is not allowed",
                },
                {
                        fsCust: []blueprint.FilesystemCustomization{
@@ -113,7 +113,7 @@ func TestCheckFilesystemCustomizationsValidates(t *testing.T) {
                                {Mountpoint: "/var"},
                        },
                        ptmode:      disk.RawPartitioningMode,
-                       expectedErr: `The following custom mountpoints are not supported ["/var"]`,
+                       expectedErr: "The following errors occurred while setting up custom mountpoints:\npath '/var' is not allowed",
                },
                {
                        fsCust: []blueprint.FilesystemCustomization{
@@ -121,7 +121,14 @@ func TestCheckFilesystemCustomizationsValidates(t *testing.T) {
                                {Mountpoint: "/var/data"},
                        },
                        ptmode:      disk.BtrfsPartitioningMode,
-                       expectedErr: `The following custom mountpoints are not supported ["/var/data"]`,
+                       expectedErr: "The following errors occurred while setting up custom mountpoints:\npath '/var/data ' is not allowed",
+               },
+               {
+                       fsCust: []blueprint.FilesystemCustomization{
+                               {Mountpoint: "/"},
+                               {Mountpoint: "/boot/"},
+                       },
+                       expectedErr: "The following errors occurred while setting up custom mountpoints:\npath must be canonical",
                },
        } {
                cust := &blueprint.Customizations{
```

```

```

test/test_build.py
```diff
@@ -359,14 +355,37 @@ def test_image_boots(image_type):
...
+        for line in output.splitlines()[1:]:
             fields = line.split()
-            if fields[0] == "/sysroot":
-                size = int(fields[1])
-                assert size > 11 * 1024 * 1024
-                break
+            # Note that df output is in 1k blocks, not bytes
+            mountpoint_sizes[fields[0]] = int(fields[1]) * 2 ** 10  # in bytes
+
+        fs_customizations = testutil.create_filesystem_customizations(image_type.rootfs)
```
mvo5, line 369: (nitpick) maybe worth extracting this into a small helper like: `assert_fs_customizations(image_type, mountpoint_sizes) ` to keep the `test_image_boots` function slightly smaller?

mvo5 (dismissed):
Thank you! This looks nice and will be a super nice feature for our users. I have some small suggestion inline mostly around the error handling/comments. It's fine to merge as is of course as my suggestions are all cosmetic.

test/test_build.py
achilleas-k: Done!

bib/cmd/bootc-image-builder/image.go
achilleas-k: Thanks.  You're absolutely right of course.

achilleas-k:
The error messages we test for will change once https://github.com/osbuild/images/pull/853 is merged and pulled in here, but we can deal with that separately if this gets merged first.

achilleas-k:
I think everything's sorted now.

cgwalters:
IOW it'd be helpful I think if the PR included changes to the docs in this repo at least.

achilleas-k:
I've updated the README with information on the new customizations.

bib/cmd/bootc-image-builder/image.go
```diff
@@ -77,21 +77,113 @@ func Manifest(c *ManifestConfig) (*manifest.Manifest, error) {
...
-			}
+	return nil
+}
+
+// updateFilesystemSizes updates the size of the root filesystem customization
+// based on the minRootSize. The new min size whichever is larger between the
+// existing size and the minRootSize. If the root filesystem is not already
+// configured, a new customization is added.
+func updateFilesystemSizes(fsCustomizations []blueprint.FilesystemCustomization, minRootSize uint64) []blueprint.FilesystemCustomization {
+	updated := make([]blueprint.FilesystemCustomization, len(fsCustomizations), len(fsCustomizations)+1)
```
mvo5, line 158: (nitpick) Sorry, this maybe a silly question, but do we need to make a copy here? Or could we just modify the slice in-place (quick test below). But fine of course, if you prefer it this way (mostly curious :)
```go
diff --git a/bib/cmd/bootc-image-builder/image.go b/bib/cmd/bootc-image-builder/image.go
index 6c2c89f..1247ab8 100644
--- a/bib/cmd/bootc-image-builder/image.go
+++ b/bib/cmd/bootc-image-builder/image.go
@@ -155,21 +155,19 @@ func checkFilesystemCustomizations(fsCustomizations []blueprint.FilesystemCustom
 // existing size and the minRootSize. If the root filesystem is not already
 // configured, a new customization is added.
 func updateFilesystemSizes(fsCustomizations []blueprint.FilesystemCustomization, minRootSize uint64) []blueprint.FilesystemCustomization {
-       updated := make([]blueprint.FilesystemCustomization, len(fsCustomizations), len(fsCustomizations)+1)
        hasRoot := false
-       for idx, fsc := range fsCustomizations {
-               updated[idx] = fsc
-               if updated[idx].Mountpoint == "/" {
-                       updated[idx].MinSize = max(updated[idx].MinSize, minRootSize)
+       for idx := range fsCustomizations {
+               if fsCustomizations[idx].Mountpoint == "/" {
+                       fsCustomizations[idx].MinSize = max(fsCustomizations[idx].MinSize, minRootSize)
                        hasRoot = true
                }
        }
 
        if !hasRoot {
                // no root customization found: add it
-               updated = append(updated, blueprint.FilesystemCustomization{Mountpoint: "/", MinSize: minRootSize})
+               fsCustomizations = append(fsCustomizations, blueprint.FilesystemCustomization{Mountpoint: "/", MinSize: minRootSize})
        }
-       return updated
+       return fsCustomizations
 }
 
 // setRootfsType sets the filesystem type for the mountable entity with target

```

bib/cmd/bootc-image-builder/image.go
```diff
@@ -77,21 +77,113 @@ func Manifest(c *ManifestConfig) (*manifest.Manifest, error) {
...
+// setRootfsType sets the filesystem type for the mountable entity with target
+// '/'.
+func setRootfsType(pt *disk.PartitionTable, rootfs string) error {
+	rootMountable := pt.FindMountable("/")
+	if rootMountable == nil {
+		return fmt.Errorf("internal partition table does not define a root filesystem")
+	}
+	rootFsEntity, isfs := rootMountable.(*disk.Filesystem)
+	if !isfs {
+		return fmt.Errorf("the root mountable disk entity of the base partition table is not an ordinary filesystem")
```
mvo5, line 184: Maybe worth including what it is instead for easier debug, something like:
```go
@@ -181,7 +179,7 @@ func setRootfsType(pt *disk.PartitionTable, rootfs string) error {
        }
        rootFsEntity, isfs := rootMountable.(*disk.Filesystem)
        if !isfs {
-               return fmt.Errorf("the root mountable disk entity of the base partition table is not an ordinary filesystem")
+               return fmt.Errorf("the root mountable disk entity of the base partition table is not an ordinary filesystem but %T", rootFsEntity)
        }
        rootFsEntity.Type = rootfs
        return nil
```

bib/cmd/bootc-image-builder/image.go
```diff
@@ -77,21 +77,113 @@ func Manifest(c *ManifestConfig) (*manifest.Manifest, error) {
...
+	}
+	return updated
+}
+
+// setRootfsType sets the filesystem type for the mountable entity with target
+// '/'.
+func setRootfsType(pt *disk.PartitionTable, rootfs string) error {
+	rootMountable := pt.FindMountable("/")
+	if rootMountable == nil {
+		return fmt.Errorf("internal partition table does not define a root filesystem")
```
mvo5, line 180: *maybe* (I'm really not sure) we should include the incorrect partition table here as well via `%+v` or something to give us a maximum of debug output for this internal error. It may make the error very ugly but maybe it's worth it because the error is supposed to never happen :)

mvo5 (commented):
Thanks for the updated version and the excellent README update. It seems some unit test is failing so I won't appove just yet but it looks very close. Some small suggestions/questions inline for your consideration.

achilleas-k:
> It seems some unit test is failing so I won't appove just yet but it looks very close.

Yeah, I spotted those too but was waiting for the full result to address them (running locally was taking too long last night and I had to stop).  I'll get to them now.

bib/cmd/bootc-image-builder/image.go
achilleas-k: Heh, it's a tricky decision isn't it.  Yeah, this "shouldn't happen", but if it does somehow, do we need the full output or do we want it to be readable?  I guess making sure we have all the information in this rare (impossible) event is better.

Ok, you convinced me.  I'll include it.

bib/cmd/bootc-image-builder/image.go
achilleas-k: I did do it like this at first, but I felt it was simpler to keep the function pure.  There's no real need to keep `customizations.Filesystem` intact currently, but I can see how modifying the user's input can get messy if for some reason we end up needing to read it after this function is called.  There's the extra issue of it not being entirely consistent.  Sometimes, what you get back will be the same object that's attached to the `customizations` and sometimes it wont (depending on whether `append()` is called or not).  I'd prefer to avoid the scenario where we modify one of the arrays after the fact and affect the other but it only happens in some scenarios and becomes a hard to trace bug.

mvo5: Thanks for the rational!

achilleas-k: I hope it makes sense :)

README.md
```diff
@@ -343,6 +343,59 @@ Example:
...
+
+#### Interaction with `rootfs`
+
+
+#### Allowed mountpoints and sizes
+
+The following restrictions and rules apply, unless the rootfs is `btrfs`:
+- `/` can be specified to set the desired (minimum) size of the root filesystem. The final size of the filesystem, mounted at `/sysroot` on a booted system, is the value specified in this configuration or 2x the size of the base container, whichever is largest.
+- `/boot`can be specified to set the desired size of the boot partition.
+- Any subdirectory under `/var` can be specified except:
```
cgwalters, line 385: How about:

> Subdirectories of `/var` are supported, but symlinks in `/var` are not. For example, `/var/home` and `/var/run` are symlinks and cannot be filesystems on their own.

achilleas-k: Done.

bib/cmd/bootc-image-builder/image.go
achilleas-k: Done.

bib/cmd/bootc-image-builder/image.go
achilleas-k: Done.

achilleas-k:
> It seems some unit test is failing so I won't appove just yet but it looks very close.

The issue was the fallback logic for the root filesystem type in cross-arch builds.  I made some changes to the internal logic there, making it an error if it's not set from outside the `Manifest()` call.  I think I prefer the explicitness.

mvo5 (approved):
Thank you

cgwalters:
I think this change stands alone and makes sense on its own; but we did also discuss supporting fully custom toplevel mount points like `/data` or whatever - is that a followup?

There's also the case for e.g. `/mnt` (a default empty toplevel directory)

achilleas-k:
> I think this change stands alone and makes sense on its own; but we did also discuss supporting fully custom toplevel mount points like `/data` or whatever - is that a followup?

Yes, after our discussion in #579, I decided the quickest way to get something working is this.  Top-level mounts will need to be handled differently (`.mount` units?), so once we figure all that out, we can have it work the same way for everything.

achilleas-k:
Needs a bit more work for fs type handling and cross-arch builds.

achilleas-k:
The new commits (last 3) add tests as well as a behaviour change: All filesystems now inherit the type from the rootfs.  This includes `/boot`, which so far was always `ext4` even when `/` was `xfs`.

The new commits also fix the cross-arch scenario by "forcing" `ext4` regardless of the selected rootfs type.  The commit messages have more details.

bib/cmd/bootc-image-builder/image.go
```diff
@@ -162,30 +270,24 @@ func manifestForDiskImage(c *ManifestConfig, rng *rand.Rand) (*manifest.Manifest
 	if !ok {
 		return nil, fmt.Errorf("pipelines: no partition tables defined for %s", c.Architecture)
 	}
-	// XXX: extract into helper
+
```
mvo5, line 273: Maybe still worth extracting the partition generation into a helper? This would also help with testing via image_test.go as the whole partition handling could be tested directly from go in a "integrationish" fashion. 

Attached is a sketched out idea that extracts the helper and does a single test to showcase the idea. The test of course would need to be transformed into a table test but I think it's a nice way to ensure that partition tests are quick

[0001-image-extract-genPartitionTable-helper-and-add-unit-.patch.txt](https://github.com/user-attachments/files/16610050/0001-image-extract-genPartitionTable-helper-and-add-unit-.patch.txt)

test/test_manifest.py
```diff
@@ -344,3 +345,100 @@ def test_manifest_target_arch_smoke(build_container, tc):
...
+
+
+@pytest.mark.parametrize("fscustomizations,rootfs", [
+    ({"/var/data": "2 GiB", "/var/stuff": "10 GiB"}, "xfs"),
+    ({"/var/data": "2 GiB", "/var/stuff": "10 GiB"}, "ext4"),
+    ({"/": "2 GiB", "/boot": "1 GiB"}, "ext4"),
+    ({"/": "2 GiB", "/boot": "1 GiB", "/var/data": "42 GiB"}, "ext4"),
+    ({"/": "2 GiB"}, "btrfs"),
+])
+def test_manifest_fs_customizations(tmp_path, build_container, fscustomizations, rootfs):
```
mvo5, line 368: I wonder if we can cover many of these cases in unit tests maybe? I think it's important to cover at least one integration test here but testing many cases at this level is relatively costly (the test went from 2min to 7min on my system) so maybe move some into go-unit tests and just cover one scenario (or two?) here.

test/test_manifest.py
```diff
@@ -344,3 +345,100 @@ def test_manifest_target_arch_smoke(build_container, tc):
...
+    filesystems = fstab_stage["options"]["filesystems"]
+
+    manifest_mountpoints = set()
+    for fs in filesystems:
+        manifest_mountpoints.add(fs["path"])
+        if fs["path"] == "/boot/efi":
+            assert fs["vfs_type"] == "vfat"
+            continue
+
+        if fstype == "btrfs" and fs["path"] == "/boot":
```
mvo5, line 402: (nitpick) this might be another small benefit if we move more of this testing into a go unit test. This helper here is now relatively clever and encodes knowledge. A go unit test would just be "inputs", "expected_outputs" which I find slightly easier to read and reason about.

test/test_manifest.py
```diff
@@ -344,3 +345,100 @@ def test_manifest_target_arch_smoke(build_container, tc):
...
+
+@pytest.mark.skipif(platform.uname().machine != "x86_64", reason="cross arch test only runs on x86")
+@pytest.mark.parametrize("fscustomizations,rootfs", [
+    ({"/var/data": "2 GiB", "/var/stuff": "10 GiB"}, "xfs"),
+    ({"/var/data": "2 GiB", "/var/stuff": "10 GiB"}, "ext4"),
+    ({"/": "2 GiB", "/boot": "1 GiB"}, "ext4"),
+    ({"/": "2 GiB", "/boot": "1 GiB", "/var/data": "42 GiB"}, "ext4"),
+    ({"/": "2 GiB"}, "btrfs"),
+])
+def test_manifest_fs_customizations_xarch(tmp_path, build_container, fscustomizations, rootfs):
```
mvo5, line 422: (parts of) this could maybe also become go unit tests - I guess we need to have at least the case here that ended up with an empty rootfs though.

mvo5 (dismissed):
Thank you, this still looks nice. A few suggestion inline most mostly about testing/shuffling code around so could be followups (or ignored if you don't like the ideas :)

test/test_manifest.py
achilleas-k: I wanted an integration test to run the paths starting from the `rootfs` command line option all the way down, and having a real container in the mix with its own rootfs type config should be overridden.  Code paths under that should be mostly covered by unit tests already, but this will catch regressions or changes with how the rootfs choice interacts with the customizations, which surfaces issues like the rootfs being empty when creating cross-arch manifests.

I think a lot of this can be sped up if we had a way to mock the base container, but that's rather difficult with how much we depend on information from the base container read through actually running or mounting it.  If we could do that, our manifest integration tests would take a few seconds.

mvo5: Thank you! Totally fine to do it this way, just wanted to have the conversation :) I will mark this as resolved.

achilleas-k: FTR, I compared integration test timing between this branch and `main` (for manifest tests).  Running
```
pytest -k test_manifest
```
I get, on `main`:
```
slowest 10 durations 
344.91s call     test/test_manifest.py::test_manifest_disksize[quay.io/centos-bootc/centos-bootc:stream9]
338.54s call     test/test_manifest.py::test_manifest_disksize[quay.io/fedora/fedora-bootc:40,btrfs]
78.21s setup    test/test_manifest.py::test_manifest_smoke[quay.io/centos-bootc/centos-bootc:stream9]
20.21s call     test/test_manifest.py::test_manifest_anaconda_module_customizations[quay.io/fedora/fedora-bootc:40,anaconda-iso,btrfs]
17.64s call     test/test_manifest.py::test_manifest_installer_customizations
16.61s call     test/test_manifest.py::test_iso_manifest_smoke[quay.io/fedora/fedora-bootc:40,anaconda-iso,btrfs]
16.48s call     test/test_manifest.py::test_iso_manifest_smoke[quay.io/centos-bootc/centos-bootc:stream9,anaconda-iso]
16.20s call     test/test_manifest.py::test_manifest_anaconda_module_customizations[quay.io/centos-bootc/centos-bootc:stream9,anaconda-iso]
16.19s call     test/test_manifest.py::test_manifest_target_arch_smoke[quay.io/centos-bootc/centos-bootc:stream9,arm64]
11.82s call     test/test_manifest.py::test_manifest_target_arch_smoke[quay.io/fedora/fedora-bootc:41,btrfs,ppc64le]
22 passed, 44 deselected in 960.91s (0:16:00) 
```
and on the PR branch:
```
slowest 10 durations 

339.93s call     test/test_manifest.py::test_manifest_disksize[quay.io/fedora/fedora-bootc:40,btrfs]
316.35s call     test/test_manifest.py::test_manifest_disksize[quay.io/centos-bootc/centos-bootc:stream9]
80.15s setup    test/test_manifest.py::test_manifest_smoke[quay.io/centos-bootc/centos-bootc:stream9]
18.54s call     test/test_manifest.py::test_iso_manifest_smoke[quay.io/fedora/fedora-bootc:40,anaconda-iso,btrfs]
17.63s call     test/test_manifest.py::test_manifest_anaconda_module_customizations[quay.io/fedora/fedora-bootc:40,anaconda-iso,btrfs]
17.54s call     test/test_manifest.py::test_manifest_installer_customizations
17.09s call     test/test_manifest.py::test_manifest_anaconda_module_customizations[quay.io/centos-bootc/centos-bootc:stream9,anaconda-iso]
16.52s call     test/test_manifest.py::test_manifest_target_arch_smoke[quay.io/centos-bootc/centos-bootc:stream9,arm64]
15.43s call     test/test_manifest.py::test_iso_manifest_smoke[quay.io/centos-bootc/centos-bootc:stream9,anaconda-iso]
12.23s call     test/test_manifest.py::test_manifest_target_arch_smoke[quay.io/fedora/fedora-bootc:41,btrfs,s390x]

22 passed, 44 deselected in 931.73s (0:15:31) 
```

I generally agree, we could reduce the number of these or make them faster by mocking some things and moving them to Go tests, but I think for now the impact of these new ones isn't high.

mvo5: Thanks, yeah, that is really a small difference, very nice!

bib/cmd/bootc-image-builder/image.go
achilleas-k: Added.  Thanks!

achilleas-k:
I applied @mvo5's patch for splitting out the PT handling and adds tests.
I also rebased on `main` and resolved the conflicts.

test/test_manifest.py
achilleas-k: You're not wrong.  I think we can reconsider some of these in a follow-up.

mvo5 (approved):
Lovely, thank you!

mvo5:
It fails right now with:
```
____ test_image_boots[quay.io/centos-bootc/centos-bootc:stream9,raw,arm64] _____
...
            # NOTE: xfs filesystems are ~40 MiB and ext4 ~26 MiB smaller than the partition - reduce minsize by 100 MiB
            minsize -= 100 * 2 ** 20
>           assert minsize < mountpoint_sizes[mountpoint]
E           assert 3116367872 < 3094126592

test/test_build.py:483: AssertionError
````
in

mvo5 (approved):
Thanks, this is fine! We *could* look into cleaning up the commit history a bit, but it's probably fine as is.

achilleas-k:
> It fails right now with:
> 
> ```
> ____ test_image_boots[quay.io/centos-bootc/centos-bootc:stream9,raw,arm64] _____
> ...
>             # NOTE: xfs filesystems are ~40 MiB and ext4 ~26 MiB smaller than the partition - reduce minsize by 100 MiB
>             minsize -= 100 * 2 ** 20
> >           assert minsize < mountpoint_sizes[mountpoint]
> E           assert 3116367872 < 3094126592
> 
> test/test_build.py:483: AssertionError
> ```
> 

Yup, that size comparison isn't great.  Was a bad idea to just take a chunk out of the expected size and call it a day.  I dropped the check and we can replace it with something a little smarter (see last commit), but I think we should leave that for another PR.

The partition sizing is either way tested at the manifest level.  The boot test now checks that all mountpoints exist when the system is booted.

achilleas-k:
> Thanks, this is fine! We _could_ look into cleaning up the commit history a bit, but it's probably fine as is.

It did get a bit long, I agree.  The last commit could be squashed, but let's keep my stupid attempt at a size comparison heuristic in there for posterity.

PR #608, opened 2024-08-14
bib: make build the default subcommand

ondrejbudai:
Prior this commit, the bootc-image-builder container image had
a custom entrypoint that hardcoded the use of the build subcommand. This
meant that if a user wanted to use a different subcommand, they had to
overwrite the entrypoint.

This commit changes the cobra code in bib to fallback to build if no
subcommand was given. This is slighly ugly, but it allows us to remove
the custom entrypoint, streamlining the use of subcommands. Let's see
an example of calling the version subcommand:

Before:
podman run --rm -it --entrypoint=/usr/bin/bootc-image-builder \
    quay.io/centos-bootc/bootc-image-builder:latest version

After:
sudo podman run --rm -it \
    quay.io/centos-bootc/bootc-image-builder:latest version

Kudos to https://github.com/IKukhta for his code from
https://github.com/spf13/cobra/issues/823#issuecomment-870027246

### DIFF
--- a/README.md
+++ b/README.md
@@ -522,8 +522,7 @@ The contents of the file `$(pwd)/wheel-passwordless-sudo` should be
 
 Please report bugs to the [Bug Tracker](https://github.com/osbuild/bootc-image-builder/issues) and include instructions to reproduce and the output of:
 ```
-$ sudo podman run --rm -it --entrypoint=/usr/bin/bootc-image-builder \
-    quay.io/centos-bootc/bootc-image-builder:latest version
+$ sudo podman run --rm -it quay.io/centos-bootc/bootc-image-builder:latest version
 ```
 
 ## 📊 Project
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -623,7 +623,6 @@ func run() error {
 	buildCmd.Flags().String("output", ".", "artifact output directory")
 	buildCmd.Flags().String("progress", "text", "type of progress bar to use")
 	buildCmd.Flags().String("store", "/store", "osbuild store for intermediate pipeline trees")
-
 	// flag rules
 	for _, dname := range []string{"output", "store", "rpmmd"} {
 		if err := buildCmd.MarkFlagDirname(dname); err != nil {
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -473,8 +473,7 @@ The contents of the file `$(pwd)/wheel-passwordless-sudo` should be
 
 Please report bugs to the [Bug Tracker](https://github.com/osbuild/bootc-image-builder/issues) and include instructions to reproduce and the output of:
 ```
-$ sudo podman run --rm -it --entrypoint=/usr/bin/bootc-image-builder \
-    quay.io/centos-bootc/bootc-image-builder:latest version
+$ sudo podman run --rm -it quay.io/centos-bootc/bootc-image-builder:latest version
```
mvo5, line 476: Love this!

bib/cmd/bootc-image-builder/main.go
```diff
@@ -638,6 +639,13 @@ func run() error {
 	}
 	buildCmd.MarkFlagsRequiredTogether("aws-region", "aws-bucket", "aws-ami-name")
 
+	// If no subcommand is given, assume the user wants to use the build subcommand
+	// See https://github.com/spf13/cobra/issues/823#issuecomment-870027246
+	cmd, _, err := rootCmd.Find(os.Args[1:])
+	if err == nil && cmd.Use == rootCmd.Use && cmd.Flags().Parse(os.Args[1:]) != pflag.ErrHelp {
+		args := append([]string{buildCmd.Use}, os.Args[1:]...)
```
mvo5, line 646: Thanks for figuring this out, this does not look entirely straightforward :)

mvo5 (dismissed):
Love it!

bib/cmd/bootc-image-builder/main.go
```diff
@@ -638,6 +637,17 @@ func run() error {
 	}
 	buildCmd.MarkFlagsRequiredTogether("aws-region", "aws-bucket", "aws-ami-name")
 
+	// If no subcommand is given, assume the user wants to use the build subcommand
+	// See https://github.com/spf13/cobra/issues/823#issuecomment-870027246
+	// which cannot be used verbatim because the arguments for "build" like
+	// "quay.io" will create an "err != nil". Ideally we could check err
+	// for something like cobra.UnknownCommandError but cobra just gives
+	// us an error string
+	_, _, err := rootCmd.Find(os.Args[1:])
```
mvo5, line 646: If https://github.com/spf13/cobra/pull/2178 gets accepted we can do the check below in a more precise way, i.e. something like `if _, ok := err.(cobra.UnknownCommandError); ok {`

ondrejbudai: Right, thanks for double-checking this! :) Let's get this merged?

mvo5 (dismissed):

bib/cmd/bootc-image-builder/main.go
```diff
@@ -635,6 +634,17 @@ func run() error {
...
+	if err != nil && !slices.Contains([]string{"help", "completion"}, os.Args[1]) {
+		args := append([]string{buildCmd.Use}, os.Args[1:]...)
```
achilleas-k, line 645: I think this should be `buildCmd.Name()`, which is the first word in the `Use` line.
The general form of `Use` is the full usage instructions of the command, including optional and position argument hints [1].  We don't use that and we only include the name of the command, but let's not have to deal with that if we ever extend the usage lines.

[`buildCmd.Name()`](https://pkg.go.dev/github.com/spf13/cobra#Command.Name) instead will always return the first word, even if we extend it.

https://pkg.go.dev/github.com/spf13/cobra#Command

achilleas-k (changes requested):

bib/cmd/bootc-image-builder/main.go
mvo5: Thanks, nice catch! Updated

achilleas-k (approved):
Very nice.  LGTM

PR #611, opened 2024-08-19
bib: update to images v0.78 and update for API changes

mvo5:
Two changes:
- The pathpolicy.PathTri type is no longer exported and it is `pathpolicy.Pathpolicies` instead [0]
- The error messages from images for path policy violations changes and need an update (arguable this is not the best layer to test the error  details here and instead we should check how much we can move down to images but that is a followup :)

[0] https://github.com/osbuild/images/pull/854

### DIFF
--- a/bib/cmd/bootc-image-builder/image.go
+++ b/bib/cmd/bootc-image-builder/image.go
@@ -116,7 +116,7 @@ var (
 	})
 )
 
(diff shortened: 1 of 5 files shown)


ondrejbudai (dismissed):
Thanks!

achilleas-k (approved):
Thank you!

achilleas-k (dismissed):

achilleas-k (approved):

Issue #614, opened 2024-08-21
--local usage was broken on latest bib build with error: unknown flag: --type


chunfuwen:
1.  pull bootc image:
```
podman pull quay.io/centos-bootc/centos-bootc:stream10

```
2. run bib command to build qcow image:

```
sudo podman run --rm -it --privileged --pull=newer --security-opt label=type:unconfined_t -v /var/lib/libvirt/images/output:/output -v /var/lib/containers/storage:/var/lib/containers/storage  -v /var/lib/libvirt/images/auth.json:/run/containers/0/auth.json  quay.io/centos-bootc/bootc-image-builder:latest  --type qcow2 --tls-verify=true  --local quay.io/centos-bootc/centos-bootc:stream10

``` 
3. Actually it throw below error message:
```
sudo podman run --rm -it --privileged --pull=newer --security-opt label=type:unconfined_t -v /var/lib/libvirt/images/output:/output -v /var/lib/containers/storage:/var/lib/containers/storage  -v /var/lib/libvirt/images/auth.json:/run/containers/0/auth.json  quay.io/centos-bootc/bootc-image-builder:latest  --type qcow2 --tls-verify=true  --local quay.io/centos-bootc/centos-bootc:stream10
Usage:
  bootc-image-builder [command]

Available Commands:
  build       
  completion  Generate the autocompletion script for the specified shell
  help        Help about any command
  manifest    

Flags:
  -h, --help               help for bootc-image-builder
      --log-level string   logging level (debug, info, error); default error

Use "bootc-image-builder [command] --help" for more information about a command.

2024/08/21 02:09:35 error: unknown flag: --type

```

chunfuwen:
@mvo5  please help take a look

shi2wei3:
Em, it looks like this only happens if you put `--local` as the last argument, otherwise it's fine.

henrywang:
Hi @mvo5, this issue is already in `bootc-image-builder:rhel-9.5`. We need a downstream release to fix this issue.

datcuandrei:
If you want to use `--local`, use the last image tag: `sha256-5ccc2079c6e28eb96df3a63e5babe221c8a28c5367aec8f9613cc126e729f76d`

Hopefully, this will be fixed in the next release.

PR #615, opened 2024-08-21
bib: fix injecting default "build" arg into root for flags like --local

mvo5:
[draft as this needs tests]

In PR#608 we dropped the container entrypoint and moved the logic to detect what command to run into the cobra parsing.

However it seems certain situations like:
```
bootc-image-builder  --type qcow2 --tls-verify=true  --local quay.io/centos-bootc/centos-bootc:stream10
```
are not detected correctly (thanka to Chunfu Wen for reporting!).

This commit detect this usage and injects the build arg then too.

Closes https://github.com/osbuild/bootc-image-builder/issues/614

### DIFF
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 	"golang.org/x/exp/slices"
 
 	"github.com/osbuild/images/pkg/arch"
@@ -565,7 +566,7 @@ func cmdVersion(_ *cobra.Command, _ []string) error {
 	return nil
 }
 
-func run() error {
+func buildCobraCmdline() (*cobra.Command, error) {
 	rootCmd := &cobra.Command{
 		Use:               "bootc-image-builder",
 		Long:              "create a bootable image from an ostree native container",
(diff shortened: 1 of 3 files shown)


bib/cmd/bootc-image-builder/main.go
```diff
@@ -640,11 +641,29 @@ func run() error {
...
 	// for something like cobra.UnknownCommandError but cobra just gives
 	// us an error string
-	_, _, err := rootCmd.Find(os.Args[1:])
-	if err != nil && !slices.Contains([]string{"help", "completion"}, os.Args[1]) {
+	cmd, _, err := rootCmd.Find(os.Args[1:])
+	injectBuildArg := func() {
 		args := append([]string{buildCmd.Name()}, os.Args[1:]...)
 		rootCmd.SetArgs(args)
 	}
+	// command not known, i.e. happens for "bib quay.io/centos/..."
```
mvo5, line 649: This is slightly ugly but I couldn't find a better way than to handle the two cases explicitly

ondrejbudai (approved):
Ugh, sorry for missing this case, and thanks for looking into this! :)

bib/cmd/bootc-image-builder/main.go
ondrejbudai: :shrug: I have no better idea.

achilleas-k:
I'm going to admin-merge this because we need it to unblock users but we have an issue with our cross-arch build tests that's preventing things from being merged through the proper merge queue.

PR #617, opened 2024-08-21
bib: add new `--partition-mode` switch to enable lvm

mvo5:
This allows us to quickly enable lvm. Draft as I think we need to decide on the UX, i.e. if this should be done via the blueprint config or commandline or both and how it interacts with --rootfs and btrfs.

This needs:
~https://github.com/osbuild/images/pull/876~
~https://github.com/osbuild/osbuild/pull/1867~
~https://github.com/osbuild/osbuild/pull/1862~

but with that I was able to:
```
$ bootc-image-builder manifest --partition-mode lvm quay.io/centos-bootc/centos-bootc:stream9
...
$ kvm -m 1500 -cpu host disk.raw
...
localhost login: root
Password: 
[systemd]
Failed Units: 1
  bootc-generic-growpart.service
[root@localhost ~]# mount|grep "/sysroot "
/dev/mapper/rootvg-rootlv on /sysroot type xfs (ro,relatime,seclabel,attr2,inode64,logbufs=8,logbsize=32k,noquota)
```

### DIFF
--- a/bib/cmd/bootc-image-builder/image.go
+++ b/bib/cmd/bootc-image-builder/image.go
@@ -63,6 +63,9 @@ type ManifestConfig struct {
 
 	// RootFSType specifies the filesystem type for the root partition
 	RootFSType string
+
+	// PartitionMode specifies the partiton mode to use (raw, lvm)
+	PartitionMode string
 }
 
 func Manifest(c *ManifestConfig) (*manifest.Manifest, error) {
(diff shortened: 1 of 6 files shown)


cgwalters:
```
Failed Units: 1
  bootc-generic-growpart.service
```

should be fixed by https://gitlab.com/fedora/bootc/base-images/-/merge_requests/33

oglok:
This would be great. Is this targeting any disk format (qcow2, raw, ami...)? (not taking into account ISO as we can use kickstart for that)

mvo5:
> This would be great. Is this targeting any disk format (qcow2, raw, ami...)? (not taking into account ISO as we can use kickstart for that)

Thanks for the feedback. This will land in slightly modified from directly via the "blueprints". The details are currently being worked on in https://github.com/osbuild/images/pull/926

mvo5:
This can be closed now that we have https://github.com/osbuild/bootc-image-builder/pull/725 - with 725 adding lvm/btrfs is now possible with much more control than it was with this prototype.

mvo5:
> This would be great. Is this targeting any disk format (qcow2, raw, ami...)? (not taking into account ISO as we can use kickstart for that)

Ups, missed this question. Yes, the target is disk formats, and with the latest support in bib for arbitirary kicksart content the installer images should also be capable of doing this (but we won't translate the requests from the blueprint into kickstart files automatically)

Issue #620, opened 2024-08-26
centos 10 stream iso vm is stuck in starting


chunfuwen:
1.  Build out centos stream 10 iso using bib as below:
```
sudo podman run --rm -it --privileged --pull=newer --security-opt label=type:unconfined_t -v /var/lib/libvirt/images/output:/output -v /var/lib/libvirt/images/config.json:/config.json   -v /var/lib/libvirt/images/auth.json:/run/containers/0/auth.json  quay.io/centos-bootc/bootc-image-builder:latest  --type anaconda-iso --tls-verify=true  --config /config.json  quay.io/centos-bootc/centos-bootc:stream10  --chown 107:107  --rootfs ext4 '
```
2. use virt-install to install VM from generated iso with below command:
```
virt-install --name install_upstream_bib_bios_install-centos-bootc --disk path=/var/lib/libvirt/images/install_upstream_bib_bios_install-centos-bootc_bios.qcow2,bus=virtio,format=qcow2,size=12 --vcpus 3 --memory 3096 --osinfo detect=on,require=off --graphics vnc --video virtio --serial pty --wait 10 --cdrom /var/lib/libvirt/images/install_upstream_bib_bios_install-centos-bootc.iso --debug --machine q35 --noreboot
```
3. The VM is stuck in booting as attached screenshot
![Screenshot from 2024-08-26 13-11-48](https://github.com/user-attachments/assets/4ae5b226-a29b-45fe-8201-c7fe2d4d525e)

mvo5:
I started looking into this in my morning and made some progress:
- with the following diff [centos10-iso.diff.txt](https://github.com/user-attachments/files/16745645/centos10-iso.diff.txt)
- it boots to 
![Screenshot from 2024-08-26 09-42-39](https://github.com/user-attachments/assets/226c73af-7ec5-45d9-bcaa-42cbe8cccdc4)

If I manually press "text" install it will finish and reboot and our smoke test works. So it's "just" a matter now of figuring out why it claims `Wayland was unable to start on your machine` and how ti fix that.

mvo5:
Some more data, it seems anaconda is crashing when it tries to read `osbuild.ks`. The error is that it cannot load the ca-certificates bundle while importing the python3-requests package (which it does not really need for getting the file locally fwiw). 

The package is installed in the boot system but it seems it is not part of the initramfs which is (AIUI) when dracut-initqueue is run. 

[...]

PR #621, opened 2024-08-26
test: add test for anaconda install on centos10

mvo5:
Small reproducer for https://github.com/osbuild/bootc-image-builder/issues/620

To test run:
```
$ sudo OSBUILD_TEST_QEMU_GUI=1 pytest -s -vv ./test/test_build.py::test_iso_installs[quay.io/centos-bootc/centos-bootc:stream10,anaconda-iso]
```

We probably do not want to merge as is (even if it woudl work) as it adds significant extra time to our tests :/

### DIFF
--- a/bib/data/defs/centos-10.yaml
+++ b/bib/data/defs/centos-10.yaml
@@ -27,6 +27,7 @@ anaconda-iso:
     - gdisk
     - glibc-all-langpacks
     - gnome-kiosk
+    - gnome-remote-desktop
     - google-noto-sans-cjk-ttc-fonts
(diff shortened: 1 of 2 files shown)


cgwalters:
> We probably do not want to merge as is (even if it woudl work) as it adds significant extra time to our tests :/

I think it makes sense to cover "expensive" integration tests as part of the (ideally jointly owned) bootc integration tests right?

mvo5:
This will be fixed via https://github.com/osbuild/bootc-image-builder/pull/625 and then we need to think about the test (and it's place).

PR #647, opened 2024-09-13
bib: add/use new `imagestype` package as a superset of the `BuildType`

mvo5:
When adding new image types like `vhd` or the coming `gcp` I noticed that we currently need to modify two different places to add a new image type. This PR consolidates the two into a single new package `imagetypes` that consolidates all knowledge of image types, iso, exports and validation into a single place. So adding `gcp` shoudl be as simple as:

```go
diff --git a/bib/internal/imagetypes/imagetypes.go b/bib/internal/imagetypes/imagetypes.go
index 4e0dc51..16a3116 100644
--- a/bib/internal/imagetypes/imagetypes.go
+++ b/bib/internal/imagetypes/imagetypes.go
@@ -20,6 +20,7 @@ var supportedImageTypes = map[string]imageType{
        "vhd":          imageType{Export: "vpc"},
        "anaconda-iso": imageType{Export: "anaconda-iso", ISO: true},
        "iso":          imageType{Export: "iso", ISO: true},
+       "gcp":          imageType{Export: "gcp"},
 }
 
 // Available() returns a comma-separated list of supported image types
```
(once images has it, plus of course adding it to the tests)


achilleas-k (dismissed):
Nice and clean!

mvo5:
@achilleas-k Thank you! tests caught a silly copy/paste mistake so I had to re-push but it should be good now.

Issue #650, opened 2024-09-18
`bootc-image-builder` fails to build  `quay.io/fedora/fedora-bootc:42` image


hrismarin:
**Describe the bug**
When I try to build an image from `quay.io/fedora/fedora-bootc:42`, `bootc-image-builder` fails.

**To Reproduce**
1. Pull the image from the registry.
`sudo podman image pull quay.io/fedora/fedora-bootc:42`
2. Create the required output directory for the built base image.
`mkdir output`
3. Build the image
```
sudo podman container run \
    --rm \
    --interactive --tty \
    --privileged \
    --pull=newer \
    --security-opt label=type:unconfined_t \
    --volume $(pwd)/config.toml:/config.toml:ro \
    --volume $(pwd)/output:/output \
    --volume /var/lib/containers/storage:/var/lib/containers/storage \
    quay.io/centos-bootc/bootc-image-builder:latest \
    --rootfs btrfs \
    --type qcow2 \
    --local \
    quay.io/fedora/fedora-bootc:42

Generating manifest manifest-qcow2.json
2024/09/18 09:56:18 error: cannot build manifest: initializing dnf in 3895ee04b9b70f11529ebd3d615ea558cd292e91a3711bd8ca3af74e56fc23dd container failed: exit status 2
output:
Missing command. Add "--help" for more information about the arguments.
```
**Expected behavior**
The image is built successfully.

**Additional context**
See: https://github.com/osbuild/bootc-image-builder/pull/651

PR #651, opened 2024-09-18
bib: run `dnf --version` in the container

hrismarin:
Recently, in bootc Base Images main branch the `dnf` package was replaced by `dnf5`.
See: https://gitlab.com/fedora/bootc/base-images/-/merge_requests/45

As a consequence `bootc-image-builder` now fails to build the `quay.io/fedora/fedora-bootc:42` image.
See: https://github.com/osbuild/bootc-image-builder/issues/650

Running plain `dnf` command in dnf5 output:
Missing command. Add "--help" for more information about the arguments.

The `dnf --version` command does not output non-zero exit code in both dnf and dnf5.


henrywang:
I reported an dnf5 issue https://bugzilla.redhat.com/show_bug.cgi?id=2313032

hrismarin:
I think it has to do with the fact that dnf and dnf5 behave differently when running a plain `dnf` command (no options or arguments).

In case of dnf (dnf 4) it otputs the vesion info and the exit code is `0`.

In case of dnf5 the output is `Missing command. Add "--help" for more information about the arguments.` and the exit code is `2`. See https://dnf5.readthedocs.io/en/latest/dnf5.8.html#exit-codes.

Note that I am not making any changes to the `quay.io/fedora/fedora-bootc:42` image i.e. I'm not building a derived image. I'm just trying to build the image straight from the registry as I explained in https://github.com/osbuild/bootc-image-builder/issues/650.

Filing a dnf5 issue on Bugzilla was probably the better option. Sorry for the noise if this PR doesn't make sense.

sigulete:
Same problem here when trying to create an image from `quay.io/fedora-ostree-desktops/kinoite:41`

supakeen (approved):
Can reproduce the exit codes vs dnf4/dnf5. I'm unsure if this actually initializes dnf when ran with `--version`? (cc @mvo5 / @achilleas-k).

mvo5:
Sorry that we left you hanging here for so long, I suspect `dnf --version` is not quite equivalent to what we need but I'm trying to reproduce this now and add a test-case.

ondrejbudai:
Running `dnf` is required to initialize repositories from `subscription-manager` and its dnf-plugin. I'm pretty sure that `dnf --version` doesn't do that (I ran a quick test).

My take on this is: let's not worry about DNF 5 here. Both RHEL 9 and 10 use DNF 4 and these are the only systems that support image mode and require subscriptions. So I would the following thing: if dnf is version 4, run `dnf`, else do nothing. Add a TODO to resolve this when RHEL 11 is going to be released (~3+ years).

mvo5:
> Running `dnf` is required to initialize repositories from `subscription-manager` and its dnf-plugin. I'm pretty sure that `dnf --version` doesn't do that (I ran a quick test).
> 
> My take on this is: let's not worry about DNF 5 here. Both RHEL 9 and 10 use DNF 4 and these are the only systems that support image mode and require subscriptions. So I would the following thing: if dnf is version 4, run `dnf`, else do nothing. Add a TODO to resolve this when RHEL 11 is going to be released (~3+ years).

I opened https://github.com/osbuild/bootc-image-builder/pull/658 with an alternative idea that might give us both.

hrismarin:
> So I would the following thing: if dnf is version 4, run dnf, else do nothing. Add a TODO to resolve this when RHEL 11 is going to be released (~3+ years).

Seems like a reasonable solution to me. Should I close this PR?

mvo5:
> > So I would the following thing: if dnf is version 4, run dnf, else do nothing. Add a TODO to resolve this when RHEL 11 is going to be released (~3+ years).
> 
> Seems like a reasonable solution to me. Should I close this PR?

It seems as is this PR would need some work, either detecting dnf4/dnf5 (and ignoring dnf5 like Ondrej suggested) or going with the alteratnive idea, I would like to credit you in the other PR though, I will need to remember to update/force-push. The only downside of `dnf clean` (in the other PR) is that it is slow on dnf5, I will check if there is another way.

hrismarin:
Please let me know if I can help with anything like testing etc.

mvo5:
@hrismarin The PR in https://github.com/osbuild/bootc-image-builder/pull/658 has landed now so the next `bib` container build should work. I think we can close this PR then.

Issue #655, opened 2024-09-24
bib failed to build out config.toml with customizations.filesystem.mountpoint


chunfuwen:
step 1 create one config.toml with customization.filesystem as below:
```


[root@dell-per430-17 tmt]# cat /var/lib/libvirt/images/config.toml
[[customizations.filesystem]]
mountpoint = "/"
minsize = "10 GiB"

[[customizations.filesystem]]
mountpoint = "/var/data"
minsize = "20 GiB"

```
step 2 , run bib to generate image, but throw error as below:
```
(92lts-p3-ve) [root@dell-per430-17 tp-libvirt]# sudo podman run --rm -it --privileged --pull=newer --security-opt label=type:unconfined_t -v /var/lib/libvirt/images/output:/output -v /var/lib/libvirt/images/config.toml:/config.toml   -v /var/lib/libvirt/images/auth.json:/run/containers/0/auth.json  quay.io/centos-bootc/bootc-image-builder:latest  --type vhd --tls-verify=false  --config /config.toml  quay.io/fedora/fedora-bootc:latest  --rootfs xfs
Generating manifest manifest-vhd.json
2024/09/24 09:08:34 error: cannot build manifest: cannot read config: cannot decode "/config.toml": unknown keys found: [customizations.filesystem.mountpoint customizations.filesystem.minsize customizations.filesystem.mountpoint customizations.filesystem.minsize]

```

mvo5:
I can reproduce and know what the issue is, it's a regression from https://github.com/osbuild/bootc-image-builder/pull/549 - it's not yet clear what the best fix but I'm on it

mvo5:
The underlying issue is https://github.com/osbuild/images/pull/951 - this needs some tweaking/thinking, we should probably do a quicker fix (by disabling the meta.Undecoded() check) short term as this is a rather big regression :/

Thanks for catching/reporting this!

PR #656, opened 2024-09-24
buildconfig: disable tests for undecoded keys for now

mvo5:
This commit partially reverts PT#549 to unblock filesystem
customizations in bib.

This is a short term fix and we should revert and do something
smarter like https://github.com/osbuild/images/pull/951 or see
if we can do better in the toml unmarshaling. But to unblock
toml customizations this is a (IMHO) reasonable first step.

Closes: https://github.com/osbuild/bootc-image-builder/issues/655


mvo5:
An alterative upstream fix: https://github.com/BurntSushi/toml/pull/426 (might be too naive, let's hear the maintainer :)

ondrejbudai (approved):
Yeah, this is reasonable, thanks!

PR #658, opened 2024-09-25
container: run `dnf check` in container for dnf4/dnf5 compat

mvo5:
This is an alternative idea to PR#651 - we run `dnf` inside the
container to init the dnf system. However with `dnf5` this no
longer works and it just throws an error.

So instead we run `dnf check` which should work on both dnf{4,5}.

---

test: move `target-arch-smoke` testcases to Fedora42 (for
 dnf5)

This will give us testing for dnf5 as well.

[draft until tests are good and I double checked dnf sources]

### DIFF
--- a/bib/internal/container/container.go
+++ b/bib/internal/container/container.go
@@ -123,13 +123,20 @@ func (c *Container) ExecArgv() []string {
 	return []string{"podman", "exec", "-i", c.id}
 }
 
-// InitDNF initializes dnf in the container. This is necessary when the caller wants to read the image's dnf
-// repositories, but they are not static, but rather configured by dnf dynamically. The primaru use-case for
-// this is RHEL and subscription-manager.
+// InitDNF initializes dnf in the container. This is necessary when
+// the caller wants to read the image's dnf repositories, but they are
+// not static, but rather configured by dnf dynamically. The primaru
+// use-case for this is RHEL and subscription-manager.
 //
-// The implementation is simple: We just run plain `dnf` in the container.
+// The implementation is simple: We just run plain `dnf` in the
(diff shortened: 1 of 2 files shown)


ondrejbudai:
It works just fine with dnf 4:

```console
# time dnf check
Updating Subscription Management repositories.

________________________________________________________
Executed in    1.80 secs      fish           external
   usr time    8.23 millis  229.00 micros    8.00 millis
   sys time   17.53 millis  333.00 micros   17.20 millis
```

However, it's quite slow with dnf 5:


```console
# time dnf check

________________________________________________________
Executed in   24.95 secs      fish           external
   usr time    5.38 millis  314.00 micros    5.07 millis
   sys time    9.21 millis   37.00 micros    9.17 millis
```

mvo5:
> It works just fine with dnf 4:
> 
> ```
> # time dnf check
> Updating Subscription Management repositories.
> 
> ________________________________________________________
> Executed in    1.80 secs      fish           external
>    usr time    8.23 millis  229.00 micros    8.00 millis
>    sys time   17.53 millis  333.00 micros   17.20 millis
> ```
> 
> However, it's quite slow with dnf 5:
> 
> ```
> # time dnf check
> 
> ________________________________________________________
> Executed in   24.95 secs      fish           external
>    usr time    5.38 millis  314.00 micros    5.07 millis
>    sys time    9.21 millis   37.00 micros    9.17 millis
> ```

Thanks for checking this ! I switched now to use `dnf check --duplicates` which takes around 1s for me compared to the 20s the `dnf check` takes with dnf5. That seems like a reasonable compromise(?).

ondrejbudai (approved):
Checked on my system, works like a charm! Thanks!

Issue #265, opened 2024-03-14
ISO install Error using virt-install :Error validating install location: Could not find an installable distribution at URL '/var/lib/libvirt/images/install.iso'


chunfuwen:
```
step 1 : generate iso file:

podman run --rm -it --privileged --pull=newer --security-opt label=type:unconfined_t -v $(pwd)/config.json:/config.json -v $(pwd)/output:/output quay.io/centos-bootc/bootc-image-builder:latest quay.io/centos-bootc/fedora-bootc:eln --type iso --tls-verify  --chown 107:107

step 2:  use virt-install install iso file:

virt-install --connect qemu:///system -n imagevm --hvm --accelerate -r 3096 --vcpus=3 --os-variant rhel9.0 --disk path=/var/lib/libvirt/images/testimgevm.qcow2,bus=virtio,format=qcow2,size=10 --network network=default,model=virtio  --location /var/lib/libvirt/images/install.iso --debug --memballoon model=virtio --cpu host,disable=vmx --graphics vnc --video virtio --machine q35


Result : virt-install throw error and terminate the installation


```

chunfuwen:
look like the issue is already there

chunfuwen:
root@hp-dl360g9-16 bootc-image-builder]# isoinfo -d -i /var/lib/libvirt/images/install.iso 
CD-ROM is in ISO 9660 format
System id: LINUX
Volume id: Container-Installer-x86_64
Volume set id: 
Publisher id: 
Data preparer id: XORRISO-1.5.6 2023.06.07.180001, LIBISOBURN-1.5.6, LIBISOFS-1.5.6, LIBBURN-1.5.6
Application id: 
Copyright File id: 
Abstract File id: 
Bibliographic File id: 
Volume set size is: 1
Volume set sequence number is: 1
Logical block size is: 2048
Volume size is: 873976
El Torito VD version 1 found, boot catalog is in sector 57
Joliet with UCS level 3 found
Rock Ridge signatures version 1 found
Eltorito validation header:
    Hid 1
    Arch 0 (x86)
    ID ''
    Key 55 AA
    Eltorito defaultboot header:
        Bootid 88 (bootable)
        Boot media 0 (No Emulation Boot)
        Load segment 0
        Sys type 0
        Nsect 4
        Bootoff 3A 58

ondrejbudai:
[...]

PR #667, opened 2024-10-08
Use a libosinfo compatible label format for the Anaconda ISO

mmartinv:
Label the anaconda ISO depending on the container OS
and make the ISO label format compatible with libosinfo.

This will make any software using libosinfo to be able
to correctly detect the OS contained in the ISO and
where to find the initrd and the kernel.

Resolves: https://github.com/osbuild/bootc-image-builder/issues/265

### DIFF
--- a/bib/cmd/bootc-image-builder/image.go
+++ b/bib/cmd/bootc-image-builder/image.go
@@ -322,6 +322,24 @@ func manifestForDiskImage(c *ManifestConfig, rng *rand.Rand) (*manifest.Manifest
 	return &mf, err
 }
 
+func labelForISO(os *source.OSRelease, arch *arch.Arch) string {
+	switch os.ID {
+	case "fedora":
+		return fmt.Sprintf("Fedora-S-dvd-%s-%s", arch, os.VersionID)
+	case "centos":
+		labelTemplate := "CentOS-Stream-%s-BaseOS-%s"
+		if os.VersionID == "8" {
+			labelTemplate = "CentOS-Stream-%s-%s-dvd"
+		}
+		return fmt.Sprintf(labelTemplate, os.VersionID, arch)
+	case "rhel":
+		version := strings.ReplaceAll(os.VersionID, ".", "-")
+		return fmt.Sprintf("RHEL-%s-BaseOS-%s", version, arch)
+	default:
+		return fmt.Sprintf("Container-Installer-%s", arch)
+	}
+}
+
 func manifestForISO(c *ManifestConfig, rng *rand.Rand) (*manifest.Manifest, error) {
 	if c.Imgref == "" {
 		return nil, fmt.Errorf("pipeline: no base image defined")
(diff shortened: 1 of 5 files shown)


test/testcases.py
```diff
@@ -41,19 +41,22 @@ def __str__(self):
 class TestCaseFedora(TestCase):
     container_ref: str = "quay.io/fedora/fedora-bootc:40"
     rootfs: str = "btrfs"
+    osinfo_template: str = "Fedora Server 40 (%s)"
```
mvo5, line 44: (nitpick) Might be worthwhile to add a small comment that the %s will be replaced with the architecture

test/test_build.py
```diff
@@ -452,6 +453,25 @@ def test_iso_installs(image_type):
...
+    if not arch:
+        arch = platform.machine()
+    osinfo = image_type.osinfo_template % arch
+    result = subprocess.run([
+        "osinfo-detect",
+        installer_iso_path,
+    ], capture_output=True, text=True, check=True)
+    osinfo_output = result.stdout
+    expected_output = f"Media is bootable.\nMedia is an installer for OS '{osinfo}'\n"
+    print(f"osinfo-detect output:\n{osinfo_output}\nexpected output:\n{expected_output}")
```
mvo5, line 471: (nitick) given that we assert the two below I guess this is not needed?

test/test_build.py
```diff
@@ -452,6 +453,25 @@ def test_iso_installs(image_type):
         assert_kernel_args(vm, image_type)
 
 
+@pytest.mark.skipif(platform.system() != "Linux", reason="osinfo detect test only runs on linux right now")
+@pytest.mark.parametrize("image_type", gen_testcases("anaconda-iso"), indirect=["image_type"])
+def test_iso_os_detection(image_type):
+    installer_iso_path = image_type.img_path
+    arch = image_type.img_arch
+    template = image_type.osinfo_template
+    if not arch:
```
mvo5, line 462: (nitpick) I would move this line right next to "arch = image_type.img_arch" as they belong together

bib/cmd/bootc-image-builder/image.go
```diff
@@ -322,6 +322,26 @@ func manifestForDiskImage(c *ManifestConfig, rng *rand.Rand) (*manifest.Manifest
 	return &mf, err
 }
 
+func labelForISO(os *source.OSRelease, arch *arch.Arch) string {
+	var label string
+	switch os.ID {
+	case "fedora":
+		label = fmt.Sprintf("Fedora-S-dvd-%s-%s", arch, os.VersionID)
```
mvo5, line 329: (niptick) this could just be a `return fmt.Sprintf("Fedora-S-dvd-%s-%s", arch, os.VersionID)` and he samefor all the other cases

mvo5 (dismissed):
Thanks, this is very nice. I approved but there are a bunch of nitpicks inline that might be worthwhile to fix but I/we could do this as a followup (or ignore, some are just preferences I guess)

test/testcases.py
mmartinv: I used a different string formatting so the comment is not needed, please take a look

mvo5: Love it, thank you!

mvo5 (approved):
This is very nice! Thank you

ondrejbudai (approved):
Thanks!

henrywang:
Yeah! I can use `--location` in `virt-install`.

PR #669, opened 2024-10-08
bib: run "dnf" inside the container again

mvo5:
In 17d3b56 osbuild-dnf-json was changed to run outside the container. This lead to a regression in accessing subscribed content. This commit partially reverts this commit to run dnf again inside the container so that we have access to the /run/secrets and RHEL repos.

[a different approach over pr#668, this one works in the (not yet public) test-case we have]


mvo5:
Closing in favor of https://github.com/osbuild/bootc-image-builder/pull/670 which has a (slightly) cleaner implementation and way more tests.

PR #670, opened 2024-10-09
bib: run "dnf" inside the container again

mvo5:
In 17d3b56 osbuild-dnf-json was changed to run outside the
container. This lead to a regression in accessing subscribed
content. This commit partially reverts this commit to run
dnf again inside the container so that we have access to
the /run/secrets and RHEL repos.

This also adds a bunch of extra tests that needs to run
on a fedora/rhel/centos machine to test dnfjson and
subscriptions inside the container environment. Those
will only run in testingfarm (or locally) not in GH actions.

Closes: https://issues.redhat.com/browse/BIFROST-429

P.S. We should probably also look into how to inject osbuild-dnf-json into
the container, this way is not ideal, we maybe need to reconsider
putting it all into a single file again or think about other ways to make
this slightly easier. The tests/refactor hopefully makes this slightly easier
now.

### DIFF
--- a/.github/workflows/testingfarm-unit.yml
+++ b/.github/workflows/testingfarm-unit.yml
@@ -1,28 +1,11 @@
 ---
-name: Testing farm tests
+name: Testing farm go unit tests
 
 on:
   pull_request_target:
     types: [opened, synchronize, reopened]
 
-# To use testing farm we need the TF_API_KEY secret available inside the
-# forked repo which requires the pull_request_target trigger. To protect
-# the secrets we need to make sure only people with repo write access
-# can trigger this workflow. This means that ouside contributors will
-# get an initial failure when the workflow is run. But once someone from
-# the team re-triggers it it will work.
-#
-# Note that "pull_requqest_target" events are always triggered even
-# when the "Fork pull request workflows from outside collaborators"
-# setting is restricted to "Require approval for all outside collaborators"
-# (see https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks)
-#
-# Note also that this precautions might be overkill because a fork
-# cannot modify this workflow and all we do is run a branch inside
-# testing farm. But a) the scope of workflow may expand over time
-# b) it feels safer this way and is not a big burden in practise.
-#
-# This follows https://michaelheap.com/access-secrets-from-forks/
+# see testingfarm.yml
 jobs:
   testingfarm:
     name: "Run in testing farm"
@@ -51,7 +34,7 @@ jobs:
       uses: sclorg/testing-farm-as-github-action@v3
       with:
         compose: Fedora-40
-        tmt_plan_filter: "integration.fmf"
+        tmt_plan_filter: "unit-go.fmf"
         api_key: ${{ secrets.TF_API_KEY }}
         git_url: ${{ github.event.pull_request.head.repo.clone_url }}
         git_ref: ${{ github.event.pull_request.head.ref }}
--- a/bib/internal/cntdnf/cntdnf.go
+++ b/bib/internal/cntdnf/cntdnf.go
@@ -1,0 +1,48 @@
+package cntdnf
+
+import (
+	"fmt"
+	"path/filepath"
+
+	"github.com/osbuild/images/pkg/arch"
+	"github.com/osbuild/images/pkg/dnfjson"
+
+	"github.com/osbuild/bootc-image-builder/bib/internal/container"
+	"github.com/osbuild/bootc-image-builder/bib/internal/source"
+)
+
+func injectDNFJson(cnt *container.Container) ([]string, error) {
+	if err := cnt.CopyInto("/usr/libexec/osbuild-depsolve-dnf", "/osbuild-depsolve-dnf"); err != nil {
+		return nil, fmt.Errorf("cannot prepare depsolve in the container: %w", err)
+	}
+	// copy the python module too
+	globPath := "/usr/lib/*/site-packages/osbuild"
+	matches, err := filepath.Glob(globPath)
+	if err != nil || len(matches) == 0 {
+		return nil, fmt.Errorf("cannot find osbuild python module in %q: %w", globPath, err)
+	}
+	if len(matches) != 1 {
+		return nil, fmt.Errorf("unexpected number of osbuild python module matches: %v", matches)
+	}
+	if err := cnt.CopyInto(matches[0], "/"); err != nil {
+		return nil, fmt.Errorf("cannot prepare depsolve python-modules in the container: %w", err)
+	}
+	return append(cnt.ExecArgv(), "/osbuild-depsolve-dnf"), nil
+}
+
+func NewContainerSolver(cacheRoot string, cnt *container.Container, architecture arch.Arch, sourceInfo *source.Info) (*dnfjson.Solver, error) {
+	depsolverCmd, err := injectDNFJson(cnt)
+	if err != nil {
+		return nil, fmt.Errorf("cannot inject depsolve into the container: %w", err)
+	}
+
+	solver := dnfjson.NewSolver(
+		sourceInfo.OSRelease.PlatformID,
+		sourceInfo.OSRelease.VersionID,
+		architecture.String(),
+		fmt.Sprintf("%s-%s", sourceInfo.OSRelease.ID, sourceInfo.OSRelease.VersionID),
+		cacheRoot)
+	solver.SetDNFJSONPath(depsolverCmd[0], depsolverCmd[1:]...)
+	solver.SetRootDir("/")
+	return solver, nil
+}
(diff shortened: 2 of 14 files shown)


bib/internal/container/container.go
```diff
@@ -143,6 +143,17 @@ func (c *Container) InitDNF() error {
 	return nil
 }
 
+func (c *Container) InitDepSolveDNF() ([]string, error) {
+	if err := c.CopyInto("/usr/libexec/osbuild-depsolve-dnf", "/osbuild-depsolve-dnf"); err != nil {
+		return nil, fmt.Errorf("cannot prepare depsolve in the container: %w", err)
+	}
+	// XXX: hardcoded python3.12
+	if err := c.CopyInto("/usr/lib//python3.12/site-packages/osbuild", "/"); err != nil {
```
achilleas-k: What's the long term plan for this?

achilleas-k (approved):
LGTM.

Couple of questions.

bib/internal/container/container_test.go
```diff
@@ -171,3 +175,52 @@ echo '%s'
...
 }
+
+func subscribeMachine(t *testing.T) (restore func()) {
+	if _, err := exec.LookPath("subscription-manager"); err != nil {
+		t.Skip("no subscription-manager found")
+		return func() {}
+	}
+
+	matches, err := filepath.Glob("/etc/pki/entitlement/*.pem")
+	if err == nil && len(matches) > 0 {
```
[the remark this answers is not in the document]
mvo5: Yeah, it's basically just double checking if the machine is subscribed already (because the helper unsubscribes machines that got subscribed just for the test again), actually there should be a comment to explain this and a "XXX: find a better way", I had trouble finding some machine readable API and stuff like "subscription-manager status"/"list" is both slow and hard to parse. But maybe I'm missing something :/

bib/internal/container/container.go
mvo5: A good question, I think we need to have a brainstorm about it -  one way would be to just make it a single file script again another to find/fix the root cause why "dnf.SetRootdir()" does not work with subscribed content.

achilleas-k: > find/fix the root cause why "dnf.SetRootdir()" does not work with subscribed content

Need me to look into this?  It sounds like we have a bug here in the depsolver.

mvo5: @achilleas-k Sure thing, feel free to look at this, I'm very curious. The easiest is to start from a subscribed machine, then revert [3540536](https://github.com/osbuild/bootc-image-builder/pull/670/commits/3540536d8d7b1a3d879b7996532fd6716f410e4a) and run (as root) `cd bib/internal/cntdnf && go test -v`

.github/workflows/testingfarm-unit.yml
```diff
@@ -0,0 +1,43 @@
+---
+name: Testing farm go unit tests
```
mvo5, line 2: This duplication is sad, we should look into something like https://github.com/actions/starter-workflows/issues/245#issuecomment-1445128447 here too.

bib/internal/cntdnf/cntdnf.go
```diff
@@ -0,0 +1,48 @@
+package cntdnf
```
mvo5, line 1: Ondrej suggest to put this container package as containerSolver/dnfsolver and then have exec on container

mvo5: This is done now.

mvo5: cc @ondrejbudai

ondrejbudai (approved):
Looks good. @achilleas-k wanna take one last look? :)

bib/internal/container/container.go
achilleas-k: WARNING: My brain's a bit of a jumble right now because of all the layers, so I might start overexplaining things just to make sure I'm not glossing over details.

The problem with the `root_dir` feature is that when the repos are initialised, the cert/key paths are set to `/etc/pki/entitlement-host/...` (and `/etc/rhsm-host/...`) which is a symlink to `/run/secrets/etc-pki-entitlement/...` (and `/run/secrets/rhsm/...`), but those paths aren't available at the mounted location of the container.

When the `root_dir` is set, key and cert paths in the repo configs get the root directory prepended to them.  This makes sense in the general case: if a repository cert path points to `/etc/pki/entitlement/whatever` and its root is at `/mnt`, you will probably find that cert at `/mnt/etc/pki/entitlement/whatever`.  With containers, and especially with what we are doing with containers, things get _weird_.  It turns out that if we disable the prepending of the `root_dir` to all the paths, things work!

There's 3 layers here: (1) the host, (2) the BIB container, and (3) the bootc (base) container.  There's also, however, the mounted bootc container, which we'll call (3m).

When we copy `osbuild-depsolve-dnf` into the container and do the depsolve there (the scenario we had working all along), everything exists and runs in (3) and everyone's happy.  Secrets get mounted from the host (1) into the BIB container (2) by podman automatically, and from the BIB container (2) into the bootc container (3) explicitly by [this bit here](https://github.com/osbuild/bootc-image-builder/blob/cfb33fd6847df1dd60628fe1977db285d99f72ea/bib/internal/container/container.go#L41-L44).

When we run the depsolve using `root_dir`, we point it to a `podman mount`ed location (3m), which **does not** have a populated `/run/secrets`.  Presumably this happens because the actual purpose of `podman mount` is to provide access to the container's root filesystem, and not every arbitrary mountpoint that's visible from inside the container itself.  So the path `/var/lib/containers/storage/overlay/<ID>/merged/etc/pki/entitlement-host` is a link to `/var/lib/containers/storage/overlay/<ID>/merged/run/secrets` which is empty (or at least, doesn't contain the secrets mounted from (2)).  However, since we're running `osbuild-depsolve-dnf` from inside the BIB container (2), skipping the `root_dir` prepend step makes the repository configs from (3m) point to secrets that actually exist in (2).

achilleas-k: Short version: When we `podman mount` the bootc container, its `/run/secrets` directory has nothing in it so the `root_dir` isn't complete.

achilleas-k: I had a couple of thoughts about how to solve this, if we really want to use the `root_dir` feature and run the depsolver from the BIB layer.
1. Make the path translation optional: bad idea, very specific to this case and probably will have issues with other paths.
2. Mount `/run/secrets` to `/var/lib/containers/storage/overlay/<ID>/merged/run/secrets` (from (2) into (3m)).

cgwalters: > Short version: When we podman mount the bootc container, its /run/secrets directory has nothing in it so the root_dir doesn't work because the filesystem tree isn't "complete".

This is exactly the kind of thing that gets fixed though if more of what bib does is "podman run" right?

Yes, `podman mount` is giving access to just the raw merged root filesystem, it doesn't e.g. mount `/proc` in that root either.

achilleas-k: > This is exactly the kind of thing that gets fixed though if more of what bib does is "podman run" right?

Yes but we have an issue here with the depsolver not being part of the base container but needing to use repo configs from the container.  So far we've had two solutions:
1. Copy the depsolver into the container and `podman run` it.  In detail this involves:
    - podman run container
    - podman mount container
    - podman copy depsolver container
    - podman run depsolver
2. Run the depsolver outside the container with configs from inside the container.
    - podman run container
    - podman mount container
    - depsolve with `root_dir` set to container mountpoint

Subscriptions make both of these a bit more complicated because we need to run _something_ in the container to initialise the repository configurations, otherwise `/etc/yum.repos.d` could be empty.

Solution 1 is simple enough if the depsolver is a single script **and** we can assume that all the libraries it imports are available in the container (which we can, since it only really imports libdnf and its dependencies).  It got a tiny bit more complicated when we split `osbuild-depsolve-dnf` into a module to reduce duplication between the dnf4 and dnf5 implementations.

Solution 2 needs more fiddling but doesn't rely on running scripts in an unknown environment that's coming from the user.

At the end of the day though, there's no simple `podman run` here unless every bootc container contains a depsolver that can talk to BIB and it can guarantee that it will work with subscription-manager managing repositories.

I guess another long term solution is to never rely on downloading packages for any build, but that's not possible for the Anaconda ISO.

achilleas-k:
Should we merge this now to fix the immediate issue and then follow up with a switch back to using `root_dir` with the `/run/secrets` mount?

achilleas-k (approved):
Good to merge from my end.  Feel free to resolve the `root_dir` conversation if you think it's done for now.  We can pick it up again elsewhere.

bib/internal/container/container.go
mvo5: I guess the trade-off we need to pick here is:
1. the ugliness of `injectDNFJson()` which got worse with the "no-longer-a-single-file" change (which we *could* revisit)
2. the ugliness of detecting/mounting /run/secrets and different behaviors if this runs directly from a host (where there are no /run/secrets) or nested from a container (where there is)

I have no good answer here neither seems nice :/

achilleas-k: Agreed.

Resolving this thread to merge and we can continue the discussion elsewhere.

Issue #528, opened 2024-07-05
kickstart users and/or groups are not compatible with user-supplied kickstart content when build iso after # 438 mereged


chunfuwen:
Steps :
1) After https://github.com/osbuild/bootc-image-builder/pull/438/files merged, build out one local boot-image-builder
2) Prepare one config.json with kickstart contents
```
cat config.json 
{"blueprint": {"customizations": {"user": [{"name": "alice", "password": "bob", "groups": ["wheel"], "key": "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABgQCa26Mh2J+PAxTVSxnBqqYNTsjwJZf7C2b/xzU6VMj5jUb2FOK6fjXI7qXc9CjHINIh++zHEe1/Z4KRHfcnfzdAsfPGUw/stSQEb+9sbziyRJWY/DvDEW5NKQPuBGDY6Pu20aJLLpS8e1Ed2QyPQXxr5eh81gbVACV3VJ+fVTaWS/0W4/CxeJk73o2kNPiqoT6Hs/h3Shs9OpbPiSOayjEoKJ194uYGO5hPF/N4sWt4eH1UQk/ZmWSWBSo1hVV1QIasWHlwjviJFrgcEof6Spng/WFXIbJzIzgnPIUjJ6NZmjvkXeop4IJ1FUQa+nqru+wd5RwE9OaVgUsmBSVDkL52nb6wy7htO+JZmWdXS1xjcHdU/5lHvmSIT6sfRXesNeqpd6HpTphtOnRN2uQ2vsIBYZn8EYx9Xf1G+nAu9V+OyxQCZM/ZrNV6Dwc9+Xda8QVskbzQtBPuGEaoEtB2jkWntjsvibcD0G3RCk= root@dxx.pek2.redhat.com"}], "kernel": {"append": "mitigations=auto,nosmt"}, "installer": {"kickstart": {"contents": "text --non-interactive\nzerombr\nclearpart --all --initlabel --disklabel=gpt\nautopart --noswap --type=lvm\nnetwork --bootproto=dhcp --device=link --activate --onboot=on"}}}}}
```

3) Execute boot-image-builder command as below
```
[...]

PR #671, opened 2024-10-13
Update README to add clarity regarding kickstart customizations

estuart:
Added language to point out that other customization blocks are not supported when using a kickstart customization block as pointed out in https://github.com/osbuild/bootc-image-builder/issues/528


mvo5 (approved):
This clarification looks fine to me.

PR #675, opened 2024-10-14
buildconfig: support passing build config via `stdin`

mvo5:
This commit adds support for reading the build config via stdin. To do that the `-config` switch now supports `-` and with that it will use `os.Stdin` insteadof readin the file. Because we have no file extension to auto-detect the format only JSON is supported in this mode.

This feature will be used by osbuild-composer when it adds bootc image mode integration.


ondrejbudai (approved):
Thanks! :heart:

PR #676, opened 2024-10-14
fix: use remove-signatures in skopeo copy stages

mmartinv:
According to [1] neither Docker daemon nor OCI support storing
the container signatures, and skopeo copy refuses to just silently drop
them and fails.

When using `skopeo copy --remove-signatures` to make the copy, the
signatures will be still read and policy.json will still be enforced,
they just won’t be written into the destination.

[1] https://github.com/containers/skopeo/issues/589#issuecomment-453894401


cgwalters:
xref https://github.com/containers/bootc/issues/812 and https://github.com/containers/image/pull/2590

mvo5 (approved):
Thank you

PR #680, opened 2024-10-17
test: fix the testcase strings to not include `osinfo_template`

mvo5:
This commit fixes the testcase string representation to not include the `osinfo_template` string and adds a test case to not accidentally regress here. Currently our test case strings look like:
```
<Dir bootc-image-builder>
  <Dir test>
    <Module test_build.py>
      <Function test_container_builds>
      <Function test_image_is_generated[quay.io/centos-bootc/centos-bootc:stream9,qcow2+raw+vmdk+vhd+gce,CentOS Stream 9 ({arch})]>
      <Function test_image_boots[quay.io/centos-bootc/centos-bootc:stream9,raw,CentOS Stream 9 ({arch})]>
...
```
which is an accident (not the {arch} part in the output above). A testcase should ideally only contain what

Instead of putting the osinfo_template into the testcase itself, have a function that generates it. It has downsides (now the osinfo_template is more disconnected from the testcase) so we could move it also back into testcase or we could exclude osinfo_template from the string generation. Ideas welcome here, my current approach feels okay but not like it's perfect yet.

### DIFF
--- a/test/test_build.py
+++ b/test/test_build.py
@@ -35,7 +35,6 @@ class ImageBuildResult(NamedTuple):
     img_type: str
     img_path: str
     img_arch: str
-    osinfo_template: str
     container_ref: str
     rootfs: str
     username: str
@@ -163,7 +162,7 @@ def build_images(shared_tmpdir, build_container, request, force_aws_upload):
             journal_output = journal_log_path.read_text(encoding="utf8")
             bib_output = bib_output_path.read_text(encoding="utf8")
             results.append(ImageBuildResult(
-                image_type, generated_img, tc.target_arch, tc.osinfo_template,
+                image_type, generated_img, tc.target_arch,
                 tc.container_ref, tc.rootfs, username, password,
                 ssh_keyfile_private_path, kargs, bib_output, journal_output))
 
@@ -298,7 +297,7 @@ def build_images(shared_tmpdir, build_container, request, force_aws_upload):
     results = []
     for image_type in image_types:
         results.append(ImageBuildResult(
-            image_type, artifact[image_type], tc.target_arch, tc.osinfo_template,
+            image_type, artifact[image_type], tc.target_arch,
             tc.container_ref, tc.rootfs, username, password,
             ssh_keyfile_private_path, kargs, bib_output, journal_output, metadata))
     yield results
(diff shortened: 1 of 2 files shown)


test/test_build.py
```diff
@@ -453,20 +452,28 @@ def test_iso_installs(image_type):
...
     if not arch:
         arch = platform.machine()
-    osinfo = image_type.osinfo_template.format(arch=arch)
     result = subprocess.run([
         "osinfo-detect",
         installer_iso_path,
     ], capture_output=True, text=True, check=True)
     osinfo_output = result.stdout
-    expected_output = f"Media is bootable.\nMedia is an installer for OS '{osinfo}'\n"
+    expected_output = f"Media is bootable.\nMedia is an installer for OS '{osinfo_for(image_type, arch)}'\n"
```
mmartinv, line 476:
```suggestion
    expected_output = osinfo_for(image_type, arch)
```

test/test_build.py
```diff
@@ -453,20 +452,28 @@ def test_iso_installs(image_type):
...
 
+def osinfo_for(it: ImageBuildResult, arch: str) -> str:
+    if it.container_ref == "quay.io/centos-bootc/centos-bootc:stream9":
+        return f"CentOS Stream 9 ({arch})"
+    if it.container_ref.startswith("quay.io/fedora/fedora-bootc:"):
+        ver = it.container_ref.split(":", maxsplit=2)[1]
+        return f"Fedora Server {ver} ({arch})"
+    raise ValueError(f"cannot find osinfo_template for '{it.container_ref}'")
+
+
```
mmartinv, line 463:
```suggestion
def osinfo_for(it: ImageBuildResult, arch: str) -> str:
    osinfo =  "Media is bootable.\nMedia is an installer for OS '{os}'\n"
    if it.container_ref == "quay.io/centos-bootc/centos-bootc:stream9":
        os = f"CentOS Stream 9 ({arch})"
    if it.container_ref.startswith("quay.io/fedora/fedora-bootc:"):
        ver = it.container_ref.split(":", maxsplit=2)[1]
        os = f"Fedora Server {ver} ({arch})"
    if os:
        return osinfo.format(os=os)"
    raise ValueError(f"cannot find osinfo_template for '{it.container_ref}'")
```

mmartinv (commented):
Why just not moving all the osinfo information to the `osinfo_for` function?

mvo5:
> Why just not moving all the osinfo information to the `osinfo_for` function?

Thank you, that is an excellent suggestion, I updated the code based on your suggestion (will probably have to force push again and if you don't mind squash) because I think the linter will not be happy.

mmartinv:
@mvo5 why do you removed my suggestion? wasn't it working?

mvo5:
> @mvo5 why do you removed my suggestion? wasn't it working?

@mmartinv I'm deep sorry, I'm a Muppet, I *thought* I applied the suggestion and pushed but clearly I did not, I am fixing this now, thanks for noticing.

ondrejbudai (dismissed):
Awesome, thanks!

achilleas-k (dismissed):

mvo5:
Sorry, had to re-push as one of the new tests broke this PR.

achilleas-k (approved):

PR #687, opened 2024-10-22
test: add anaconda-iso build tests with signed containers

mmartinv:
Add anaconda-iso iso build tests with signed containers.
The rest of the images can be also added to the test once
[1] and [2] are merged

[1] https://github.com/osbuild/images/pull/990
[2] https://github.com/osbuild/osbuild/pull/1906

Signed-off-by: Miguel Martín <mmartinv@redhat.com>

### DIFF
--- a/test/test_build.py
+++ b/test/test_build.py
@@ -11,6 +11,7 @@ import tempfile
 import uuid
 from contextlib import contextmanager
 from typing import NamedTuple
+from dataclasses import dataclass
 
 import pytest
 # local test utils
@@ -47,14 +48,169 @@ class ImageBuildResult(NamedTuple):
     metadata: dict = {}
 
 
+@dataclass
+class GPGConf:
+    email: str
+    key_length: str
+    home_dir: str
+    pub_key_file: str
+    key_params: str
+
+
+@dataclass
+class RegistryConf:
+    local_registry: str
+    sigstore_dir: str
+    registries_d_dir: str
+    policy_file: str
+    lookaside_conf_file: str
+    lookaside_conf: str
+
+
 @pytest.fixture(name="shared_tmpdir", scope='session')
 def shared_tmpdir_fixture(tmpdir_factory):
     tmp_path = pathlib.Path(tmpdir_factory.mktemp("shared"))
     yield tmp_path
 
 
+@pytest.fixture(name="gpg_conf", scope='session')
+def gpg_conf_fixture(shared_tmpdir):
+    key_params_tmpl = """
+      %no-protection
+      Key-Type: RSA
+      Key-Length: {key_length}
+      Key-Usage: sign
+      Name-Real: Bootc Image Builder Tests
+      Name-Email: {email}
+      Expire-Date: 0
+    """
+    email = "bib-tests@redhat.com"
+    key_length = "3072"
+    home_dir = f"{shared_tmpdir}/.gnupg"
+    pub_key_file = f"{shared_tmpdir}/GPG-KEY-bib-tests"
+    key_params = key_params_tmpl.format(key_length=key_length, email=email)
+
+    os.makedirs(home_dir, mode=0o700, exist_ok=False)
+    subprocess.run(
+        ["gpg", "--gen-key", "--batch"],
+        check=True, env={"GNUPGHOME": home_dir},
+        input=key_params,
+        text=True)
+    subprocess.run(
+        ["gpg", "--output", pub_key_file,
+         "--armor", "--export", email],
+        check=True, env={"GNUPGHOME": home_dir})
+
+    yield GPGConf(email=email, home_dir=home_dir,
+                  key_length=key_length, pub_key_file=pub_key_file, key_params=key_params)
+
+
+@pytest.fixture(name="registry_conf", scope='session')
+def registry_conf_fixture(shared_tmpdir, request):
+    lookaside_conf_tmpl = """
+    docker:
+      {local_registry}:
+        lookaside: file:///{sigstore_dir}
+    """
+    registry_port = testutil.get_free_port()
+    # We cannot use localhost as we need to access the registry from both
+    # the host system and the bootc-image-builder container.
+    default_ip = testutil.get_ip_from_default_route()
+    local_registry = f"{default_ip}:{registry_port}"
+    sigstore_dir = f"{shared_tmpdir}/sigstore"
+    registries_d_dir = f"{shared_tmpdir}/registries.d"
+    policy_file = f"{shared_tmpdir}/policy.json"
+    lookaside_conf_file = f"{registries_d_dir}/lookaside.yaml"
+    lookaside_conf = lookaside_conf_tmpl.format(
+        local_registry=local_registry,
+        sigstore_dir=sigstore_dir
+    )
+    os.makedirs(registries_d_dir, mode=0o700, exist_ok=True)
+    os.makedirs(sigstore_dir, mode=0o700, exist_ok=True)
+
+    registry_container_name = f"registry_{registry_port}"
+
+    registry_container_running = subprocess.run([
+        "podman", "ps", "-a", "--filter", f"name={registry_container_name}", "--format", "{{.Names}}"
+    ], check=True, capture_output=True, text=True).stdout.strip()
+    if registry_container_running != registry_container_name:
+        subprocess.run([
+            "podman", "run", "-d",
+            "-p", f"{registry_port}:5000",
+            "--restart", "always",
+            "--name", registry_container_name,
+            "registry:2"
+        ], check=True)
+
+    registry_container_state = subprocess.run([
+        "podman", "ps", "-a", "--filter", f"name={registry_container_name}", "--format", "{{.State}}"
+    ], check=True, capture_output=True, text=True).stdout.strip()
+
+    if registry_container_state in ("paused", "exited"):
+        subprocess.run([
+            "podman", "start", registry_container_name
+        ], check=True)
+
+    def remove_registry():
+        subprocess.run([
+            "podman", "rm", "--force", registry_container_name
+        ], check=True)
+
+    request.addfinalizer(remove_registry)
+    yield RegistryConf(
+        local_registry=local_registry,
+        sigstore_dir=sigstore_dir,
+        registries_d_dir=registries_d_dir,
+        policy_file=policy_file,
+        lookaside_conf=lookaside_conf,
+        lookaside_conf_file=lookaside_conf_file,
+    )
+
+
+def get_signed_container_ref(local_registry: str, container_ref: str):
+    container_ref_path = container_ref[container_ref.index('/'):]
+    return f"{local_registry}{container_ref_path}"
+
+
+def sign_container_image(gpg_conf: GPGConf, registry_conf: RegistryConf, container_ref):
+    registry_policy = {
+        "default": [{"type": "insecureAcceptAnything"}],
+        "transports": {
+            "docker": {
+                f"{registry_conf.local_registry}": [
+                    {
+                        "type": "signedBy",
+                        "keyType": "GPGKeys",
+                        "keyPath": f"{gpg_conf.pub_key_file}"
+                    }
+                ]
+            },
+            "docker-daemon": {
+                "": [{"type": "insecureAcceptAnything"}]
+            }
+        }
+    }
+    with open(registry_conf.policy_file, mode="w", encoding="utf-8") as f:
+        f.write(json.dumps(registry_policy))
+
+    with open(registry_conf.lookaside_conf_file, mode="w", encoding="utf-8") as f:
+        f.write(registry_conf.lookaside_conf)
+
+    signed_container_ref = get_signed_container_ref(registry_conf.local_registry, container_ref)
+    cmd = [
+        "skopeo", "--registries.d", registry_conf.registries_d_dir,
+        "copy", "--dest-tls-verify=false", "--remove-signatures",
+        "--sign-by", gpg_conf.email,
+        f"docker://{container_ref}",
+        f"docker://{signed_container_ref}",
+    ]
+    print(cmd)
+    subprocess.run(cmd, check=True, env={"GNUPGHOME": gpg_conf.home_dir})
+
+
 @pytest.fixture(name="image_type", scope="session")
-def image_type_fixture(shared_tmpdir, build_container, request, force_aws_upload):
+# pylint: disable=too-many-arguments
+def image_type_fixture(shared_tmpdir, build_container, request, force_aws_upload, gpg_conf, registry_conf):
     """
     Build an image inside the passed build_container and return an
     ImageBuildResult with the resulting image path and user/password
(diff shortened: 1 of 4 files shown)


chunfuwen:
From user perspective, can we assume bib can now build out one anaconda-iso with signed bootc container image ?

mmartinv:
> From user perspective, can we assume bib can now build out one anaconda-iso with signed bootc container image ?

Yes, since #676 was merged

test/testcases.py
```diff
@@ -23,6 +23,12 @@ class TestCase:
     # rootfs to use (e.g. ext4), some containers like fedora do not
     # have a default rootfs. If unset the container default is used.
     rootfs: str = ""
+    # osinfo_template is a string template describing the OS detected by
+    # 'osinfo-detect'. It can contain '{arch}' that will be replaced with the
+    # actual container image arch
+    osinfo_template: str = ""
+    # Sign the container_ref and use the new signed image instead of the original one
+    sign: bool = False
```
mvo5, line 31: This will conflict with https://github.com/osbuild/bootc-image-builder/pull/680 (see details why osinfo_template is not ideal as part of the testcase class) - having sign here is good though.

test/test_build.py
```diff
@@ -113,11 +113,20 @@ def build_images(shared_tmpdir, build_container, request, force_aws_upload):
     password = "password"
     kargs = "systemd.journald.forward_to_console=1"
 
+    default_ip = testutil.get_ip_from_default_route()
```
mvo5, line 116: Silly question - isn't `localhost` here good enough? And if not, maybe a short comment why not?

test/testutil.py
```diff
@@ -147,3 +148,102 @@ def create_filesystem_customizations(rootfs: str):
...
+        "route",
+        "list",
+        "default"
+    ], check=True, capture_output=True).stdout
+    return default_route.split()[8].decode("utf-8")
+
+
+@dataclasses.dataclass
+class GPGConfig():
+    email: str = "bootc-image-builder@redhat.com"
```
mvo5, line 165: (nitpick) maybe indicate in this string (and the one below) something like "testing" (it's not really relevant as this is a perfectly fine config that could be used outsidde of testing too but I like making things like this clear)

test/testutil.py
```diff
@@ -147,3 +148,102 @@ def create_filesystem_customizations(rootfs: str):
...
+    pub_key: str = "/etc/pki/rpm-gpg/RPM-GPG-KEY-booc-image-builder"
```
mvo5, line 166: Hm, this will create a key in the real root filesystem, is this intentional? I.e. if I run "sudo pytest" this will modify my real root? Is there a different way maybe?

test/testutil.py
```diff
@@ -147,3 +148,102 @@ def create_filesystem_customizations(rootfs: str):
...
+      Name-Email: {email}
+      Expire-Date: 0
+    """
+
+
+@dataclasses.dataclass
+class RegistryConfig():
+    local_registry: str = "localhost:5000"
+    sigstore_dir: str = "/var/lib/containers/sigstore"
+    lookaside_config: str = "/etc/containers/registries.d/bib.yaml"
```
mvo5, line 183: This will also use the real root, is there a way to avoid this?

test/testutil.py
```diff
@@ -147,3 +148,102 @@ def create_filesystem_customizations(rootfs: str):
...
+        "default"
+    ], check=True, capture_output=True).stdout
+    return default_route.split()[8].decode("utf-8")
+
+
+@dataclasses.dataclass
+class GPGConfig():
+    email: str = "bootc-image-builder@redhat.com"
+    pub_key: str = "/etc/pki/rpm-gpg/RPM-GPG-KEY-booc-image-builder"
+    key_length = "3072"
```
mvo5, line 167: (not to self): key generation is really fast nowdays, no need to use a shorter key here to speed things up

test/testcases.py
```diff
@@ -68,7 +74,11 @@ def gen_testcases(what):  # pylint: disable=too-many-return-statements
     if what == "ami-boot":
         return [TestCaseCentos(image="ami"), TestCaseFedora(image="ami")]
     if what == "anaconda-iso":
-        return [TestCaseCentos(image="anaconda-iso"), TestCaseFedora(image="anaconda-iso")]
+        return [
+            klass(image="anaconda-iso", sign=sign, local=False)
+            for klass in (TestCaseCentos, TestCaseFedora)
+            for sign in [True, False]
```
mvo5, line 80: The iso tests are already pretty slow because of the depsolving/rpms involved. While this is a perfect test in the sense that it replicates exactly the issue we had and therefore validates the fix I do wonder if we could get away with a "cheaper" test. Maybe not but I think it's worth thinking about it .)

mvo5 (commented):
Thank you for this very nice test!

Some quick notes inline and *maybe* we can get away without skopeo copy in bib if we actually implement https://github.com/osbuild/osbuild/pull/1907 - having the test is still good though.

I like the thoroughness of the test but it's expensive (timewise in CI) so I wonder if we can find a faster way (see my inline comment). Maybe not, but worth thinking about I feel.

test/test_build.py
mmartinv: No, as we need to copy the image from the host and from within the bib container

test/testcases.py
mmartinv: Honestly, I am not sure we can test this without building the anaconda-iso image and the same will apply to the other disk images in the future. Maybe we can limit the test to only a subset of them: build just an ISO and one disk image for only one OS. This will test both paths (the one taken by anaconda-iso by calling directly `org.osbuild.skopeo stage` and the disk image path by calling `org.osbuild.container-deploy`). The next obvious question is: which OS should we choose?

test/testutil.py
mmartinv: I moved everything to a temporary directory except the lookaside config. We need that in the host system so the sigstore is created by skopeo.

test/testutil.py
mmartinv: I changed this so the file is removed once skopeo has created the sigstore

test/testcases.py
mmartinv: I have limited the tests to Fedora 40 ISO only

test/testutil.py
```diff
@@ -147,3 +149,270 @@ def create_filesystem_customizations(rootfs: str):
...
+    signed_container_ref = get_signed_container_ref(local_registry, container_ref)
+    system_lookaside_conf_file = os.path.join(
+        "/etc/containers/registries.d",
+        os.path.basename(lookaside_conf_file)
+    )
+    # We need to temporarily configure system's lookaside for skopeo to
+    # create the sigstore dir.
+    shutil.copy(lookaside_conf_file, system_lookaside_conf_file)
+    subprocess.run([
+        "skopeo", "copy",
```
chunfuwen, line 411: Offline I tried to use skopeo copy to sign the container image, but it looks like it is time consuming since it also need copy container image.  I try this way as per https://github.com/containers/podman/blob/main/docs/tutorials/image_signing.md. Here firstly retag image to localhost, then use sudo -E GNUPGHOME=$HOME/.gnupg \
    podman push \
    --tls-verify=false \
    --sign-by sgrunert@suse.com \
    localhost:5000/alpine.
It is straightforward, and much time saving

mmartinv: I am not sure I understand, doing it with podman would be exactly the same as doing it with skopeo. Doing it with podman requires multiple commands: first pull, then tag and then sign+push the container to the new registry. Skopeo does all of that in one single command and, I haven't tested, but it should take moreless the same amount of time. Are you saying that there is a big difference? How much?

mmartinv:
> In order to validate feature related to this, assuming I had one signed container image,e.g:registry.redhat.io/rhel9/rhel-bootc:9.4. So do I have to execute sudo skopeo copy --remove-signatures registry.redhat.io/rhel9/rhel-bootc:9.4 registry.redhat.io/rhel9/rhel-bootc:9.4 before I call sudo podman run --rm -i --privileged --pull=newer --security-opt label=type:unconfined_t -v /var/lib/containers/storage:/var/lib/containers/storage quay.io/centos-bootc/bootc-image-builder:latest --type anaconda-iso --local registry.redhat.io/rhel9/rhel-bootc:9.4 ?

No, you don't need to copy anything, using the signed container directly should work as is. Not sure if the change is already available in quay.io/centos-bootc/bootc-image-builder:latest though.

test/testutil.py
chunfuwen: My tried actions follow as:
1)firstly pull image to local host
2)podman tag image to local registry
3)podman sign local image and push back to local registry.
Since all operations almost happen on container on local registry, it happen very fast. Since it aim to  testing purpose only, so i don't need to push back to remote registry,such as quay.io. 

From your analysis, it looks like the process is almost same.
Maybe I missed something in last attempt using skopeo copy

test/test_build.py
```diff
@@ -113,11 +113,20 @@ def build_images(shared_tmpdir, build_container, request, force_aws_upload):
     password = "password"
     kargs = "systemd.journald.forward_to_console=1"
 
+    container_ref = tc.container_ref
+
+    # We cannot use localhost as we need to access the registry from both
```
mvo5, line 118: Thanks for this comment!

test/test_build.py
```diff
@@ -113,11 +113,20 @@ def build_images(shared_tmpdir, build_container, request, force_aws_upload):
...
+    # the host system and the bootc-image-builder container.
+    default_ip = testutil.get_ip_from_default_route()
+    local_registry = f"{default_ip}:5000"
```
mvo5, line 121: (nitpick) we have a "testutil:get_free_port()" helper (which is sadly inherently racy) I think would be slightly nicer in case 5000 is already taken - or will that break things if we don't use 5000 (it seems we pass the port everywhere?)

test/testutil.py
```diff
@@ -147,3 +149,270 @@ def create_filesystem_customizations(rootfs: str):
...
+    _key_params_tmpl: str = """
+      %no-protection
+      Key-Type: RSA
+      Key-Length: {key_length}
+      Key-Usage: sign
+      Name-Real: Bootc Image Builder Tests
+      Name-Email: {email}
+      Expire-Date: 0
+    """
+    _base_dir: str = "/tmp/bib-tests"
```
mvo5, line 175: Hm, my gut feeling is that we want something like `_base_dir: str = tempfile.mkdtemp(prefix="bib-tests")` here or maybe a small comment why we use a fixed dirname in tmp?

test/testutil.py
```diff
@@ -147,3 +149,270 @@ def create_filesystem_customizations(rootfs: str):
...
+
+def gpg_gen_key(gpg_conf: GPGConf):
+    if os.path.exists(gpg_conf.home_dir):
+        return
+
+    os.makedirs(gpg_conf.home_dir, mode=0o700, exist_ok=False)
+
+    subprocess.run(
+        ["gpg", "--gen-key", "--batch"],
+        check=True, capture_output=True,
```
mvo5, line 249: I wonder if we want `capture_output=True` here. By default pytest will hide the output but if anything goes wrong this catpure_output=True will mean the errors are not visible by default in the exception (subprocess.CalledProcessError). Maybe just not capture here and below (unless I miss something?)

test/testutil.py
```diff
@@ -147,3 +149,270 @@ def create_filesystem_customizations(rootfs: str):
...
+
+@dataclasses.dataclass
+class RegistryConf():
+    _lookaside_conf_tmpl: str = """
+    docker:
+      {local_registry}:
+        lookaside: file:///{sigstore_dir}
+    """
+    _local_registry: str = "localhost:5000"
+    _base_dir: str = "/tmp/bib-tests"
```
mvo5, line 269: Same question as above if we should maybe use `tempfile.mkdtemp()` here (or a comment why we don't)

test/testutil.py
```diff
@@ -147,3 +149,270 @@ def create_filesystem_customizations(rootfs: str):
...
 ]
+
+
+def get_ip_from_default_route():
+    default_route = subprocess.run([
+        "ip",
+        "route",
+        "list",
+        "default"
+    ], check=True, capture_output=True).stdout
```
mvo5, line 160: (nitpick) when using `text=True` here the `decode("utf-8")` should not be necessary (as the output is captured as text already)

test/testutil.py
```diff
@@ -147,3 +149,270 @@ def create_filesystem_customizations(rootfs: str):
...
+
+    @lookaside_conf.setter
+    def lookaside_conf(self, conf: str) -> None:
+        self._lookaside_conf = conf
+
+
+def ensure_registry():
+    registry_container_name = subprocess.run([
+        "podman", "ps", "-a", "--filter", "name=registry", "--format", "{{.Names}}"
+    ], check=True, capture_output=True).stdout.decode("utf-8").strip()
```
mvo5, line 345: (niptick) (we could use `text=True` here and below too I think)

test/testutil.py
```diff
@@ -147,3 +149,270 @@ def create_filesystem_customizations(rootfs: str):
...
+        return self._base_dir
+
+    @base_dir.setter
+    def base_dir(self, base_dir: str) -> None:
+        self._base_dir = base_dir
+        self._home_dir = f"{base_dir}/.gnupg"
+        self._pub_key_file = f"{base_dir}/GPG-KEY-bib-tests"
+
+    @property
+    def home_dir(self) -> str:
```
mvo5, line 193: Sorry, two maybe silly questions:
1. should we be smart about the base dir here too, i.e. just like above build the home dir relative to the base dir?
2. if not, would it make sense to just avoid the getter/setter here and use the dataclass directly (i.e. change _home_dir to home_dir) because we don't do anything with the `home_dir` in the getter/setter (unlike the base_dir above)

test/testutil.py
```diff
@@ -147,3 +149,270 @@ def create_filesystem_customizations(rootfs: str):
...
+    # create the sigstore dir.
+    shutil.copy(lookaside_conf_file, system_lookaside_conf_file)
+    subprocess.run([
+        "skopeo", "copy",
+        "--dest-tls-verify=false",
+        "--remove-signatures",
+        "--sign-by", gpg_conf.email,
+        f"docker://{container_ref}",
+        f"docker://{signed_container_ref}",
+    ], check=True, capture_output=True, env={"GNUPGHOME": gpg_conf.home_dir})
```
mvo5, line 417: Same question about `capture_output=True` as above about it potentially eating useful debug output (but feel free to just ignore if the answer is the same as for the other question above :)

test/testutil.py
```diff
@@ -147,3 +149,270 @@ def create_filesystem_customizations(rootfs: str):
...
+
+    with open(lookaside_conf_file, mode="w", encoding="utf-8") as f:
+        f.write(lookaside_conf)
+
+    signed_container_ref = get_signed_container_ref(local_registry, container_ref)
+    system_lookaside_conf_file = os.path.join(
+        "/etc/containers/registries.d",
+        os.path.basename(lookaside_conf_file)
+    )
+    # We need to temporarily configure system's lookaside for skopeo to
```
mvo5, line 407: Thanks for this comment, I do feel uneasy about modifying the host system that run the tests. So far bib tests are pretty clean and I don't think we do this. It also raises questions about running tests in parallel (pytest -n) which we will want to do in the future to speed up things. Is there any way this could be done differently?

mvo5 (commented):
Thank you for the branch and the update. I added a bunch of small questions/suggestions (hope they are useful). The one bigger point that prevents me from approving is that it seems to modify the host system configuration which I would really love to avoid. If there is no way to avoid it, maybe we can add a pytest condition that skips the test unless e.g. an env like `BIB_TEST_MODIFY_SYSTEM=1` is set (or a parameter to pytest with a similar name9 and then we set this for the GH action/testing farm tests but not for the default, wdyt?

test/testutil.py
mmartinv: It looks like skopeo supports `--registries.d` global option, I didn't realize earlier because I was just looking at `skopeo copy` command options, I will update the code to save everything within the base dir. This way nothing in the host system will be modified/changed

test/test_build.py
mmartinv: The port number itself is not important but the idea is to share the same registry to save all the signed images locally. I will add a global `registry_port` variable, not sure if that's the best way to do it though

test/testutil.py
mmartinv: `base_dir`  is a simple string and is the default value (chosen from other parts in the code) but notice that the actual directory creation is not handled by the class at any point.
In this case we just use the `output_dir` for each build when `RegistryConf` and `GPGConf` classes are instantiated (see lines 272 and 274 in the `tests_build.py` file). I mean, the actual value of the default `_base_dir` does not really matter as it will not be actually used.
If I use your suggestion an additional temporary directory will be created for each build that will not be actually used.

test/testutil.py
mmartinv: Yeah, you are right, we can use the getter/setter from the dataclass so the `home_dir` can be set independently from `base_dir

test/test_build.py
achilleas-k: I think the "correct" way would be a pytest fixture with the scope set to session.

test/testutil.py
achilleas-k: Perhaps it's better than to instead have no default values and error out if any of the getters are used before setting the base?

mmartinv: Ok, tried a different/simpler approach, please take a look

test/test_build.py
mmartinv: Used the `get_free_port()` function as requested

test/testutil.py
mmartinv: Resolving as this no longer applies

mvo5 (approved):
Thank you! This is very nice now.

mvo5 (approved):

PR #696, opened 2024-10-28
Drop removed packages from CentOS Stream 10 and RHEL 10.

achilleas-k:
**data/defs/centos-10: drop removed packages**

These packages are no longer part of CentOS Stream 10 (or RHEL 10).

---

**test: run tests with centos-bootc:stream10 as well**

Rename TestCaseCentos to TestCaseS9S and add a TestCaseC10S.

---

### DIFF
--- a/test/testcases.py
+++ b/test/testcases.py
@@ -60,6 +60,13 @@ class TestCaseC9S(TestCase):
         "quay.io/centos-bootc/centos-bootc:stream9")
 
 
+@dataclasses.dataclass
+class TestCaseC10S(TestCase):
+    container_ref: str = os.getenv(
+        "BIB_TEST_BOOTC_CONTAINER_TAG",
+        "quay.io/centos-bootc/centos-bootc:stream10")
+
+
 def test_testcase_nameing():
     """
     Ensure the testcase naming does not change without us knowing as those
@@ -68,14 +75,6 @@ def test_testcase_nameing():
     tc = TestCaseFedora()
     expected = "container_ref=quay.io/fedora/fedora-bootc:40,rootfs=btrfs"
     assert f"{tc}" == expected, f"{tc} != {expected}"
-
-
-@dataclasses.dataclass
-class TestCaseC10S(TestCase):
-    container_ref: str = os.getenv(
-        "BIB_TEST_BOOTC_CONTAINER_TAG",
-        "quay.io/centos-bootc/centos-bootc:stream10")
-    osinfo_template: str = "CentOS Stream 10 ({arch})"
 
 
 def gen_testcases(what):  # pylint: disable=too-many-return-statements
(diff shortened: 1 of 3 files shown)


henrywang (approved):
LGTM. Thanks!

achilleas-k:
`osinfo-detect` on the C10S ISO just returns
```
Media is bootable.
```
which is making a test fail.  I don't know if that's a bug or something we should change in tests.

achilleas-k:
> `osinfo-detect` on the C10S ISO just returns
> 
> ```
> Media is bootable.
> ```
> 
> which is making a test fail. I don't know if that's a bug or something we should change in tests.

@mmartinv I assume this depends on the version of libosinfo and what distro versions it knows about.  Would it make sense to skip (or adapt) the test for CentOS Stream 10 or should this be working already?

mmartinv:
@achilleas-k this PR will conflict with #680. I like the of moving the `osinfo_template` logic to a function instead of putting it on each test case.

achilleas-k:
> @achilleas-k this PR will conflict with #680. I like the of moving the `osinfo_template` logic to a function instead of putting it on each test case.

Right.  Let's get that one in first then and I'll adapt this one accordingly.

henrywang:
Hi @achilleas-k, can we land this PR? Thanks!

achilleas-k:
Rebased and resolved conflicts.

mvo5 (approved):
Thank you!

test/testcases.py
```diff
@@ -70,29 +70,38 @@ def test_testcase_nameing():
     assert f"{tc}" == expected, f"{tc} != {expected}"
 
 
+@dataclasses.dataclass
+class TestCaseC10S(TestCase):
+    container_ref: str = os.getenv(
+        "BIB_TEST_BOOTC_CONTAINER_TAG",
+        "quay.io/centos-bootc/centos-bootc:stream10")
+    osinfo_template: str = "CentOS Stream 10 ({arch})"
```
mvo5, line 78: Ups, I overlooked this, this will probably need to be tweaked slightly as this code has changed: 
[0001-test-tweak-TestCaseC10S-testcase-and-osinfo_for-help.patch.txt](https://github.com/user-attachments/files/17951111/0001-test-tweak-TestCaseC10S-testcase-and-osinfo_for-help.patch.txt)

mvo5: We should probably refactor this again :/ it would be nice if osinfo would life closer to the testcases, they did originally belong together but then the test-case naming became too verbose, but the current split (which is my fault) is also not ideal

mvo5 (dismissed):
Thank you!

achilleas-k:
https://gitlab.com/libosinfo/osinfo-db/-/tree/main/data/os/centos.org?ref_type=heads

Added 5 days ago.

mvo5 (approved):

achilleas-k:
Thanks for pushing the fix!

mvo5 (dismissed):

achilleas-k:
Rebased and fixed conflicts.

mvo5 (approved):

Issue #700, opened 2024-11-03
Bib break anaconda-iso build on Fedora 41


chunfuwen:
```
sudo podman run --rm -it --privileged --pull=newer --security-opt label=type:unconfined_t -v /var/lib/libvirt/images/output:/output -v /var/lib/libvirt/images/config.json:/config.json    -v /var/lib/libvirt/images/auth.json:/run/containers/0/auth.json   quay.io/centos-bootc/bootc-image-builder:latest  --type anaconda-iso --tls-verify=true  --config /config.json  quay.io/fedora/fedora-bootc:41  --chown 107:107  --rootfs xfs

Trying to pull quay.io/centos-bootc/bootc-image-builder:latest...
Getting image source signaturesr/lib/libvirt/images/
Copying blob 2b054662280c done   | 
Copying blob 27bd6671c90f done   | 
Copying blob d651a781d40e done   | 
Copying config b7befecd2e done   | 
Writing manifest to image destination
Generating manifest manifest-anaconda-iso.json
Trying to pull quay.io/fedora/fedora-bootc:41...
Getting image source signatures
Copying blob 953770343a39 done   | 
Copying blob 953770343a39 done   | 
Copying blob 2a318413e6e2 done   | 
...
Copying blob a855da5d6fbe done   | 
Copying blob 4f0bf6084203 done   | 
Copying blob 9dad063a624b done   | 
Copying config 7562d2b879 done   | 
Writing manifest to image destination
Traceback (most recent call last):
  File "/osbuild-depsolve-dnf", line 30, in <module>
    from osbuild.solver.dnf import DNF as Solver
  File "/osbuild/solver/dnf.py", line 7, in <module>
    import dnf
ModuleNotFoundError: No module named 'dnf'
2024/11/03 01:40:05 error: cannot build manifest: cannot depsolve: running osbuild-depsolve-dnf failed:

```

chunfuwen:
```
[root@xx-mages]# skopeo inspect docker://quay.io/fedora/fedora-bootc:41 | jq  -r '.Labels' |tee label.log
{
  "containers.bootc": "1",
  "org.opencontainers.image.version": "41.20241101.0",
  "ostree.bootable": "true",
  "ostree.commit": "dd4dc5440e420dbfd79c4581b62f50ad2078cad7c59e751060d8c3a45884712b",
[...]

PR #704, opened 2024-11-05
bib: resolve container content from the external dnf again

mvo5:
This commit goes back to the approach we had in PR#565 to resolve using
the solver from outside of the container. The reason to back is that
some containers (like f41) do not have the python dnf available so
they break with the "run dnfjson inside" approach (that was meant
as a quick fix only anyway).

Huge kudos to Achilleas for most of the work here.

Closes: https://github.com/osbuild/bootc-image-builder/issues/700

It would be ideal if this could be tested end-to-end on a rhel9 system building a rhel9 iso installer.

### DIFF
--- a/bib/internal/container/solver.go
+++ b/bib/internal/container/solver.go
@@ -2,6 +2,7 @@ package container
 
 import (
 	"fmt"
+	"os"
 	"os/exec"
 	"path/filepath"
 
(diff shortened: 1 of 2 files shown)


achilleas-k (commented):
LGTM.  The nested test is a bit out there but I like it!

Linter is unhappy though.

achilleas-k:
Smoke test failed with "Bad gateway" -_-
Restarted

achilleas-k (approved):

mvo5:
> LGTM. The nested test is a bit out there but I like it!

Yeah, not super happy about it, ideas (very) welcome how to make it cleaner!
 
> Linter is unhappy though.

Fixed now. There was also another issue that the new unit tests in testingfarm uncovered so yay for tests.

achilleas-k (approved):

PR #707, opened 2024-11-08
fix: add a missing dependency for the anaconda-iso installer

mmartinv:
Add the the python3-pam package to the list of dependencies
as it seems to be required for the anaconda-core rpm.

Resolves: fedora-iot/iot-distro#66

Signed-off-by: Miguel Martín <mmartinv@redhat.com>


achilleas-k (approved):
Thanks!

Issue #708, opened 2024-11-09
enable FIPS in kernel args at install time for anaconda-iso builds


stevenperaltaf5:
when settings FIPS to true in blueprint, kernel args for the install ISO should be appended with `fips=1` so that our system boots with the kernel in FIPS mode during installation

an example config.toml:
```toml
[customizations]
fips = true

[customizations.installer.kickstart]
contents = """
text --non-interactive
network --bootproto=dhcp --device=eno1 --activate --onboot=on

rootpw --lock --iscrypted locked
user --name admin --password admin --groups wheel

zerombr
clearpart --all --initlabel --disklabel=gpt
autopart --noswap --type=lvm

reboot
"""
```

after building the image, the resultant manifest doesn't show any indication that `fips=1` has been added to the kernel args in the `org.osbuild.grub2.iso` stage of the `efiboot-tree` pipeline:
```json
{
  "version": "2",
  "pipelines": [
    ...
    {
      "name": "efiboot-tree",
      "build": "name:build",
      "stages": [
        {
          "type": "org.osbuild.grub2.iso",
          "options": {
            "product": {
              "name": "AlmaLinux",
              "version": "9.4"
            },
            "kernel": {
              "dir": "/images/pxeboot",
              "opts": [
                "inst.stage2=hd:LABEL=Container-Installer-aarch64",
                "inst.ks=hd:LABEL=Container-Installer-aarch64:/osbuild.ks"
              ]
            },
            "isolabel": "Container-Installer-aarch64",
            "architectures": [
              "AA64"
            ],
            "vendor": "almalinux"
          }
        }
      ]
    },
    ...
  ]
  ...
}
```

viewing the kernel args in the grub boot menu option shows that `fips=1` hasn't been appended:
![Screenshot 2024-11-08 at 9 58 28 PM](https://github.com/user-attachments/assets/246d693f-885d-42bb-8b52-05e3047e85e1)

after installation, `fips-mode-setup --check` yields a "FIPS mode is disabled." response:
![image](https://github.com/user-attachments/assets/a0dfe149-62e5-4f83-a386-51b91408ded5)

henrywang:
Does this feature conflict with the other non iso/anaconda-iso image type which enabled FIPS in base image https://gitlab.com/fedora/bootc/base-images/-/merge_requests/31? Thanks.

cgwalters:
[...]

PR #709, opened 2024-11-09
fix: set fips in manifest according to our customizations

stevenperaltaf5:
addresses #708

### DIFF
--- a/bib/cmd/bootc-image-builder/image.go
+++ b/bib/cmd/bootc-image-builder/image.go
@@ -373,7 +373,7 @@ func manifestForISO(c *ManifestConfig, rng *rand.Rand) (*manifest.Manifest, erro
 	if c.Config != nil {
 		customizations = c.Config.Customizations
 	}
-
+	img.FIPS = customizations.GetFIPS()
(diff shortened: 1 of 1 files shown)


cgwalters (commented):
I don't know enough about this code to 100% say this is correct but it looks plausible.

Basically as long as we end up with `fips=1` on the Anaconda kernel commandline, that's what we want.

mvo5:
Thanks, this looks good, I did a small followup in https://github.com/osbuild/bootc-image-builder/pull/715 that adds a (manifest level) test to ensure this does not regress.

mvo5 (approved):
I guess we also want to support this customization for non-iso images(?)

mvo5:
I also did a small followup in https://github.com/osbuild/images/pull/1035 for non-iso bootc containers but there is an open questions if this should come via customization or the bootc container itself would set it.

henrywang:
> I guess we also want to support this customization for non-iso images(?)

Only ISO use non-bootc to deploy image. bootc supports kargs. So for non-ISO, FIPS can be enabled in Containerfile. Just in case `fips=1` configured twice.

PR #689, opened 2024-10-23
bib: remove usage of `BootcLegacyDiskImage` (for qemu-9.1)

mvo5:
This commit enables bib to work with the qemu-9.1+openat2 patches as proposed in https://src.fedoraproject.org/rpms/qemu/pull-request/69.

With the qemu-user-statuc-aarch64 rpm produced by `fedpkg mockbuild` from this package the cross-arch test build works again.
```
$ sudo PYTHONPATH=. pytest './test/test_build.py::test_image_boots[quay.io/centos-bootc/centos-bootc:stream9,raw,CentOS Stream 9 ({arch}),arm64]' -s -vv
============================= test session starts ==============================
platform linux -- Python 3.12.7, pytest-7.4.3, pluggy-1.3.0 -- /usr/bin/python3
cachedir: .pytest_cache
rootdir: /home/mvogt/devel/osbuild/bootc-image-builder
configfile: pytest.ini
plugins: repeat-0.9.3, xdist-3.5.0
collected 1 item

test/test_build.py::test_image_boots[quay.io/centos-bootc/centos-bootc:stream9,raw,CentOS Stream 9 ({arch}),arm64] [1/2] STEP 1/9: FROM registry.fedoraproject.org/fedora:40 AS builder
...
PASSEDChecking disk usage for /var/tmp/bib-tests/shared0/68b20145da2cd3f2/image/disk.raw
NOTE: disk usage after /var/tmp/bib-tests/shared0/68b20145da2cd3f2/image/disk.raw: 712472.71936 / 1998694.907904
Untagged: quay.io/centos-bootc/centos-bootc:stream9
Deleted: 0de1f90b11cbe7e2768101f6c55b4dd9841c13a2f87c8d9c177aea74ade88050

============================= slowest 10 durations =============================
1033.54s setup    test/test_build.py::test_image_boots[quay.io/centos-bootc/centos-bootc:stream9,raw,CentOS Stream 9 ({arch}),arm64]
67.62s call     test/test_build.py::test_image_boots[quay.io/centos-bootc/centos-bootc:stream9,raw,CentOS Stream 9 ({arch}),arm64]
2.30s teardown test/test_build.py::test_image_boots[quay.io/centos-bootc/centos-bootc:stream9,raw,CentOS Stream 9 ({arch}),arm64]
======================== 1 passed in 1103.58s (0:18:23) ========================
```

---

This will also re-enable the cross-arch test via an updated qemu-user-static (same patches for the GH runners as we use for the f41 rpm packages of qemu)

PR #712, opened 2024-11-11
workflow: move workflow to run on ubuntu-24.04

mvo5:
This commit moves the GH runner to move to 24.04. This means we no longer need a extra repository to get a current version of podman and it generally means our software stack is more modern and closer to what version people run.

Note that this also drops "podman,skopeo" as those are preinstalled.

Unfortuantely the default 24.04 podman configuration is buggy and a workaround is needed.

[split out from https://github.com/osbuild/bootc-image-builder/pull/689]

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -67,7 +67,7 @@ jobs:
   integration:
     # TODO: run this also via tmt/testing-farm
     name: "Integration"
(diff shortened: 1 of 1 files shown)


ondrejbudai (approved):
Thanks!

achilleas-k (approved):

PR #713, opened 2024-11-11
workflow: run pytest in parallel

mvo5:
A bit of an experiment to see how much a naive parallel approach wins us


mvo5:
It seems parallel runs give lot's of strange errors, we really need to investigate if this is just in GH action or if there is a deeper problem with running bib in parallel but for now I will close the PR as there are no easy wins here it seems :(

PR #714, opened 2024-11-12
tests: drop (unsigned) fedora ISO testcase

mvo5:
This commit drops the "unsigned" fedora ISO genartion testcases. We also test unsigned ISO generation via centos-9 and we test the fedora ISO generation via the "signed" fedora ISO test so running this specific test gives us little and it is also one of the most "expensive" tests we run (around 10min).


ondrejbudai (approved):
Thanks! ❤️

PR #715, opened 2024-11-12
test: add test that ensures fips customizations work

mvo5:
Followup for https://github.com/osbuild/bootc-image-builder/pull/709 that adds a test.

This commit adds a quick manifest level test that ensures that the
FIPS customizations set for the iso installer are actually set
in the manifest.


ondrejbudai (approved):
Awesome, thanks! ❤️

PR #721, opened 2024-11-21
bib: add integration test for the new `Disk` customizations

mvo5:
[It is draft right now because it has quite a few rough edges in the code but mostly because the LVM case is not working right now and it's unclear why, but pushing it will help with exposure/feedback, there are also some issues in "images" that are uncovered here that should be fixed there (like minsize seems to be not working for strings)]

This PR adds support for the new `Disk` customizations from https://github.com/osbuild/images/pull/1041 into bib.

### DIFF
--- a/test/test_build.py
+++ b/test/test_build.py
@@ -38,6 +38,7 @@ class ImageBuildResult(NamedTuple):
     img_arch: str
     container_ref: str
     rootfs: str
+    disk_config: str
     username: str
     password: str
     ssh_keyfile_private_path: str
@@ -203,7 +204,6 @@ def sign_container_image(gpg_conf: GPGConf, registry_conf: RegistryConf, contain
         f"docker://{container_ref}",
         f"docker://{signed_container_ref}",
     ]
-    print(cmd)
     subprocess.run(cmd, check=True, env={"GNUPGHOME": gpg_conf.home_dir})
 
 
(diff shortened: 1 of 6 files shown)


test/test_build.py
```diff
@@ -38,6 +38,7 @@ class ImageBuildResult(NamedTuple):
     img_arch: str
     container_ref: str
     rootfs: str
+    partition_mode: str
```
achilleas-k, line 41: This confused me a bit because it made me think of the [`disk.PartitioningMode`](https://github.com/osbuild/bootc-image-builder/blob/47e0292b184645d8a8d82278f67af9c7d274028a/bib/cmd/bootc-image-builder/image.go#L210-L216) that we have for the older partitioning customizations.  This is only meant for tests and not user facing, so it's not really that important, but it might be confusing for test writing and reading.

I suggest we rename it to something else, something that implies it's a keyword for a particular test partitioning configuration (which is what it is, right?).  Partitioning config maybe?

mvo5: Yeah, excellent feedback, I went with `disk_config` for now but `partion_config` or `disk_customization` would of course also be an option

achilleas-k:
The tests seem fine by themselves but they fail when I run them all at the same time.  Meaning, if I run `pytest -k 'test_image_boots and lvm'` and `pytest -k 'test_image_boots and btrfs'` they succeed, but running `pytest -k 'test_image_boots'` causes failures in those two tests.  I'll have a look at the pytest fixture handling.

test/test_build.py
```diff
@@ -663,17 +657,52 @@ def test_multi_build_request(images):
...
+    exit_status, output = test_vm.run("findmnt --json", user="root",
+                                      keyfile=image_type.ssh_keyfile_private_path)
+    assert exit_status == 0
+    findmnt = json.loads(output)
+    if dc := image_type.disk_config:
+        if dc == "lvm":
+            mnts = [mnt for mnt in findmnt["filesystems"][0]["children"]
+                    if mnt["target"] == "/sysroot"]
+            assert len(mnts) == 1
+            assert "/dev/mapper/vg00-rootlv" == mnts[0]["source"]
```
achilleas-k, line 701: Ideally we would use the config to create `vg00-rootlv` dynamically from the input, but this is ok.

achilleas-k (dismissed):

achilleas-k (approved):

achilleas-k (approved):

PR #734, opened 2024-11-28
bib: load repo keys using external paths (COMPOSER-2408)

achilleas-k:
Since d192f85ae9e5b6e5665b8ea1ae39677b24a9352a, we switched to using dnf (osbuild-depsolve-dnf) from BIB itself, but using the base (bootc) container root to read configurations, certificates, keys, etc. Unfortunately, what we didn't change, is the way in which we load the repository certs after the depsolve.

The extractTLSKeys() function currently takes a Container instance which implements ReadFile() by `cat`ing a file from inside the running container.  However, the paths we get from the depsolve are relative to the BIB worktree.  Therefore, we can simply read the files directly.

Add a SimpleFileReader that implements the fileReader interface to simply read a file from a path.  We could simplify this to not require a reader interface at all, but let's keep the implementation in case we ever need to switch back to the older way of depsolving from inside the container.

Resolves COMPOSER-2408


bib/cmd/bootc-image-builder/mtls.go
```diff
@@ -19,6 +20,18 @@ type fileReader interface {
 	ReadFile(string) ([]byte, error)
 }
 
+type SimpleFileReader struct{}
+
+func (SimpleFileReader) ReadFile(path string) ([]byte, error) {
+	fp, err := os.Open(path)
```
ondrejbudai, line 26: I think you can just use https://pkg.go.dev/os#ReadFile

ondrejbudai (dismissed):
Thanks! Should we implement an E2E test in a follow-up? This logic seems to be quite brittle. :/

bib/cmd/bootc-image-builder/mtls.go
achilleas-k: Indeed.  Silly me.  Amended.

ondrejbudai (approved):
Thanks!

mvo5 (approved):
Thanks for finding/fixing this! This is great!

PR #737, opened 2024-11-29
go.mod: update osbuild/images to v0.103.0

achilleas-k:
This update includes support for swap partitions and logical volumes.

See: https://github.com/osbuild/images/pull/1072

Tests will be added to #721.


ondrejbudai:
We need minsize defined in the configs here, I think this is caused by https://github.com/osbuild/images/commit/4c2bce4bdf875b40f2f15091eec6fc09053d2be1

PR #739, opened 2024-12-02
README: drop $(pwd) from volume mounts in podman commands

achilleas-k:
Using `./` to prefix a path relative to the current directory works just
as well and doesn't require quoting for potential paths with spaces, nor
a subshell.

The `./` prefix (or an abs path) is required when referring to single
local directories with just one path component, otherwise podman
interprets the path as a volume name.


ondrejbudai (approved):
Thanks!

mvo5 (approved):

PR #743, opened 2024-12-04
main,osbuildprogress: add --progress=text,debug support (COMPOSER-2394)

mvo5:
This adds a new `progress` flag that makes use of the
osbuild jsonseq progress information to show progress and
hide the low-level details from the user.

Here is what it looks like:
https://asciinema.org/a/cJpB6066UoDDzAylwHVHjEEIN

(c.f. COMPOSER-2394)

### DIFF
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -487,6 +487,11 @@ func cmdBuild(cmd *cobra.Command, args []string) error {
 
 	pbar.SetMessagef("Build complete!")
 	if upload {
+		// XXX: pass our own progress.ProgressBar here
+		// *for now* just stop our own progress and let the uploadAMI
+		// progress take over - but we really need to fix this in a
+		// followup
+		pbar.Stop()
 		for idx, imgType := range imgTypes {
 			switch imgType {
 			case "ami":
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -244,9 +244,7 @@ func manifestFromCobra(cmd *cobra.Command, args []string, pbar progress.Progress
 	}
 
 	pbar.SetPulseMsgf("Manifest generation step")
-	if err := pbar.Start(); err != nil {
-		return nil, nil, fmt.Errorf("cannot start progress: %v", err)
-	}
+	pbar.Start()
 
 	if err := setup.ValidateHasContainerTags(imgref); err != nil {
 		return nil, nil, err
@@ -434,11 +432,7 @@ func cmdBuild(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return fmt.Errorf("cannto create progress bar: %w", err)
 	}
-	defer func() {
-		if err := pbar.Stop(); err != nil {
-			logrus.Warnf("progressbar stopping failed: %v", err)
-		}
-	}()
+	defer pbar.Stop()
 
 	manifest_fname := fmt.Sprintf("manifest-%s.json", strings.Join(imgTypes, "-"))
 	pbar.SetMessagef("Generating manifest %s", manifest_fname)
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -171,16 +171,17 @@ func saveManifest(ms manifest.OSBuildManifest, fpath string) error {
 	return nil
 }
 
+// manifestFromCobra generate an osbuild manifest from a cobra commandline.
+//
+// It takes an unstarted progres bar and will start it at the right
+// point (it cannot be started yet to avoid the "podman pull" progress
+// and our progress fighting). The caller is responsible for stopping
+// the progress bar (this function cannot know what else needs to happen
+// after manifest generation).
+//
+// TODO: provide a podman progress reader to integrate the podman progress
+// into our progress.
 func manifestFromCobra(cmd *cobra.Command, args []string, pbar progress.ProgressBar) ([]byte, *mTLSConfig, error) {
-	if pbar == nil {
-		var err error
-		pbar, err = progress.New("plain")
-		// this should never happen
-		if err != nil {
-			return nil, nil, err
-		}
-	}
-
 	cntArch := arch.Current()
 
 	imgref := args[0]
@@ -332,7 +333,14 @@ func manifestFromCobra(cmd *cobra.Command, args []string, pbar progress.Progress
 }
 
 func cmdManifest(cmd *cobra.Command, args []string) error {
-	mf, _, err := manifestFromCobra(cmd, args, nil)
+	pbar, err := progress.New("")
+	if err != nil {
+		// this should never happen
+		return fmt.Errorf("cannot create progress bar: %w", err)
+	}
+	defer pbar.Stop()
+
+	mf, _, err := manifestFromCobra(cmd, args, pbar)
 	if err != nil {
 		return fmt.Errorf("cannot generate manifest: %w", err)
 	}
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -21,12 +21,12 @@ import (
 	"github.com/osbuild/images/pkg/container"
 	"github.com/osbuild/images/pkg/dnfjson"
 	"github.com/osbuild/images/pkg/manifest"
-	"github.com/osbuild/images/pkg/osbuild"
 	"github.com/osbuild/images/pkg/rpmmd"
 
 	"github.com/osbuild/bootc-image-builder/bib/internal/buildconfig"
 	podman_container "github.com/osbuild/bootc-image-builder/bib/internal/container"
 	"github.com/osbuild/bootc-image-builder/bib/internal/imagetypes"
+	"github.com/osbuild/bootc-image-builder/bib/internal/progress"
 	"github.com/osbuild/bootc-image-builder/bib/internal/setup"
 	"github.com/osbuild/bootc-image-builder/bib/internal/source"
 	"github.com/osbuild/bootc-image-builder/bib/internal/util"
@@ -171,7 +171,16 @@ func saveManifest(ms manifest.OSBuildManifest, fpath string) error {
 	return nil
 }
 
-func manifestFromCobra(cmd *cobra.Command, args []string) ([]byte, *mTLSConfig, error) {
+func manifestFromCobra(cmd *cobra.Command, args []string, pbar progress.ProgressBar) ([]byte, *mTLSConfig, error) {
+	if pbar == nil {
+		var err error
+		pbar, err = progress.New("plain")
+		// this should never happen
+		if err != nil {
+			return nil, nil, err
+		}
+	}
+
 	cntArch := arch.Current()
 
 	imgref := args[0]
@@ -232,6 +241,11 @@ func manifestFromCobra(cmd *cobra.Command, args []string) ([]byte, *mTLSConfig, 
 		}
 	} else {
 		logrus.Debug("Using local container")
+	}
+
+	pbar.SetPulseMsgf("Manifest generation step")
+	if err := pbar.Start(); err != nil {
+		return nil, nil, fmt.Errorf("cannot start progress: %v", err)
 	}
 
 	if err := setup.ValidateHasContainerTags(imgref); err != nil {
@@ -320,7 +334,7 @@ func manifestFromCobra(cmd *cobra.Command, args []string) ([]byte, *mTLSConfig, 
 }
 
 func cmdManifest(cmd *cobra.Command, args []string) error {
-	mf, _, err := manifestFromCobra(cmd, args)
+	mf, _, err := manifestFromCobra(cmd, args, nil)
 	if err != nil {
 		return fmt.Errorf("cannot generate manifest: %w", err)
 	}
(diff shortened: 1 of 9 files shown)


mvo5:
Moved back to draft as the "pb.Pool" idea does not work with podman as it does not give us a real pty

mvo5:
I moved this back to draft as the pb.Pool used does not actually work with "podman run" (the integration test found this). podman run does give us alog-driver=passthough" by default (and *not* passthrough-tty (to avoid TIOCSTI attacks)). So the sensible thing is probably to look into how we can build the progressbar without pb.Pool and without terminal raw mode (which is a bit of a challenge).

bib/internal/progress/progress.go
```diff
@@ -0,0 +1,296 @@
...
+
+	"github.com/osbuild/images/pkg/osbuild"
+)
+
+var (
+	osStderr         io.Writer = os.Stderr
+	isattyIsTerminal           = isatty.IsTerminal
+)
+
+// ProgressBar is an interfacs for progress reporting when there is
```
ondrejbudai, line 23: interfacs => interface

bib/internal/progress/progress.go
```diff
@@ -0,0 +1,296 @@
...
+// an arbitrary amount of sub-progress information (like osbuild)
+type ProgressBar interface {
+	// SetProgress sets the progress details at the given "level".
+	// Levels should start with "0" and increase as the nesting
+	// gets deeper.
+	SetProgress(level int, msg string, done int, total int) error
+	// The high-level message that is displayed in a spinner
```
ondrejbudai, line 30: I think I need examples here. Both messages are described as high-level. But also the word "Pulse" is used, and I don't understand what it means.

bib/internal/progress/progress.go
```diff
@@ -0,0 +1,296 @@
...
+	// (e.g. "Building image foo")
+	SetPulseMsg(fmt string, args ...interface{})
```
ondrejbudai, line 32: what about `SetPulseMsgf` to indicate that formatting is supported? It seems to be a pretty standard practice in Go.

bib/internal/progress/progress.go
```diff
@@ -0,0 +1,296 @@
...
+	// A high level message with the last high level status
+	// (e.g. "Started downloading")
+	SetMessage(fmt string, args ...interface{})
+	Start() error
```
ondrejbudai, line 36: When do I need to call `Start` and `Stop`?

bib/internal/progress/progress.go
```diff
@@ -0,0 +1,296 @@
...
+type termProgressBar struct {
+	spinnerPb   *pb.ProgressBar
+	msgPb       *pb.ProgressBar
+	subLevelPbs []*pb.ProgressBar
+
+	pool        *pb.Pool
+	poolStarted bool
+}
+
+// NewProgressBar creates a new default pb3 based progressbar suitable for
```
ondrejbudai, line 71: NewProgressBar => NewTermProgressBar

or maybe even s/term/terminal? The short version is kinda confusing me, but it migth be a me-issue.

bib/internal/progress/progress.go
```diff
@@ -0,0 +1,296 @@
...
+func shorten(msg string) string {
+	msg = strings.Replace(msg, "\n", " ", -1)
+	// XXX: make this smarter
+	if len(msg) > 60 {
+		return msg[:60] + "..."
+	}
+	return msg
+}
+
+func (ppb *termProgressBar) SetPulseMsg(msg string, args ...interface{}) {
```
ondrejbudai, line 123: ppb => tpb? or just b

bib/internal/progress/progress.go
```diff
@@ -0,0 +1,296 @@
...
+}
+
+// NewPlainProgressBar starts a new "plain" progressbar that will just
+// prints message but does not show any progress.
+func NewPlainProgressBar() (ProgressBar, error) {
+	np := &plainProgressBar{w: osStderr}
+	return np, nil
+}
+
+func (np *plainProgressBar) SetPulseMsg(msg string, args ...interface{}) {
```
ondrejbudai, line 159: np => ppb? or just b...

bib/internal/progress/progress.go
```diff
@@ -0,0 +1,296 @@
...
+		return runOSBuildOld(pb, manifest, store, outputDirectory, exports, extraEnv)
+	}
+}
+
+func runOSBuildOld(pb ProgressBar, manifest []byte, store, outputDirectory string, exports, extraEnv []string) error {
+	_, err := osbuild.RunOSBuild(manifest, store, outputDirectory, exports, nil, extraEnv, false, os.Stderr)
+	return err
+}
+
+func runOSBuildNew(pb ProgressBar, manifest []byte, store, outputDirectory string, exports, extraEnv []string) error {
```
ondrejbudai, line 235: I know that this is temporary, but the function name is really bad. :D

bib/internal/progress/progress.go
```diff
@@ -0,0 +1,296 @@
...
+		st, err := osbuildStatus.Status()
+		if err != nil {
+			return fmt.Errorf("error reading osbuild status: %w", err)
+		}
+		if st == nil {
+			break
+		}
+		i := 0
+		for p := st.Progress; p != nil; p = p.SubProgress {
+			// XXX: osbuild gives us bad progress messages
```
ondrejbudai, line 280: What does this mean?

bib/internal/progress/progress_test.go
```diff
@@ -0,0 +1,106 @@
...
+
+func TestPlainProgress(t *testing.T) {
+	var buf bytes.Buffer
+	restore := progress.MockOsStderr(&buf)
+	defer restore()
+
+	// plain progress never generates output
+	pbar, err := progress.NewPlainProgressBar()
+	assert.NoError(t, err)
+	err = pbar.SetProgress(0, "set-prorgress", 1, 100)
```
ondrejbudai, line 60: prorgress => progress

But it really doesn't matter. :)

bib/internal/progress/progress_test.go
```diff
@@ -0,0 +1,106 @@
...
+		}
+	}
+}
+
+func TestPlainProgress(t *testing.T) {
+	var buf bytes.Buffer
+	restore := progress.MockOsStderr(&buf)
+	defer restore()
+
+	// plain progress never generates output
```
ondrejbudai, line 57: I suggest moving this comment 2 lines lower, so it's clear that it talks about `SetProgress`

bib/internal/progress/progress.go
```diff
@@ -0,0 +1,296 @@
...
+	fmt.Fprintf(np.w, "%s[%v / %v] %s", strings.Repeat("  ", subLevel), done, total, msg)
+	fmt.Fprintf(np.w, "\n")
+	return nil
+}
+
+// XXX: merge variant back into images/pkg/osbuild/osbuild-exec.go
+func RunOSBuild(pb ProgressBar, manifest []byte, store, outputDirectory string, exports, extraEnv []string) error {
+	switch pb.(type) {
+	case *termProgressBar, *debugProgressBar:
+		return runOSBuildNew(pb, manifest, store, outputDirectory, exports, extraEnv)
```
ondrejbudai, line 224: Stupid question: cannot we use this method for all progress bars?

bib/cmd/bootc-image-builder/main.go
```diff
@@ -434,7 +459,8 @@ func cmdBuild(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("cannot save manifest: %w", err)
 	}
 
-	fmt.Printf("Building %s\n", manifest_fname)
+	pbar.SetPulseMsg("Image generation step")
```
ondrejbudai, line 462: I would go with "Image building step"

README.md
```diff
@@ -49,6 +49,7 @@ sudo podman run \
     --rm \
     -it \
     --privileged \
+    --log-driver=passthrough-tty \
```
ondrejbudai, line 52: Ugh, this is a bummer. Let's talk about this.

ondrejbudai (changes requested):
This is awesome, I only found some issues mostly with comments and method names. PTAL :)

And then ofc we need to discuss the whole `--log-driver=passthrough-tty` matter. :)

bib/internal/progress/progress.go
mvo5: I updated the comment(s) and include the following in
the commit message now:
```
The interface is very tailored to the needs of bib but it should
be similar enough to be reusable by `image-builder-cli`.

It contains a top level spinner that gives the current step:
"Make manifest" or "Building image". The the subprogress from
osbuild: the pipelines and the stages from the pipelines as
progress bars and the last line is the latest status message,
usually something from osbuild like "Starting pipeline qcow2".

The spinenr is a bit of a creative choice, it could also be a 
progress bar because we know the steps (manifest+build). But the 
runtime is totally dominated by the build step so having a first level
progress bar is odd as the UI will spend 90% of the time in it.
Plus the spinner gives it more a more "dynamic" appearnce.
```
I hope this makes it a bit clearer.

bib/internal/progress/progress.go
mvo5: Uhhh, indeed it is, sorry for that! Fixed now 

/me grabs a brown paperbag.

bib/internal/progress/progress.go
mvo5: Sorry, I think this is not actually the case anymore, I fixed this (removed it)

bib/internal/progress/progress.go
mvo5: I will push a small commit with cleanups and a docstring for this.

bib/internal/progress/progress.go
mvo5: Yes, I  added a comment - I wanted to be conservative for users like the "podman bootc" people who AFAIK currently read the raw osbuild terminal output. I don't know if/how-much (if anything) they do with it but until that is discussed I wanted to keep the existing behavior. I think we will give them a jsonseq progress.

bib/internal/progress/progress.go
mvo5: Fwiw, I can provide a version without the spinner and three levels of subprogress instead, it's pretty trivial to do this, I played with it but I don't like it that much TBH, i prefer the spinner.

bib/internal/progress/progress.go
mvo5: I tweaked the naming, please let me know if you have more suggestions here (naming is hard!)

bib/cmd/bootc-image-builder/main.go
```diff
@@ -171,7 +171,16 @@ func saveManifest(ms manifest.OSBuildManifest, fpath string) error {
 	return nil
 }
 
-func manifestFromCobra(cmd *cobra.Command, args []string) ([]byte, *mTLSConfig, error) {
+func manifestFromCobra(cmd *cobra.Command, args []string, pbar progress.ProgressBar) ([]byte, *mTLSConfig, error) {
+	if pbar == nil {
```
ondrejbudai, line 175: I'm thinking about `progress.ProgressBar` as a resource, because it needs to be `Stop`ped at some point. So I think there should be a clean owner that's responsible for calling the `Stop` method. This logic currently makes the ownership murky. Can we restructure it somehow?

bib/cmd/bootc-image-builder/main.go
```diff
@@ -234,6 +243,9 @@ func manifestFromCobra(cmd *cobra.Command, args []string) ([]byte, *mTLSConfig,
 		logrus.Debug("Using local container")
 	}
 
+	pbar.SetPulseMsgf("Manifest generation step")
+	pbar.Start()
```
ondrejbudai, line 247: There's no `defer pbar.Stop()` which can lead to "leaking".

ondrejbudai (changes requested):
This is very nice! :) PTAL at the small concerns inline.

bib/cmd/bootc-image-builder/main.go
mvo5: I tweaked this and added the following comment:
```
manifestFromCobra generate an osbuild manifest from a cobra
commandline and it also takes a progress.ProgressBar as input.

It needs an unstarted progres bar and will start it at the right
point (it cannot be started yet to avoid the "podman pull" progress
and our progress fighting). The caller is responsible for stopping
the progress bar (this functtion cannot know what else needs to happen
after manifest generation).

TODO: provide a podman progress reader to integrate the podman progress
into our progress.
```

I can look into tweaking this further, right now the issue is that `manifestFromCobra` is calling `podman pull` and here we do want to also provide a progress stream which we currently just get directly from podman. So we cannto yet start our own progress (or the two would fight over the screen). 

I guess an alternative would be to split manifestFromCobra into pullContainer and generateManifest or something and then the pullContainer would not take a progress but generateManifest would and that can then be started too (or we would even remove start as any creation of a progress would automatically start it).

I can look into this Monday if you feel the existing comment/structure is still murky .

ondrejbudai (approved):
Thanks for clarifying!

bib/internal/progress/progress.go
```diff
@@ -200,6 +220,12 @@ func (b *terminalProgressBar) Stop() {
 		logrus.Warnf("no progress channel shutdown after 1sec")
 	}
 	b.shutdownCh = nil
+	// This should never happen but be parandoid, this should
```
ondrejbudai, line 228: parandoid => paranoid

cool word, though

bib/cmd/bootc-image-builder/main.go
```diff
@@ -171,16 +171,17 @@ func saveManifest(ms manifest.OSBuildManifest, fpath string) error {
 	return nil
 }
 
+// manifestFromCobra generate an osbuild manifest from a cobra commandline.
+//
+// It takes an unstarted progres bar and will start it at the right
+// point (it cannot be started yet to avoid the "podman pull" progress
+// and our progress fighting). The caller is responsible for stopping
+// the progress bar (this functtion cannot know what else needs to happen
```
ondrejbudai, line 179: functtion => function

ondrejbudai (approved):
Love this, thank you! ❤️

achilleas-k (approved):
Awesome stuff.  Thank you!

ondrejbudai (approved):
Thanks!

PR #745, opened 2024-12-05
test: add integration test for the new `swap` customization

mvo5:
This commit adds an integration test for the new swap disk customization that got added to the `images` library in PR https://github.com/osbuild/images/pull/1072

### DIFF
--- a/test/test_manifest.py
+++ b/test/test_manifest.py
@@ -633,3 +633,95 @@ def test_manifest_disk_customization_btrfs(tmp_path, build_container):
     st = find_bootc_install_to_fs_stage_from(output)
     assert st["mounts"][0]["type"] == "org.osbuild.btrfs"
     assert st["mounts"][0]["target"] == "/"
+
+
+def find_mkswap_stage_from(manifest_str):
+    manifest = json.loads(manifest_str)
+    for pipeline in manifest["pipelines"]:
+        if pipeline["name"] == "image":
+            for st in pipeline["stages"]:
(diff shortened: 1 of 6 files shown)


test/test_manifest.py
```diff
@@ -633,3 +633,97 @@ def test_manifest_disk_customization_btrfs(tmp_path, build_container):
...
+    config = {
+        "customizations": {
+            "disk": {
+                "partitions": [
+                    {
+                        "type": "lvm",
+                        "minsize": "10 GiB",
+                        "logical_volumes": [
+                            {
+                                # XXX: drop once https://github.com/osbuild/images/pull/1088 is merged
```
ondrejbudai: This can be dropped now.

test/testutil.py
```diff
@@ -159,6 +159,12 @@ def maybe_create_disk_customizations(cfg, tc):
                             "minsize": "10 GiB",
                             "fs_type": "xfs",
                             "mountpoint": "/",
+                        },
+                        {
+                            # XXX: drop once https://github.com/osbuild/images/pull/1088 is merged
```
ondrejbudai: Same here.

ondrejbudai (changes requested):
Looks great! I appreciate the simplification of the list of test cases, it's now much more straightforward. I found two minor issues inline.

test/test_manifest.py
mvo5: Indeed, noticed too and fixed now :)THANKS!

Issue #748, opened 2024-12-06
bib failed to generate raw image when config.json contains disk.swap customization


chunfuwen:
```
Running 'sudo podman run --rm -it --privileged --pull=newer --security-opt label=type:unconfined_t -v /var/lib/libvirt/images/output:/output -v /var/lib/libvirt/images/config.json:/config.json    -v /var/lib/libvirt/images/auth.json:/run/containers/0/auth.json   quay.io/centos-bootc/bootc-image-builder:latest  --type raw --tls-verify=false  --config /config.json  quay.io/fedora/fedora-bootc:latest  --rootfs ext4 '
[stderr] time="2024-12-05T22:05:30-05:00" level=warning msg="The input device is not a TTY. The --tty and --interactive flags might not work properly"
[stdout] Generating manifest manifest-raw.json
[stdout] Trying to pull quay.io/fedora/fedora-bootc:latest...
[stdout] Getting image source signatures
[stdout] Writing manifest to image destination
[stdout] DONE
[stdout] Building manifest-raw.json
[stdout] ^[[1m<stdin>^[[0m has ^[[1m^[[31merrors^[[0m:
[stdout]
[stdout] ^[[1m.pipelines[1].stages[2].options.volumes[1].size^[[0m:
[stdout]   '0B' does not match '[1-9][0-9]*[bBsSkKmMgGtTpPeE]?'
[stdout]
[stdout] 2024/12/06 03:07:08 error: cannot run osbuild: running osbuild failed: exit status 2

```
```
[root@dell-xx images]# cat config.json 
[...]

PR #749, opened 2024-12-06
test: add regression test/fix for gh748 swap space creation

mvo5:
This commit adds a regression test for the manifest creation when custom disk are used. When just adding a swap device the rootlv size is not set/updated correctly which leads to a manifest like:
```json
...
        {
          "type": "org.osbuild.lvm2.create",
          "options": {
            "volumes": [
              {
                "name": "swaplv",
                "size": "2147483648B"
              },
              {
                "name": "rootlv",
                "size": "0B"
              }
            ]
          },
...
```
for a config like:
```json
"customizations": {
            "disk": {
                "partitions": [
                    {
                        "type": "lvm",
                        "minsize": "10 GiB",
                        "logical_volumes": [
                            {
                                "minsize": "2 GiB",
                                "fs_type": "swap",
                            }
                        ]
                    }
                ]
            }
```
and indeed, we test that config in our unit tests (test_manifest_disk_customization_lvm_swa) but we do not run the full osbuild validation on it :(

Closes: https://github.com/osbuild/bootc-image-builder/issues/748

### DIFF
--- a/bib/cmd/bootc-image-builder/image.go
+++ b/bib/cmd/bootc-image-builder/image.go
@@ -16,6 +16,7 @@ import (
 	"github.com/osbuild/images/pkg/customizations/anaconda"
 	"github.com/osbuild/images/pkg/customizations/kickstart"
 	"github.com/osbuild/images/pkg/customizations/users"
+	"github.com/osbuild/images/pkg/datasizes"
 	"github.com/osbuild/images/pkg/disk"
 	"github.com/osbuild/images/pkg/image"
 	"github.com/osbuild/images/pkg/manifest"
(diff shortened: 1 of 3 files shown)


ondrejbudai (approved):
This is partly my fault. @achilleas-k originally enforced the disk size everywhere, but I told him that I would like if the `disk` library is stayed with as few hardcoded values as possible, so he extracted it outside.

Maybe the custom pt generation code in images should check this case?

Additionally, should we still enforce the min size of / of 2x the container image size as we do with the old partitioning? On the other hand, I can totally see cases when people don't want it. I actually fully expect some people not to ever run `bootc upgrade` because they will spin new instances and terminate the old ones. So they might not like the doubling...

mvo5:
> This is partly my fault. @achilleas-k originally enforced the disk size everywhere, but I told him that I would like if the `disk` library is stayed with as few hardcoded values as possible, so he extracted it outside.
> 
> Maybe the custom pt generation code in images should check this case?

I think from an API PoV this is sensible, however ideally we would error in the partitioning if we encounter a zero size partition (IMHO).

> Additionally, should we still enforce the min size of / of 2x the container image size as we do with the old partitioning? On the other hand, I can totally see cases when people don't want it. I actually fully expect some people not to ever run `bootc upgrade` because they will spin new instances and terminate the old ones. So they might not like the doubling...

This is a good point, we should probably fix that too, I think the issue is that for LVM the rules are different. IIRC when we create a disk with a minsize we will up the not-used space for "/". However with LVM we expect that (some) people will want to create  LVM partitions that have intentional space left so we do not auto-fill. So maybe this fix is actually wrong^Wnot ideal and indead it should put `RootfsMinsize` here instead of the hardcoded values. But then we have "/" and /"usr" and if we set both to rootfsMinsize (2x container size) we get a disk that is too big for cases where "/" and "/usr" are defined afaict.

achilleas-k:
[#748](https://github.com/osbuild/bootc-image-builder/pull/750#issuecomment-2523003531)

Same question.

ondrejbudai:
Addressed by #750

PR #750, opened 2024-12-06
bib: improve the "/" size calculation for LVM

mvo5:
This commit improves the rootfs size calculation for advanced
partitioning.  Just like for the tranditional filesystem
customization where we use `updateFilesystemSizes()` to ensure
that the rootfs is at least the two times size of the container
we need to do the same for the advanced disk partitioning.

The disk library will need a `RequiredMinSizes` parameter
(see gh#748) to know what sizes to pick. So far we did not
pass this which can leads to a "0B" root partition.

The naive fix was to just include a hardcoded:
```
requiredDirectorySizes = map[string]uint64{
        "/":    1 * datasizes.GiB,
        "/usr": 2 * datasizes.GiB,
}
```
but of course that is not ideal because we calculate the
size of disk based on 2x the container size. We set this
for the rootfs so the entire disk is big enough. However
because in advanced partitioning "/" is not automatically
expanded we end up with a potentially too small "/".

This commit tweaks the calculcation so that it sets the
requiredDirectorySizes for "/" to be at least 2x container
size.

Note that a custom "/usr" is not supported in image mode so splitting
rootfsMinSize between / and /usr is not a concern.

Closes: https://github.com/osbuild/bootc-image-builder/issues/748

### DIFF
--- a/bib/cmd/bootc-image-builder/image.go
+++ b/bib/cmd/bootc-image-builder/image.go
@@ -16,7 +16,6 @@ import (
 	"github.com/osbuild/images/pkg/customizations/anaconda"
 	"github.com/osbuild/images/pkg/customizations/kickstart"
 	"github.com/osbuild/images/pkg/customizations/users"
-	"github.com/osbuild/images/pkg/datasizes"
 	"github.com/osbuild/images/pkg/disk"
 	"github.com/osbuild/images/pkg/image"
 	"github.com/osbuild/images/pkg/manifest"
@@ -219,6 +218,44 @@ func genPartitionTable(c *ManifestConfig, customizations *blueprint.Customizatio
 	}
 }
 
+// calcRequiredDirectorySizes will calculate the minimum sizes for /
+// for disk customizations. We need this because with advanced partitioning
+// we never grow the rootfs to the size of the disk (unlike the tranditional
+// filesystem customizations).
+//
+// So we need to go over the customizations and ensure the min-size for "/"
+// is at least rootfsMinSize.
+//
+// Note that a custom "/usr" is not supported in image mode so splitting
+// rootfsMinSize between / and /usr is not a concern.
+func calcRequiredDirectorySizes(distCust *blueprint.DiskCustomization, rootfsMinSize uint64) (map[string]uint64, error) {
+	// XXX: this has *way* too much low-level knowledge about the
+	// inner workings of blueprint.DiskCustomizations plus when
+	// a new type it needs to get added here too, think about
+	// moving into "images" instead (at least partly)
+	mounts := map[string]uint64{}
+	for _, part := range distCust.Partitions {
+		switch part.Type {
+		case "", "plain":
+			mounts[part.Mountpoint] = part.MinSize
+		case "lvm":
+			for _, lv := range part.LogicalVolumes {
+				mounts[lv.Mountpoint] = part.MinSize
+			}
+		case "btrfs":
+			for _, subvol := range part.Subvolumes {
+				mounts[subvol.Mountpoint] = part.MinSize
+			}
+		default:
+			return nil, fmt.Errorf("unknown disk customization type %q", part.Type)
+		}
+	}
+	// ensure rootfsMinSize is respected
+	return map[string]uint64{
+		"/": max(rootfsMinSize, mounts["/"]),
+	}, nil
+}
+
 func genPartitionTableDiskCust(c *ManifestConfig, diskCust *blueprint.DiskCustomization, rng *rand.Rand) (*disk.PartitionTable, error) {
 	if err := diskCust.ValidateLayoutConstraints(); err != nil {
 		return nil, fmt.Errorf("cannot use disk customization: %w", err)
@@ -234,19 +271,16 @@ func genPartitionTableDiskCust(c *ManifestConfig, diskCust *blueprint.DiskCustom
 	if err != nil {
 		return nil, err
 	}
-
-	// XXX: copied from images, take from container instead?
-	requiredDirectorySizes := map[string]uint64{
-		"/":    1 * datasizes.GiB,
-		"/usr": 2 * datasizes.GiB,
+	requiredMinSizes, err := calcRequiredDirectorySizes(diskCust, c.RootfsMinsize)
+	if err != nil {
+		return nil, err
 	}
 	partOptions := &disk.CustomPartitionTableOptions{
 		PartitionTableType: basept.Type,
 		// XXX: not setting/defaults will fail to boot with btrfs/lvm
-		BootMode:      platform.BOOT_HYBRID,
-		DefaultFSType: defaultFSType,
-		// ensure sensible defaults for /,/usr
-		RequiredMinSizes: requiredDirectorySizes,
+		BootMode:         platform.BOOT_HYBRID,
+		DefaultFSType:    defaultFSType,
+		RequiredMinSizes: requiredMinSizes,
 	}
 	return disk.NewCustomPartitionTable(diskCust, partOptions, rng)
 }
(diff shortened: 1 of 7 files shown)


achilleas-k:
> Note that for users with a custom "/usr" but no "/" this may lead to
> large "/" but the users can tweak it easily by adding "/".

I wonder... does `/usr` actually work on a separate filesystem?  I know we don't block it in the policy, but is it a valid configuration?  (quick test shows it's probably not).

bib/cmd/bootc-image-builder/image.go
```diff
@@ -218,6 +218,35 @@ func genPartitionTable(c *ManifestConfig, customizations *blueprint.Customizatio
...
+	for _, part := range distCust.Partitions {
+		if part.Type == "lvm" {
+			for _, lv := range part.LogicalVolumes {
+				lvmMounts[lv.Mountpoint] = lv.MinSize
+			}
+		}
+	}
+	// no LVM used, nothing to do
+	if len(lvmMounts) == 0 {
+		return nil
```
ondrejbudai, line 242: We need to enforce the root size even on a plain partition, otherwise we get a too small root:

```toml
[[customizations.disk.partitions]]
mountpoint = "/var"
minsize = "20 GiB"
fs_type = "xfs"
```

results into:

```
⏱  Duration: 0s
org.osbuild.mkfs.xfs: f41472c20dedde704b2ba04b2acbab2b270c588ce5d6fcc19c47b0df9ad0f10e {
  "uuid": "00e77a6f-fb67-4525-80a8-111b8b0cc68d",
  "label": "root"
}
device/device (org.osbuild.loopback): loop0 acquired (locked: True)
size 251 of data subvolume is too small, minimum 4096 blocks
```

mvo5: Aha, indeed, I thought we would auto-grow the root for the non-lvm case but I remember we decided we do to support advanced use-cases where special partitions are created on first boot etc. I will make this more generic.

bib/cmd/bootc-image-builder/image_test.go
```diff
@@ -448,3 +449,236 @@ func TestGenPartitionTableDiskCustomizationRunsValidateLayoutConstraints(t *test
...
+			cus := &blueprint.Customizations{
+				Disk: &blueprint.DiskCustomization{
+					Partitions: tc.partitions,
+				},
+			}
+			pt, err := bib.GenPartitionTable(cnf, cus, rng)
+			assert.NoError(t, err)
+
+			var rootSize uint64
+			pt.ForEachMountable(func(mnt disk.Mountable, path []disk.Entity) error {
```
ondrejbudai, line 669: This line needs `_ =` to supress the linter. Or a proper error handling. :)

bib/cmd/bootc-image-builder/image.go
mvo5: This should be fixed now and I added a bunch of tests to ensure this is hopefully well covered now. I can build your example above now.

bib/cmd/bootc-image-builder/image_test.go
mvo5: Silly me, thank you

ondrejbudai (dismissed):
Thanks, this looks awesome! :)

bib/cmd/bootc-image-builder/image_test.go
```diff
@@ -448,3 +449,234 @@ func TestGenPartitionTableDiskCustomizationRunsValidateLayoutConstraints(t *test
...
+							rootSize = parent.GetSize()
+							break
+						}
+					}
+				}
+				return nil
+			})
+			assert.NoError(t, err)
+			// expected size is within a reasonable limit
+			assert.True(t, rootSize >= tc.expectedMinRootSize, rootSize < tc.expectedMinRootSize+500*datasizes.MiB)
```
mvo5, line 679: This can probably a bit too much,we can tweak this in a followup, but:
```go
-                       assert.True(t, rootSize >= tc.expectedMinRootSize, rootSize < tc.expectedMinRootSize+500*datasizes.MiB)
+                       assert.True(t, rootSize >= tc.expectedMinRootSize, rootSize < tc.expectedMinRootSize+2*datasizes.MiB)
```
or we are precise and do:
```go
$ git diff
diff --git a/bib/cmd/bootc-image-builder/image_test.go b/bib/cmd/bootc-image-builder/image_test.go
index c4416fd..2a9d59c 100644
--- a/bib/cmd/bootc-image-builder/image_test.go
+++ b/bib/cmd/bootc-image-builder/image_test.go
@@ -477,7 +477,7 @@ func TestGenPartitionTableDiskCustomizationSizes(t *testing.T) {
                        "empty disk customizaton, root expands to rootfsMinsize",
                        2 * datasizes.GiB,
                        nil,
-                       2 * datasizes.GiB,
+                       2*datasizes.GiB + 1031680,
                },
                // plain
                {
@@ -492,7 +492,7 @@ func TestGenPartitionTableDiskCustomizationSizes(t *testing.T) {
                                        },
                                },
                        },
-                       5 * datasizes.GiB,
+                       5*datasizes.GiB + 1031680,
                },
                {
                        "plain, small root minsize, expands to rootfsMnSize",
@@ -506,7 +506,7 @@ func TestGenPartitionTableDiskCustomizationSizes(t *testing.T) {
                                        },
                                },
                        },
-                       5 * datasizes.GiB,
+                       5*datasizes.GiB + 1031680,
                },
                {
                        "plain, big root minsize",
@@ -520,7 +520,7 @@ func TestGenPartitionTableDiskCustomizationSizes(t *testing.T) {
                                        },
                                },
                        },
-                       10 * datasizes.GiB,
+                       10*datasizes.GiB + 1031680,
                },
                // btrfs
                {
@@ -675,8 +675,7 @@ func TestGenPartitionTableDiskCustomizationSizes(t *testing.T) {
                                return nil
                        })
                        assert.NoError(t, err)
-                       // expected size is within a reasonable limit
-                       assert.True(t, rootSize >= tc.expectedMinRootSize, rootSize < tc.expectedMinRootSize+500*datasizes.MiB)
+                       assert.Equal(t, rootSize, tc.expectedMinRootSize)
                })
        }
 }

```
but I don't know where this number comes from?

mvo5:
The run failed with:
```
...
2024-12-06T17:32:50.2211320Z LVM2: using vg name '1d1a2221-7acf-4759-a066-9df765ea01ea'
2024-12-06T17:32:50.2211651Z   Physical volume "/dev/loop0" successfully created.
2024-12-06T17:32:50.2211982Z   Creating devices file /etc/lvm/devices/system.devices
2024-12-06T17:32:50.2212370Z   Volume group "1d1a2221-7acf-4759-a066-9df765ea01ea" successfully created
2024-12-06T17:32:50.2212971Z   WARNING: Logical volume 1d1a2221-7acf-4759-a066-9df765ea01ea/var_loglv not zeroed.
2024-12-06T17:32:50.2213355Z   Logical volume "var_loglv" created.
2024-12-06T17:32:50.2213708Z   WARNING: Logical volume 1d1a2221-7acf-4759-a066-9df765ea01ea/swaplv not zeroed.
2024-12-06T17:32:50.2214078Z   Logical volume "swaplv" created.
2024-12-06T17:32:50.2214374Z   Size is not a multiple of 512. Try using 3132294144 or 3132294656.
2024-12-06T17:32:50.2214692Z   Invalid argument for --size: 3132294152B
2024-12-06T17:32:50.2214955Z   Error during parsing of command line.
```

it seems we will need https://github.com/osbuild/images/pull/1092 too so that we correctly align when EnsureSize() is called becaue of the requiredMinSizes calculations.

ondrejbudai:
osbuild/images#1092 got released as 0.105.0, so I force-pushed a new go.mod. This should be now ready to go. :)

ondrejbudai (approved):
Looks good, but I would like a "cross-approval" since I was the last one who touched this. :)

mvo5:
> Looks good, but I would like a "cross-approval" since I was the last one who touched this. :)

Looks good to me as well, thank you!

PR #755, opened 2024-12-12
progress: auto-select progress based on $PS1

mvo5:
This commit adds automatic progress bar selection based on the
environment. Ideally we would check for a terminal but with
podman we never have one. So we check for PS1 and anf if that is
used we assume it's an interactive shell and give the nice
"terminalProgressBar". If that is not set we assuem we run in
a script or CI/CD environment and select plainProgressBar.


mvo5:
Silly me, this does not work

PR #757, opened 2024-12-15
Do not trigger push pipeline if only tekton files are updated

lmilbaum:
Do not trigger push pipeline if only tekton files are updated


lmilbaum (commented):
LGTM

ondrejbudai (approved):

PR #764, opened 2024-12-19
progress: fix missing "\n" in plain progress

mvo5:
This commit fixes the missing \n when plain progress is used. Without the output looks rather silly :/


ondrejbudai (approved):
Good catch, thanks!

PR #766, opened 2024-12-19
test: disable the fedora iso test for now

mvo5:
This commit disables the fedora ISO test for now. The download of the RPMs tends to be very fragile, often mirrors are outdated or broken and we get many 404 or 500 errors.

The real fix is of couse to switch to a librepo based downloading in osbuild but for now we just disable this test so that we can merge PRs again without having to retry multiple times :(

### DIFF
--- a/test/testcases.py
+++ b/test/testcases.py
@@ -87,7 +87,9 @@ def gen_testcases(what):  # pylint: disable=too-many-return-statements
         return [TestCaseC9S(image="ami"), TestCaseFedora(image="ami")]
     if what == "anaconda-iso":
         return [
-            TestCaseFedora(image="anaconda-iso", sign=True),
+            # 2024-12-19: disabled for now until the mirror situation becomes
+            # a bit more stable
+            # TestCaseFedora(image="anaconda-iso", sign=True),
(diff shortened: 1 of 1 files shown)


achilleas-k (approved):
Good call.  Fully agree.

mvo5:
Sorry had to force push for a silly linter error

achilleas-k (approved):

brad-jones:
I believe I'm running into this issue on my own custom bootc build.
Is there anyway I can influence the mirror(s) that it picks?

image

brad-jones:
I just did this in my oci build which seemed to force the issue for the `bootc-image-builder`.

```
RUN rm -f /etc/yum.repos.d/fedora-cisco-openh264.repo
RUN FILES=(/etc/yum.repos.d/fedora*.repo); sed --in-place \
    -e 's ^metalink= #metalink= ' \
    -e "s ^#baseurl=http://download.example/ baseurl=https://dl.fedoraproject.org/ " \
    "${FILES[@]}";
```

mvo5:
There is also https://github.com/osbuild/image-builder-cli/issues/40 now which describes a similar problem.

PR #776, opened 2025-01-06
main,README: improve docs around `--progress`

mvo5:
This PR adds some missing documentation around the new `--progress` switch. This hopefully helps with the questions/uncertainty raised in e.g. https://github.com/osbuild/bootc-image-builder/pull/765

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,7 @@ Usage:
 Flags:
       --chown string           chown the ouput directory to match the specified UID:GID
       --tls-verify             require HTTPS and verify certificates when contacting registries (default true)
+      --progress string        type of progress bar to use (e.g. plain,term)
       --type string            image type to build [qcow2, ami] (default "qcow2")
       --target-arch string     architecture to build image for (default is the native architecture)
 ```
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -137,6 +138,7 @@ Flags:
 | Argument          | Description                                                                                               | Default Value |
 |-------------------|-----------------------------------------------------------------------------------------------------------|:-------------:|
 | **--chown**       | chown the output directory to match the specified UID:GID                                                 |       ❌      |
+| --progress        | Show progress in the given format, supported: plain,term,debug. If empty it is auto-detected              |       ❌      |
```
achilleas-k, line 141: The rest of the arguments in this table are **bold**.  This should be too.

achilleas-k, line 141: Future: Should we have an `auto` value which is the explicit default?
(not very important)

achilleas-k (changes requested):

README.md
mvo5: I updated this and also included the suggesion from Ondrej to rename "plain" to "verbose"

achilleas-k: > we have this line saying `*💡 Tip: Flags in **bold** are the most important ones.*`

Oops, my bad.  In that case it makes more sense to have it non-bold and even de-bold some of the other ones.
Or even de-bold the whole lot?  None of these seem really **important** to me.

achilleas-k (approved):

README.md
mvo5: I can do in a followup, I guess --chown is also not one of the "important" ones but happy to discuss that in the PR :)

ondrejbudai (approved):
Love this, thanks! :)

mvo5:
Sorry, had to force push to fix a test failure, please re-review.

ondrejbudai (approved):
Thanks!

PR #779, opened 2025-01-06
build-sys: Bump to Fedora 41

cgwalters:
Fedora N-1 (40 now) is stable, but will eventually get stale. In my case we happen to be using this image as a "builder" image and I added a new feature to rpm-ostree, which I didn't ship to F40. I may still do so.

However it's clearly a good idea for us to keep updated.

This all said, it's actually not clear to me that Fedora is the right default base image - it may make sense to target e.g. c10s or c9s?

This all relates to https://gitlab.com/fedora/bootc/tracker/-/issues/2 as well.


mvo5 (approved):
Thank you! I think we should merge this as it will benefit us right away.

And then also have the discussion about what base to use with e.g. @ondrejbudai and @achilleas-k - I would not mind moving to c9s or c10s but I'm not sure I have the full picture of the ramifications.

PR #786, opened 2025-01-08
main: add new  `--use-librepo` switch

mvo5:
This commit adds a new `--use-librepo` switch that will
enable librepo based downloading so that people can play with
the new backend.

Note that --use-librepo defaults to False currently until we have sufficient testing that all cases (including when secrets are required) work.

(with that we can re-enable the fedora iso test-case as it should be no longer flaky and works fine in my local test. however it will increase the test run times again as the isos are pretty "expensive")


ondrejbudai (approved):
This is awesome, thank you!

achilleas-k (approved):
Ship it!

PR #788, opened 2025-01-08
Makefile: implement `make help`

schuellerf:
Implement `make help` to be more developer friendly


mvo5 (approved):
It looks like this slipped through the cracks, sorry for that.

LGTM

PR #789, opened 2025-01-09
README: Clarify optional SELinux dependency

ochosi:
SELinux may not be enforced, e.g. as is the case on Debian.

Ported over from https://github.com/osbuild/osbuild.github.io/pull/138

Please note that I'm not the original author of this suggestion, I merely wanted to help @egodigitus by re-opening the change in the correct place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ $ podman machine start
 
 ## ✍ Prerequisites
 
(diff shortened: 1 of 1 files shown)


mvo5 (approved):
Thank you!

PR #790, opened 2025-01-09
main: add --verbose,-v persistent flag that increases  verbosity

mvo5:
This commit adds `--verbose,-v` which will increase the verbosity
of logrus and also switch the --progress to "verbose". This is
addressing the feedback we got in
https://github.com/osbuild/bootc-image-builder/pull/765
and a followup for #776

The new `-v` clashes unfortunately with cobras default for version,
so there is no single dash flag for version anymore. Most unix tools
(e.g. cp,rsync,mv,curl,ssh,tar) use "-v" for "--verbose" so IMHO we
should follow suite. Unfortuantely there is no consistency in linux,
e.g. git,gcc are counter examples where it means version). I would
still go with -v for verbose as ssh,tar,curl are probably used
more often to get verbose output.

---

main,test: tweak cmdVersion

This commit tweak the new and very nice functionality of cmdVersion (see commit message for details)

---

(note that the above commit is not strictly needed, because of the slightly inflexible way of how cobra handles the version I had to look at this function and did some quick tweaks but I could drop it from this PR and/or drop it entirely)

Thanks to Ondrej for the suggestion about -v/--verbose

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,9 @@ Flags:
       --progress string        type of progress bar to use (e.g. verbose, term)
       --type string            image type to build [qcow2, ami] (default "qcow2")
       --target-arch string     architecture to build image for (default is the native architecture)
+
+Global Flags:
+  -v, --verbose            Switch to verbose mode
 ```
 
 ### Detailed description of optional flags
@@ -143,6 +146,7 @@ Flags:
 | **--tls-verify**  | Require HTTPS and verify certificates when contacting registries                                          |    `true`     |
 | **--type**        | [Image type](#-image-types) to build                                                                      |    `qcow2`    |
 | **--target-arch** | [Target arch](#-target-architecture) to build                                                             |       ❌      |
+| **--verbose**     | Switch output/progress to verbose mode    |   `false`   |
 
 The `--type` parameter can be given multiple times and multiple outputs will
 be produced.
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -546,17 +546,25 @@ func chownR(path string, chown string) error {
 var rootLogLevel string
 
 func rootPreRunE(cmd *cobra.Command, _ []string) error {
-	if rootLogLevel == "" {
+	verbose, _ := cmd.Flags().GetBool("verbose")
+	progress, _ := cmd.Flags().GetString("progress")
+	switch {
+	case rootLogLevel != "":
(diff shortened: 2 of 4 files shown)


ondrejbudai (changes requested):
Nice cleanups. The ambiguity of `-v` across the linux stack is sad, but I agree that having it for verbose, and then `--version` and `version` for displaying the version info is probably the better choice (100% subjective ofc). We should make sure that image-builder-cli uses the same convention.

I requested changes, because `-v` should be documented in the README. ;)

mvo5:
> Nice cleanups. The ambiguity of `-v` across the linux stack is sad, but I agree that having it for verbose, and then `--version` and `version` for displaying the version info is probably the better choice (100% subjective ofc). We should make sure that image-builder-cli uses the same convention.
> 
> I requested changes, because `-v` should be documented in the README. ;)

Thanks, I added it now to the README, we should probably do a followup as well that documents more options in the README, its a bit out of sync with reality. I can look after this one here is merged.

chunfuwen:
verified by below command, it remind that error: unknown flag: --verbose
```
sudo podman run --rm -it --privileged --pull=newer --security-opt label=type:unconfined_t -v /var/lib/libvirt/images/output:/output -v /var/lib/libvirt/images/config.json:/config.json    -v /var/lib/libvirt/images/auth.json:/run/containers/0/auth.json   quay.io/centos-bootc/bootc-image-builder:latest  --type qcow2 --tls-verify=false  --config /config.json  --verbose  quay.io/centos-bootc/centos-bootc:stream9
```
But using "-v " command ,it output below:
```
 sudo podman run --rm -it --privileged --pull=newer --security-opt label=type:unconfined_t -v /var/lib/libvirt/images/output:/output -v /var/lib/libvirt/images/config.json:/config.json    -v /var/lib/libvirt/images/auth.json:/run/containers/0/auth.json   quay.io/centos-bootc/bootc-image-builder:latest  --type qcow2 --tls-verify=false  --config /config.json  -v quay.io/centos-bootc/centos-bootc:stream9 
Trying to pull quay.io/centos-bootc/bootc-image-builder:latest...
Getting image source signatures
Copying blob 22358cc5ee9a done   | 
Copying blob e554e6faffe5 done   | 
Copying blob 2ebe41b89b1d done   | 
Copying config bfc3eaae84 done   | 
Writing manifest to image destination
build_revision: 1754664
build_time: 2025-01-08T19:09:35Z
build_status: tainted

```

PR #791, opened 2025-01-10
README.md: update supported flags table

mvo5:
This commit updates the documentation around the supported flags. It also switches on "bold" for a smaller subset of the important flags.

Note that still not all flags are shows, some are documented below (like the `--aws-*`) and some (like --store, --rpmmd) require more context, i.e. explaining that they require calling podman with the right `--volume` mappings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,13 +126,17 @@ Usage:
     <imgref>
 
 Flags:
-      --chown string           chown the ouput directory to match the specified UID:GID
-      --tls-verify             require HTTPS and verify certificates when contacting registries (default true)
-      --progress string        type of progress bar to use (e.g. verbose, term)
-      --type string            image type to build [qcow2, ami] (default "qcow2")
-      --target-arch string     architecture to build image for (default is the native architecture)
+      --chown string          chown the ouput directory to match the specified UID:GID
+      --output string         artifact output directory (default ".")
+      --progress string       type of progress bar to use (e.g. verbose,term) (default "auto")
+      --rootfs string         Root filesystem type. If not given, the default configured in the source container image is used.
+      --target-arch string    build for the given target architecture (experimental)
+      --tls-verify            require HTTPS and verify certificates when contacting registries (default true)
+      --type stringArray      image types to build [ami, anaconda-iso, gce, iso, qcow2, raw, vhd, vmdk] (default [qcow2])
+      --version               version for bootc-image-builder
 
 Global Flags:
+      --log-level string   logging level (debug, info, error); default error
   -v, --verbose            Switch to verbose mode
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -126,27 +126,33 @@ Usage:
...
+| --progress        | Show progress in the given format, supported: verbose,term,debug. If empty it is auto-detected            |     `auto`    |
 | **--rootfs**      | Root filesystem type. Overrides the default from the source container. Supported values: ext4, xfs, btrfs |       ❌      |
-| **--tls-verify**  | Require HTTPS and verify certificates when contacting registries                                          |    `true`     |
-| **--type**        | [Image type](#-image-types) to build                                                                      |    `qcow2`    |
-| **--target-arch** | [Target arch](#-target-architecture) to build                                                             |       ❌      |
-| **--verbose**     | Switch output/progress to verbose mode    |   `false`   |
+| --tls-verify      | Require HTTPS and verify certificates when contacting registries                                          |     `true`    |
+| **--type**        | [Image type](#-image-types) to build (can be passed multiple times)                                       |     `qcow2`   |
```
p5, line 152: There's a mix of information on how `--type` should be passed in.  One place says it can be passed in multiple times (`--type qcow2 --type raw`), and another says to use a string array (`--type [qcow2,raw]`).  

Even if both are supported, I think it would be good to standardise on which is recommended to avoid confusion.

README.md
```diff
@@ -126,27 +126,33 @@ Usage:
...
-      --type string            image type to build [qcow2, ami] (default "qcow2")
-      --target-arch string     architecture to build image for (default is the native architecture)
+      --chown string          chown the ouput directory to match the specified UID:GID
+      --output string         artifact output directory (default ".")
+      --progress string       type of progress bar to use (e.g. verbose,term) (default "auto")
+      --rootfs string         Root filesystem type. If not given, the default configured in the source container image is used.
+      --target-arch string    build for the given target architecture (experimental)
+      --tls-verify            require HTTPS and verify certificates when contacting registries (default true)
+      --type stringArray      image types to build [ami, anaconda-iso, gce, iso, qcow2, raw, vhd, vmdk] (default [qcow2])
+      --version               version for build
```
p5, line 136:
```suggestion
      --version               version for bootc-image-builder
```

"version for build" is slightly confusing because I'm not sure whether it's the version of _my_ image build or the BIB image.

p5 (commented):
Hope you don't mind a drive-by review :) 

I've spent quite a lot of time in these docs over the past week, so thought I could be helpful.

ondrejbudai (commented):
Looks good. I agree with @p5's concerns. :)

README.md
mvo5: Good catch, thank you! I fixed this now in the README. Fixing this in the code is a bit more work as we need to workaround cobra here.

README.md
supakeen: Let's start here and turn the above comment into an issue (I'll click the button).

supakeen (approved):

PR #799, opened 2025-01-16
plan: give testingfarm more time to complete

mvo5:
We recently hit the issue that bootc-image-builder is no longer releasing. It turns out we hit the limit of our integration tests. Increase this limit for now.


achilleas-k (approved):
Thanks for looking into this.  I know it wasn't easy.

PR #804, opened 2025-01-21
progress: move from an internal package to external

mvo5:
This commit makes the previously internal `progress` package an external API. The is the result of the discussion in images PR#1150 where we decided that the progress module is not a great fit for the "images" library.

We want to share this code between bootc-image-builder and image-builder-cli now. In the future we will also want to use it in the `worker-executor` in `osbuild-composer` to parse the stream data from `osbuild`.

We plan to merge bootc-image-builder and image-builder-cli medium term so importing code from bootc-image-buider in image-builder cli is not that stange.

When we (longer-term) use this code the `worker-executor` we will need to think about this again and maybe put it back into images. However this commit unblocks us without making anything worse.


achilleas-k (approved):
Nice and simple.

I realise it's weird to have ib-cli import bib for this, but I'm fine with unblocking this work while we discuss what should go where and tidying up later.

supakeen (approved):
Nice (and temporary while we think about the rest).

supakeen:
This will fail with the current `osbuild` + `images` main until https://github.com/osbuild/images/pull/1162 and https://github.com/osbuild/osbuild/pull/1990 are in.

PR #811, opened 2025-01-27
progress: add new `BuildLog` option

mvo5:
This commit adds a new `BuildLog` option to the `OSBuildOptions`
that can be used to generate a streamed buildlog (e.g. to a file
or a websocket).

### DIFF
--- a/bib/pkg/progress/progress.go
+++ b/bib/pkg/progress/progress.go
@@ -317,8 +317,18 @@ func (b *debugProgressBar) SetProgress(subLevel int, msg string, done int, total
 	return nil
 }
 
+type OSBuildOptions struct {
+	StoreDir  string
+	OutputDir string
+	ExtraEnv  []string
+}
+
 // XXX: merge variant back into images/pkg/osbuild/osbuild-exec.go
-func RunOSBuild(pb ProgressBar, manifest []byte, store, outputDirectory string, exports, extraEnv []string) error {
+func RunOSBuild(pb ProgressBar, manifest []byte, exports []string, opts *OSBuildOptions) error {
+	if opts == nil {
+		opts = &OSBuildOptions{}
+	}
+
 	// To keep maximum compatibility keep the old behavior to run osbuild
 	// directly and show all messages unless we have a "real" progress bar.
 	//
@@ -328,20 +338,20 @@ func RunOSBuild(pb ProgressBar, manifest []byte, store, outputDirectory string, 
 	// just run with the new runOSBuildWithProgress() helper.
 	switch pb.(type) {
 	case *terminalProgressBar, *debugProgressBar:
-		return runOSBuildWithProgress(pb, manifest, store, outputDirectory, exports, extraEnv)
+		return runOSBuildWithProgress(pb, manifest, exports, opts)
 	default:
-		return runOSBuildNoProgress(pb, manifest, store, outputDirectory, exports, extraEnv)
+		return runOSBuildNoProgress(pb, manifest, exports, opts)
 	}
 }
 
-func runOSBuildNoProgress(pb ProgressBar, manifest []byte, store, outputDirectory string, exports, extraEnv []string) error {
-	_, err := osbuild.RunOSBuild(manifest, store, outputDirectory, exports, nil, extraEnv, false, os.Stderr)
+func runOSBuildNoProgress(pb ProgressBar, manifest []byte, exports []string, opts *OSBuildOptions) error {
+	_, err := osbuild.RunOSBuild(manifest, opts.StoreDir, opts.OutputDir, exports, nil, opts.ExtraEnv, false, os.Stderr)
 	return err
 }
 
 var osbuildCmd = "osbuild"
 
-func runOSBuildWithProgress(pb ProgressBar, manifest []byte, store, outputDirectory string, exports, extraEnv []string) (err error) {
+func runOSBuildWithProgress(pb ProgressBar, manifest []byte, exports []string, opts *OSBuildOptions) (err error) {
 	rp, wp, err := os.Pipe()
 	if err != nil {
 		return fmt.Errorf("cannot create pipe for osbuild: %w", err)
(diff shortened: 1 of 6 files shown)


bib/pkg/progress/progress.go
```diff
@@ -311,29 +312,64 @@ func (b *debugProgressBar) SetProgress(subLevel int, msg string, done int, total
...
-	return err
+	// There is a slight wrinkle here: when requesting a buildlog
+	// we can no longer write to separate stdout/stderr streams
+	// without being racy and give potential out-of-order output
+	// (which is very bad and confusing in a log). The reason is
+	// that if cmd.Std{out,err} are different "go" will start two
+	// go-routine to monitor/copy those are racy when both stdout,stderr
+	// output happens close together (TestRunOSBuildWithBuildlog demos
+	// that). We cannot have our cake and eat it so here we need to
+	// combine osbuilds stderr into our stdout.
```
supakeen, line 355: Just a question but could we turn off buffering and how both goros write to a queue that we then write into the log or are there bears on that road too?

mvo5: I tried a few strategies, including writing a syncedMultiWriter that uses a mutex to ensure ordering but I couldn't get it race-free and the test would show out of sync output every 1/4 or of the time on my system. It is probably worth another attempt but I'm not super hopeful. I can think more about it.

supakeen: We can skip this bit for now since there's something working, right? :)

mvo5: Yes, it does work reliably and arguably redirecting stderr from osbuild directly to the users stderr is a bit of a bug anyway so I'm okay with this. Plus for the "verbose" case the old behavior is kept (no old scripts or integrations that depend on stderr will break)

supakeen (approved):

bib/pkg/progress/progress.go
```diff
@@ -311,31 +312,64 @@ func (b *debugProgressBar) SetProgress(subLevel int, msg string, done int, total
...
+		osbuildStdout = osStdout
 	default:
-		return runOSBuildNoProgress(pb, manifest, store, outputDirectory, exports, extraEnv)
+		// hide the direct osbuild output by default
+		osbuildStdout = io.Discard
 	}
-}
 
-func runOSBuildNoProgress(pb ProgressBar, manifest []byte, store, outputDirectory string, exports, extraEnv []string) error {
-	_, err := osbuild.RunOSBuild(manifest, store, outputDirectory, exports, nil, extraEnv, false, os.Stderr)
```
mvo5, line 332: This cannot go away just yet for verbose output this is still needed because osbuild needs to be run with the "LogMontior" instead of the "JSONSeqMontior" if run like this :/

mvo5: There is another wrinkle here - we can get buildlog messages from both the jsonseq montior but also from e.g.  a crashing osbuild so we need to write both the "Trace" messages to the build log and also all stdout/stderr output.

mvo5: This should be fixed now.

bib/pkg/progress/progress_test.go
```diff
@@ -175,7 +175,66 @@ func TestRunOSBuildWithProgressIncorrectJSON(t *testing.T) {
...
+	err = progress.RunOSBuild(pbar, []byte(`{"fake":"manifest"}`), nil, nil)
 	assert.EqualError(t, err, `errors parsing osbuild status:
 cannot scan line "invalid-json": invalid character 'i' looking for beginning of value`)
 }
+
+func TestRunOSBuildWithBuildlogTerm(t *testing.T) {
+	restore := progress.MockOsbuildCmd(makeFakeOsbuild(t, `
+echo osbuild-stdout-output
+>&2 echo osbuild-stderr-output
+
```
mvo5, line 187: This needs something like a `sleep 0.1` because
``` 
# XXX: without the sleep this is racy as two different go routines poll
# this does not matter (much) in practise because osbuild output and
# stage output are using the syncedMultiWriter so output is not garbled
sleep 0.1
```

supakeen (approved):
Not terribly excited about the stderr/stdout thing but it's also understandable.

bib/pkg/progress/progress.go
```diff
@@ -316,8 +320,21 @@ func (b *debugProgressBar) SetProgress(subLevel int, msg string, done int, total
...
+	// BuildLog writes the osbuild output to the given writer
+	BuildLog io.Writer
+}
+
 // XXX: merge variant back into images/pkg/osbuild/osbuild-exec.go
-func RunOSBuild(pb ProgressBar, manifest []byte, store, outputDirectory string, exports, extraEnv []string) error {
+func RunOSBuild(pb ProgressBar, manifest []byte, exports []string, opts *OSBuildOptions) error {
+	if opts == nil {
+		opts = &OSBuildOptions{}
+	}
```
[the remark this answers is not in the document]
mvo5: Indeed, we should probably merge the two. Note that this struct here also defines the env and the buildlog and it seems(?) nicer to keep them in a single struct as they are all related to how to drive osbuild. I would really like to move this part of the code into images proper (well, all of progress, maybe just as an interface or something if the actual implementations feel too out of place).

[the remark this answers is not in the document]
mvo5: We should sync about this, if we cleanup the osbuild.RunOSBuild() signature, maybe we can do even more?
Strawman:
```go
func Run(manifest []byte, clopts RunOptions) (*Result, error) {
```
and put ` extraEnv []string, result bool, errorWriter io.Writer` into the RunOptions as well? (and `osbuild.Run()` has a nice ring to it :)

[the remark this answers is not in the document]
achilleas-k: I'm open to discussing it.  I wanted to put the options only on the struct to separate them from other stuff, like I mention in the linked commit message.  `env` fits too, as does result, but I don't know about the `errorWriter`.  I could be convinced :)

mvo5:
@supakeen Sorry, had to force push to resolve conflicts.

bib/pkg/progress/progress.go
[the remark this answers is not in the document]
mvo5: Thanks! Yeah, lets brainstorm this a bit, Given that we will need to modify the callers(compose etc) once we modify `osbuild-exec.go` we should take a bit more care. Are you okay with keeping this here for now as a testbed or would you rather unify now?

supakeen (approved):

bib/pkg/progress/progress.go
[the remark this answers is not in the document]
achilleas-k: Yup.  Ship it!

achilleas-k (approved):

PR #812, opened 2025-01-27
progress: set --cache-max-size in osbuild

mvo5:
This commit allows controlling the `osbuild --cache-max-size`
option. By default it will set the cache to 20GiB but allows
this to be controlled by the user.

Thanks to Simon for raising this.

### DIFF
--- a/bib/pkg/progress/progress.go
+++ b/bib/pkg/progress/progress.go
@@ -16,6 +16,7 @@ import (
 	"github.com/mattn/go-isatty"
 	"github.com/sirupsen/logrus"
 
+	"github.com/osbuild/images/pkg/datasizes"
 	"github.com/osbuild/images/pkg/osbuild"
 )
 
@@ -328,6 +329,8 @@ type OSBuildOptions struct {
 
 	// BuildLog writes the osbuild output to the given writer
 	BuildLog io.Writer
+
+	CacheMaxSize int64
 }
 
 // XXX: merge variant back into images/pkg/osbuild/osbuild-exec.go
@@ -349,6 +352,26 @@ func RunOSBuild(pb ProgressBar, manifest []byte, exports []string, opts *OSBuild
 	default:
 		return runOSBuildNoProgress(pb, manifest, exports, opts)
 	}
+}
+
+func newOsbuildCmd(manifest []byte, exports []string, opts *OSBuildOptions) *exec.Cmd {
+	cacheMaxSize := int64(20 * datasizes.GiB)
+	if opts.CacheMaxSize != 0 {
+		cacheMaxSize = opts.CacheMaxSize
+	}
+	cmd := exec.Command(
+		osbuildCmd,
+		"--store", opts.StoreDir,
+		"--output-directory", opts.OutputDir,
+		fmt.Sprintf("--cache-max-size=%v", cacheMaxSize),
+		"-",
+	)
+	for _, export := range exports {
+		cmd.Args = append(cmd.Args, "--export", export)
+	}
+	cmd.Env = append(os.Environ(), opts.ExtraEnv...)
+	cmd.Stdin = bytes.NewBuffer(manifest)
+	return cmd
 }
 
 func runOSBuildNoProgress(pb ProgressBar, manifest []byte, exports []string, opts *OSBuildOptions) error {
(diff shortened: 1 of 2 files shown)


supakeen (dismissed):
Happy about this; a quick note on the size. A single Fedora Minimal build takes up about 10 GiB of space so 20 GiB is indeed a decent minimum.

Note also that due to how we setup pipelines right now basically no output gets cached (at least in in the `image-builder`-case, I didn't verify the `bootc-image-builder` case yet).

achilleas-k:
Without checkpoints, osbuild will only cache sources.

supakeen:
> > I didn't verify the `bootc-image-builder` case yet
> 
> https://github.com/osbuild/bootc-image-builder/blob/9270d2e2b42b71ed55ad8b004580bfaaae09c197/bib/pkg/progress/progress.go#L332
> 
> No checkpoints there.
> 
> Actually, we never set any checkpoints in osbuild-composer or bootc-image-builder (so far).

Correct; it would be 'OK' to pass `--checkpoint="*"`, something we implemented a while back I think. However; as I've said previously as well with the current way we generate pipelines nothing will actually get cached post the buildroot.

Perhaps it'd be fine to then only pass `--checkpoint="buildroot"` for things that are generated by images?

achilleas-k:
> with the current way we generate pipelines nothing will actually get cached post the buildroot even when passing checkpoints.

Why wouldn't they be cached?  Do you mean because our build pipeline is dynamic it wont be reused or that it actually wont be cached?

Even with dynamic build packages, there are a lot of cases where a different build wont affect the build packages.  Rebuilding the same image type (or a very similar one) with a different blueprint, depending on customizations, might or might not affect build packages.  Introducing customizations often adds a package, but changing values might not.  Neither does adding a blueprint package.

Going back to a fat build root isn't impossible or out of the question though.  The reason we never had checkpoints in osbuild-composer and the service so far was lack of cleanup, which @mvo5 added [a while ago](https://github.com/osbuild/osbuild/pull/1503).

supakeen:
> > with the current way we generate pipelines nothing will actually get cached post the buildroot even when passing checkpoints.
> 
> Why wouldn't they be cached? Do you mean because our build pipeline is dynamic it wont be reused or that it actually wont be cached?
> 
> Even with dynamic build packages, there are a lot of cases where a different build wont affect the build packages. Rebuilding the same image type (or a very similar one) with a different blueprint, depending on customizations, might or might not affect build packages. Introducing customizations often adds a package, but changing values might not. Neither does adding a blueprint package.
> 
> Going back to a fat build root isn't impossible or out of the question though. The reason we never had checkpoints in osbuild-composer and the service so far was lack of cleanup, which @mvo5 added [a while ago](https://github.com/osbuild/osbuild/pull/1503).

The `build` pipeline can be cached as long as it's the same one twice (or multiple can be in the cache). However, the `os` pipeline contains UUIDs which are different for every build invalidating it, and all pipelines dependent on it.

achilleas-k:
> The `build` pipeline can be cached as long as it's the same one twice (or multiple can be in the cache). However, the `os` pipeline contains UUIDs which are different for every build invalidating it, and all pipelines dependent on it.

aaah of course.  Yes, sorry for the nonsense then.  I was just focused on the build pipeline.

supakeen (approved):

achilleas-k (approved):

achilleas-k:
```
osbuild [-h] [--cache DIRECTORY] [-l DIRECTORY] [--cache-max-size SIZE]\n               [--checkpoint ID] [--export ID] [--json]\n               [--output-directory DIRECTORY] [--inspect] [--monitor NAME]\n               [--monitor-fd FD] [--stage-timeout STAGE_TIMEOUT] [--version]\n               [--break [DEBUG_BREAK]] [--quiet]\n               MANIFEST\nosbuild: error: the following arguments are required: MANIFEST
```

bib/pkg/progress/progress.go
```diff
@@ -351,6 +354,25 @@ func RunOSBuild(pb ProgressBar, manifest []byte, exports []string, opts *OSBuild
...
+		"--output-directory", opts.OutputDir,
+		fmt.Sprintf("--cache-max-size=%v", cacheMaxSize),
+	)
+	for _, export := range exports {
+		cmd.Args = append(cmd.Args, "--export", export)
+	}
+	cmd.Env = append(os.Environ(), opts.ExtraEnv...)
+	cmd.Stdin = bytes.NewBuffer(manifest)
+	return cmd
+}
```
achilleas-k, line 374: osbuild needs an explicit `-` arg to read the manifest from stdin.

achilleas-k (changes requested):

bib/pkg/progress/progress.go
mvo5: Thanks for spotting this! Fixed now

supakeen (approved):

achilleas-k (approved):

supakeen (changes requested):
I will reapply this in `image-builder` as this PR targets files that have migrated there.

mvo5:
Yeah, lets close it here and reopen in ibcli.

PR #818, opened 2025-02-04
progress: tweak how `os.Stderr` is mocked

mvo5:
This commit changes the way `os.Stderr` is mocked so that higher level consumes of the libary can use helpers can replace os.Stderr (like `testutil.CaptureStdio()` is doing). The existing approach assigns the "real" os.Stderr to osStderr so early that it cannot be changed later.

This will allow the test outlined in https://github.com/osbuild/image-builder-cli/pull/124


ondrejbudai (approved):
Awesome, thanks!

PR #825, opened 2025-02-06
distrodef: fix incorrect detection of `centos-10` vs `10.0`

mvo5:
This commit fixes the issue that when the `os-release` has an
`ID=10.0` in it, a distrodef file ending with `-10` will not
be identified as the closest match because the code is using
a "strictly bigger" relation to sort versions but for semver
`10.0` and `10` are just identical so we need ">=" instead
of ">" (see the actual diff).

We saw this issue with the `rhel-10.0` beta that has a
`VERSION_ID=10.0` field and a `rhel-10.yaml` distro-def. Here the
code was skipping over rhel-10.yaml and felt back to rhel-9 when
it really should have used `rhel-10.yaml`.

Thanks to Achilleas for tracking this down.

### DIFF
--- a/bib/internal/distrodef/distrodef.go
+++ b/bib/internal/distrodef/distrodef.go
@@ -51,7 +51,7 @@ func findDistroDef(defDirs []string, distro, wantedVerStr string) (string, error
 			if err != nil {
 				return "", fmt.Errorf("cannot parse distro version from %q: %w", m, err)
 			}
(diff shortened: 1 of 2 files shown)


achilleas-k (approved):
LGTM but there's a typo in the commit message.

mvo5:
> LGTM but there's a typo in the commit message.

What is the typo?

achilleas-k (approved):

PR #830, opened 2025-02-10
github/workflows/tests: fix linter by adapting versions

schuellerf:
For some reason the linter fails but also the golang version should be consistent with `go.mod`


schuellerf:
The linter seems to be broken since
https://github.com/osbuild/bootc-image-builder/actions/runs/13236086119/job/36941149060?pr=828

mvo5 (approved):

Issue #834, opened 2025-02-14
[RFE] Inject files and directories during conversion


ericzolf:
I want to inject files (secrets, which I don't want to have in the container image) during the conversion phase (i.e. with bootc-image-builder). I thought using the blueprint, but according to https://osbuild.org/docs/user-guide/blueprint-reference/#files-and-directories those options are not supported by bootc. So I'm creating this issue to have this feature or equivalent one.

PR #841, opened 2025-02-17
bib: add support for file/directory customizations

mvo5:
This commit adds support for files/directories in blueprint
customizations.

This needs https://github.com/osbuild/images/pull/1227

Closes: https://github.com/osbuild/bootc-image-builder/issues/834

[edit this will also need an update in https://osbuild.org/docs/user-guide/blueprint-reference/#files-and-directories once merged]


achilleas-k (approved):

supakeen (approved):
Looks OK to me. We should probably (when documenting) point out the specific use case for this.

We should also update the documentation with *actual* allowed file paths for both package-based and image-based image types. They're out of sync right now and we don't differentiate.

PR #842, opened 2025-02-20
 go.mod: update osbuild/images to v0.118.0

achilleas-k:
Replace ifcfg with net-lib for EL10 and F42 installers.

 In RHEL 10 and Fedora 42, the ifcfg module was replaced by net-lib.


ondrejbudai (approved):
Thanks! :)

mvo5 (approved):

PR #549, opened 2024-07-18
Switch to BurntSushi/toml and error out on unknown toml keys

ondrejbudai:
This pull request includes two changes. First, it switches from `pelletier/go-toml` to `BurntSushi/toml` in the `bib` module. This change removes a dependency and aligns with the use of `BurntSushi` elsewhere in the codebase. Second, it adds error handling for unknown toml keys in the `buildconfig` module. This change improves clarity and prevents confusion when using older versions of `bib` with recent customizations.

PR #860, opened 2025-03-20
Enable strict toml checks again

mvo5:
This PR enables https://github.com/osbuild/bootc-image-builder/pull/549 again. 

With the new version of burntSushi/toml we have a fix for https://github.com/burntSushi/toml/issues/425 so we can enable strict checking again.


ondrejbudai (approved):
Awesome, love it! ❤️ 

We should also enable it in `image-builder-cli` I think.

PR #890, opened 2025-04-09
Revert "build(deps): bump golangci/golangci-lint-action from 6 to 7"

schuellerf:
This reverts commit 1c3d2dab0dae47e2c50bba2fcfa9da9d5d592d8a to fix CI.
golangci-lint-action v7 needs golangci v2.
Changing from v1 to v2 needs more changes than this and will be done separately.
see https://github.com/golangci/golangci-lint-action?tab=readme-ov-file#compatibility


schuellerf:
We'll continue discussion a better solution in #892 as the new linter should work…

PR #891, opened 2025-04-09
github: bump golangci-lint version

schuellerf:
alternative solution to #890

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,7 +15,7 @@ env:
   GO_VERSION: 1.22
   # see https://golangci-lint.run/product/changelog
   # to select a version that supports the GO_VERSION given above
-  GOLANGCI_LINT_VERSION: v1.59.1
+  GOLANGCI_LINT_VERSION: v2.0.2
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,9 +13,6 @@ on:
 
 env:
   GO_VERSION: 1.22
-  # see https://golangci-lint.run/product/changelog
-  # to select a version that supports the GO_VERSION given above
-  GOLANGCI_LINT_VERSION: v2.0.2
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
@@ -44,10 +41,14 @@ jobs:
       - name: Install libgpgme devel package
         run: sudo apt install -y libgpgme-dev libbtrfs-dev libdevmapper-dev
 
+      - name: Extract golangci-lint version from Makefile
+        id: golangci_lint_version
+        run: echo "GOLANGCI_LINT_VERSION=$(awk -F '=' '/^GOLANGCI_LINT_VERSION *=/{print $2}' Makefile)" >> "$GITHUB_OUTPUT"
+
       - name: Run golangci-lint
         uses: golangci/golangci-lint-action@v7
         with:
-          version: ${{ env.GOLANGCI_LINT_VERSION }}
+          version: ${{ steps.golangci_lint_version.outputs.GOLANGCI_LINT_VERSION }}
           args: --timeout 5m0s
           working-directory: bib
 
--- a/bib/cmd/bootc-image-builder/cloud.go
+++ b/bib/cmd/bootc-image-builder/cloud.go
@@ -8,6 +8,7 @@ import (
 	"github.com/cheggaaa/pb/v3"
 	"github.com/spf13/pflag"
 
+	"github.com/osbuild/bootc-image-builder/bib/internal/utils"
 	"github.com/osbuild/images/pkg/cloud"
 )
 
@@ -29,7 +30,7 @@ func upload(uploader cloud.Uploader, path string, flags *pflag.FlagSet) error {
 	if err != nil {
 		return fmt.Errorf("cannot upload: %v", err)
 	}
-	defer file.Close()
+	defer utils.LogClose(file)
 
 	var r io.Reader = file
 	if pbar != nil {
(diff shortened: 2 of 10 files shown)


achilleas-k:
Agree with this.  Let's get it updated and address the new issues it detects.

mvo5 (approved):
Thanks for the update! Yeah, lets merge and fix the errors, I can push the fixes to this branch or we do it in a followup, no strong opinion

schuellerf:
Actually I don't know why this works in the github action with go 1.22, where golangci-lint v2.0.2 says it needs 1.23 when I run it locally.
Any ideas? but it seems to work now - so 🤷

schuellerf:
hmm - there is a different go version in the golangci-lint container? I'll followup on this how to solve this version mixture…

mvo5 (changes requested):
Lets merge the first two commits, but the third needs more work/discussion ( I will followup with details).

.github/workflows/tests.yml
```diff
@@ -44,10 +41,14 @@ jobs:
       - name: Install libgpgme devel package
         run: sudo apt install -y libgpgme-dev libbtrfs-dev libdevmapper-dev
 
+      - name: Extract golangci-lint version from Makefile
+        id: golangci_lint_version
+        run: echo "GOLANGCI_LINT_VERSION=$(awk -F '=' '/^GOLANGCI_LINT_VERSION *=/{print $2}' Makefile)" >> "$GITHUB_OUTPUT"
```
achilleas-k, line 46: Perhaps it would be cleaner, if we want to keep these things aligned, to put them in a nicer machine readable (or just plain text) third place that both places can read.  Perhaps a `versions/` dir with some plain files like `versions/golang-lint-version` with just the number in there that the action and makefile can `cat`.

Also the message for this commit is too long.  Please keep it below 79 chars.

schuellerf:
🤔 the "third" is essentially fixing the linter problems - so let's follow up on the discussion - I'm not sure if we should merge without "[the third](https://github.com/osbuild/bootc-image-builder/pull/891/commits/b6ec6f173fe711efb8b27d5a3646b28b151bf256)"

mvo5:
> 🤔 the "third" is essentially fixing the linter problems - so let's follow up on the discussion - I'm not sure if we should merge without "[the third](https://github.com/osbuild/bootc-image-builder/pull/891/commits/b6ec6f173fe711efb8b27d5a3646b28b151bf256)"

Well, its broken already so nothing would be made worse. But sure, we can also fix it:
- the non-trival parts need separate commits
- I don't want a helper like this, for read-only Close() or in tests, I think its better to have the idiomatic defer f.Close() (the standard library uses this a lot too and if its good enough for them it should be good enough for us). I don't see how read-only close fail (the man-page on close is silent here). So just silencing the warnings here is IMO the way to go.
- For the Write close cases those need to actually become a real error as they can happen on e.g. NFS (aiui), so this needs to become a `if err := f.Close(); err != nil {return err}` (maybe with context)

bib/internal/utils/fileutils.go
```diff
@@ -0,0 +1,14 @@
+package utils
```
achilleas-k, line 1: Not a fan of a `utils` package.  It's also generally frowned upon.
Also the commit message says "fix linter problems" but there's a lot more going on here than just linter fixes (a new package is introduced, for one).

Also, I think you meant that problems _arose_, not _arouse_ in the commit message :)

bib/internal/utils/fileutils.go
```diff
@@ -0,0 +1,14 @@
...
+
+	"github.com/sirupsen/logrus"
+)
+
+// LogClose closes the file and logs any error encountered during the operation.
+func LogClose(file *os.File) {
+	if err := file.Close(); err != nil {
+		logrus.WithError(err).Errorf("failed to close file")
+	}
+}
```
achilleas-k, line 14: Do we need this?  An inline deferred function wouldn't be too bad.  From this diff it seems we only need it in a handful of places.  And having it inline would also let us add more context to each error (like `failed to close config file`).

achilleas-k (changes requested):

bib/cmd/bootc-image-builder/cloud.go
```diff
@@ -29,7 +30,7 @@ func upload(uploader cloud.Uploader, path string, flags *pflag.FlagSet) error {
 	if err != nil {
 		return fmt.Errorf("cannot upload: %v", err)
 	}
-	defer file.Close()
+	defer utils.LogClose(file)
```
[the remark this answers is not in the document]
supakeen: I don't think the question is if it is common, but the question is what the effect is. The statement "only one situation when an error can be returned" isn't entirely true, the [concrete implementation of `file`](https://cs.opensource.google/go/go/+/refs/tags/go1.24.2:src/os/file_unix.go;l=325-343) returns the underlying errors from `close(2)` (though it goes through an event loop).

This means that the usual errors from calling `close(2)` apply; in some cases (mostly writing to networked filesystems) it is possible to silently truncate writes.

Now the question is: is that case important to us, or important enough that we should make a decision per-`Close()` instead of globally.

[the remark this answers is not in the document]
mvo5: I agree with lzap here, this is not worth it for read-only FDs, so we should drop it there. And for RW fds we need to return a real error (as iirc delayed NFS writes can trigger errors on close). but in both cases we don't want this.

.github/workflows/tests.yml
schuellerf: which one?
```
git log -4 --pretty=format:"%s

%b"|while read line; do echo "${#line}: ${line}"; done
49: Makefile: include running go tests in `make test`
0: 
0: 
24: bib: fix linter problems
0: 
65: New linter problems arose after updating golangci-lint's version.
0: 
75: Makefile,github: implement `make lint` and centralize GOLANGCI_LINT_VERSION
0: 
0: 
34: github: bump golangci-lint version
```

achilleas-k: Sorry, it's actually 72.  Miscounted :)

lzap:
Ugly and interesting :-)

PR #892, opened 2025-04-10
Update golangci lint "ignore version"

schuellerf:
This is the alternative implementation to #891 (and #890 )
Which ignores the defer file.Close() remarks of the linter.
Only for the file where we write to we return an error message, similar message as we do _while_ writing.

### DIFF
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"log"
@@ -166,7 +167,7 @@ func makeManifest(c *ManifestConfig, solver *dnfjson.Solver, cacheRoot string) (
 	return mf, depsolvedRepos, nil
 }
 
-func saveManifest(ms manifest.OSBuildManifest, fpath string) error {
+func saveManifest(ms manifest.OSBuildManifest, fpath string) (err error) {
 	b, err := json.MarshalIndent(ms, "", "  ")
 	if err != nil {
 		return fmt.Errorf("failed to marshal data for %q: %s", fpath, err.Error())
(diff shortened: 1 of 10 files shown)


bib/cmd/bootc-image-builder/main.go
```diff
@@ -176,11 +176,15 @@ func saveManifest(ms manifest.OSBuildManifest, fpath string) error {
 	if err != nil {
 		return fmt.Errorf("failed to create output file %q: %s", fpath, err.Error())
 	}
-	// nolint:errcheck
-	defer fp.Close()
+	defer func() {
```
mvo5, line 179: if we use this I think `errors.Join(fp.Close())` is fine (the path is already part of path error, see ,https://github.com/golang/go/blob/master/src/os/file_unix.go#L316) I slightly prefer to just have a final `return fp.Close()` here but no strong opinion

bib/cmd/bootc-image-builder/main.go
```diff
@@ -176,11 +176,15 @@ func saveManifest(ms manifest.OSBuildManifest, fpath string) error {
...
+		if err = fp.Close(); err != nil {
+			err = fmt.Errorf("failed to close output file %q: %s", fpath, err.Error())
```
mvo5, line 181: We need to keep the error with `%w` here, but lets just not do it and use either `errors.Join()` or add  `return fp.Close()` at the end.

mvo5 (changes requested):
Thanks! I like this version much better. The write case needs some tweaks still.

bib/cmd/bootc-image-builder/main.go
schuellerf: I think in case of other errors the defer is a good Idea?
I'll adapt and propose your short version, which is a good "pattern" I guess.

mvo5 (dismissed):
Thanks, lets merge it.

mvo5 (approved):

PR #905, opened 2025-04-25
fix: allow distro definitition override

### LINKED ISSUE #904: Allow users to override OS matching to a specific OS version for manifest purposes
If you rename the OS when building custom images it will cause bootc image builder to fail to build. Instead of adding dozens of manifests it would be better to add a flag to allow the user to specify that this if a Fedora 42-based image or CentOS or Alma, etc.

hanthor:
Fixes #904

This accepts --override-distro-def=fedora-40.yaml to set the manifest. I'm not sure if this is ideal or if it would better to have it do into the the distrodef function like normal and allow for fuzzy match


hanthor:
```
	// use lorax-templates-rhel if the source distro is not Fedora with the exception of Fedora ELN
	img.UseRHELLoraxTemplates =
		c.SourceInfo.OSRelease.ID != "fedora" || c.SourceInfo.OSRelease.VersionID == "eln"

```

This is problematic for me https://github.com/hanthor/bootc-image-builder/blob/725b710052510c35e15ff09ff571b5a8e3be83af/bib/cmd/bootc-image-builder/image.go#L505-L508

supakeen:
> ```
> 	// use lorax-templates-rhel if the source distro is not Fedora with the exception of Fedora ELN
> 	img.UseRHELLoraxTemplates =
> 		c.SourceInfo.OSRelease.ID != "fedora" || c.SourceInfo.OSRelease.VersionID == "eln"
> ```
> 
> This is problematic for me https://github.com/hanthor/bootc-image-builder/blob/725b710052510c35e15ff09ff571b5a8e3be83af/bib/cmd/bootc-image-builder/image.go#L505-L508

This logic can likely be inverted to always use `-generic` *unless* it's RHEL, Alma Linux, or ELN. Thoughts? cc @mvo5 / @ondrejbudai?

Venefilyn:
So with distros that are based on a specific system there is always `ID_LIKE`. Could that be the primary thing to look for with `ID` as a backup, or are there edge-cases here that I do not know of? 

Example:
```
NAME="Bazzite"
VERSION="42.20250501.0 (Silverblue)"
RELEASE_TYPE=stable
ID=bazzite
ID_LIKE="fedora"
VERSION_ID=42
VERSION_CODENAME="Holographic"
PLATFORM_ID="platform:f42"
PRETTY_NAME="Bazzite 42 (FROM Fedora Silverblue)"
```

hanthor:
ID_LIKE can have multiple entries in it I think. Like Centos has rhel and fedora

hanthor:
We either need to ignore edits to os-relase and find a more hardcoded way to detect the distro or allow for overrides/have helpful text for the user when the manifest isn't found guiding them to just override to fedora or centos etc.

PR #907, opened 2025-04-25
image: set `UseRHELLoraxTemplates` only on rhel/centos/eln

mvo5:
This commit flips the detection if we need the rhel lorax template. It will only use it when it detect running on rhel/centos/eln. This should help with the common case of distros that need the generic lorax template but do not set the name to fedora.

This might help with the immediate issue in https://github.com/osbuild/bootc-image-builder/issues/904 but is probably useful on its own (as the generic lorax template is the (more) common case).

### DIFF
--- a/bib/cmd/bootc-image-builder/image.go
+++ b/bib/cmd/bootc-image-builder/image.go
@@ -7,6 +7,7 @@ import (
 	"math"
 	"math/big"
 	"math/rand"
+	"slices"
 	"strconv"
 	"strings"
 
@@ -435,6 +436,10 @@ func labelForISO(os *source.OSRelease, arch *arch.Arch) string {
 	}
 }
 
+func needsRHELLoraxTemplates(si source.OSRelease) bool {
+	return si.ID == "rhel" || slices.Contains(si.IDLike, "rhel") || si.VersionID == "eln"
+}
+
 func manifestForISO(c *ManifestConfig, rng *rand.Rand) (*manifest.Manifest, error) {
 	if c.Imgref == "" {
 		return nil, fmt.Errorf("pipeline: no base image defined")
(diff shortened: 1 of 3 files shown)


bib/cmd/bootc-image-builder/image.go
```diff
@@ -498,9 +498,8 @@ func manifestForISO(c *ManifestConfig, rng *rand.Rand) (*manifest.Manifest, erro
 	img.Kickstart.OSTree = &kickstart.OSTree{
 		OSName: "default",
 	}
-	// use lorax-templates-rhel if the source distro is not Fedora with the exception of Fedora ELN
 	img.UseRHELLoraxTemplates =
-		c.SourceInfo.OSRelease.ID != "fedora" || c.SourceInfo.OSRelease.VersionID == "eln"
+		c.SourceInfo.OSRelease.ID == "rhel" || c.SourceInfo.OSRelease.ID == "centos" || c.SourceInfo.OSRelease.VersionID == "eln"
```
supakeen, line 502: Also needs `almalinux`.

Conan-Kudo: Yes, `ID_LIKE` having `rhel` is a better check here.

mvo5: Thanks @Conan-Kudo and @supakeen - I tweaks this to use ID_LIKE now.

Conan-Kudo (approved):

supakeen (approved):

bcl (approved):
Looks like a good idea. Tests are unhappy but seems unrelated to this?

mvo5:
It looks like this needs https://github.com/osbuild/bootc-image-builder/pull/909 first :-/

PR #910, opened 2025-04-29
workflow: run pytests via matrix to get more parallel runs

mvo5:
This commit moves the test running into a matrix so that
we get more parallel testing. It will still be (much) dominated
by `test_build_disk.py` but at least this way a flaky test in
e.g. `test_container` is much faster to re-run.

With multiple VMs we can probably also parallelize the tests [1]
because we have less images per VM to test so diskspace
may be less of an issue.

---

test: split test_build.py into test_build_{disk,iso}

Split this big test into smaller files because we
will run the tests in parallel via a dynamic (per-file)
github matrix. This will allow faster tests and easier
re-runs if a single test is flaky only a small subset
will have to be retriggered.

[1] https://github.com/osbuild/bootc-image-builder/pull/787

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -70,10 +70,30 @@ jobs:
         # allow seemingly unreachable commands
         SHELLCHECK_OPTS: -e SC1091 -e SC2002 -e SC2317
 
+  collect_tests:
+    runs-on: ubuntu-latest
+    outputs:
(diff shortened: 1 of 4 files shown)


achilleas-k (dismissed):
Very nice!

mvo5:
(sorry, had to force push to rebase and fix a conflict)

achilleas-k (approved):

PR #911, opened 2025-04-30
test: split cross-arch test into its own file

mvo5:
This commit moves the cross arch build into its own file
so that it ran run in parallel in the GH runners. Its is
a relatively expensive test (~20min on my machine) so
moving it out should save quite a bit of time.

### DIFF
--- a/test/test_build_cross.py
+++ b/test/test_build_cross.py
@@ -1,0 +1,23 @@
+import platform
+
+import pytest
+
+from testcases import gen_testcases
+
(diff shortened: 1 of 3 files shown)


achilleas-k (approved):

achilleas-k:
Linter's unhappy, but it seems we don't require it.

mvo5:
> Linter's unhappy, but it seems we don't require it.

Yeah, linter fix is in https://github.com/osbuild/bootc-image-builder/pull/914

PR #916, opened 2025-05-07
main: drop rootfs cross-arch disable

supakeen:
All platforms where `bootc-image-builder` runs that have `qemu-user` available have upgrade to a version >= 9.1.0.


mvo5 (approved):
Very nice, thank you!

supakeen:
Since we now have the newer qemu in @mvo5's repository I found a failing test that assumed cross-arch to always be ext4. I've adjusted it and force-pushed.

mvo5 (approved):

PR #917, opened 2025-05-07
container: fedora 42

supakeen:
Let's base our upstream container on Fedora 42 now that it is GA.


mvo5 (approved):
Thank you

supakeen:
@mvo5 I added a commit that also updates the containers used in testing farm, and in the `devel/` subdirectory just so that we do all of them in this PR.

mvo5 (approved):

PR #920, opened 2025-05-08
Extract SELinux policy from container

alexlarsson:
Currently we are always hardcoding "targeted", which is not working for the centos automotive sig that use a custom policy.

To completely work, this also needs https://github.com/osbuild/images/pull/1506, otherwise the extracted selinux policy is not used in the build pipeline.


alexlarsson:
Closed in favor of https://github.com/osbuild/bootc-image-builder/pull/928

PR #928, opened 2025-05-13
Automotive initial work

alexlarsson:
This is some initial work for supporting bootc in the automotive projects. It includes the following:

 * Extract the active selinux policy from the image
 * Support specifying a custom build container image ref

This depends on: https://github.com/osbuild/images/pull/1519

PR #932, opened 2025-05-20
Apply disk or filesystem customization from embedded image customization

alexlarsson:
This allows bootc images to embed a  file in `/usr/lib/bootc-image-builder` in the image called `config.json` or `config.toml`, which specify more detailed requirements for the  partitioning. From this (which is in the full blueprint format) we extract (only) the disk and/or filesystem customization part.

This is useful to either add extra partitions (like a separate /var), or to override details of the normal partitions (like uuids, labels, etc).

This is discussed in https://github.com/bootc-dev/bootc/issues/926

I used the entire blueprint format so that we can reuse existing docs, and so that we can later use more customization fields if needed.

Note: This (mostly trivially) conflicts with https://github.com/osbuild/bootc-image-builder/pull/928

### DIFF
--- a/bib/cmd/bootc-image-builder/image.go
+++ b/bib/cmd/bootc-image-builder/image.go
@@ -209,6 +209,18 @@ func genPartitionTable(c *ManifestConfig, customizations *blueprint.Customizatio
 	if err != nil {
 		return nil, fmt.Errorf("error reading disk customizations: %w", err)
 	}
+
+	// Embedded disk customization applies if there was no local customization
+	if fsCust == nil && diskCust == nil && c.SourceInfo != nil && c.SourceInfo.ImageCustomization != nil {
+		imageCustomizations := c.SourceInfo.ImageCustomization
+
+		fsCust = imageCustomizations.GetFilesystems()
+		diskCust, err = imageCustomizations.GetPartitioning()
+		if err != nil {
+			return nil, fmt.Errorf("error reading disk customizations: %w", err)
+		}
+	}
+
 	switch {
 	// XXX: move into images library
 	case fsCust != nil && diskCust != nil:
--- a/bib/internal/buildconfig/config.go
+++ b/bib/internal/buildconfig/config.go
@@ -101,6 +101,16 @@ func loadConfig(path string) (*externalBlueprint.Blueprint, error) {
 	}
 }
 
+func LoadConfig(path string) (*imagesBlueprint.Blueprint, error) {
+	externalBp, err := loadConfig(path)
+	if err != nil {
+		return nil, err
+	}
+
+	bp := externalBlueprint.Convert(*externalBp)
+	return &bp, nil
+}
+
 func readWithFallback(userConfig string) (*externalBlueprint.Blueprint, error) {
 	// user asked for an explicit config
 	if userConfig != "" {
--- a/bib/internal/source/source.go
+++ b/bib/internal/source/source.go
@@ -8,8 +8,12 @@ import (
 
 	"github.com/sirupsen/logrus"
 
+	"github.com/osbuild/bootc-image-builder/bib/internal/buildconfig"
+	"github.com/osbuild/images/pkg/blueprint"
 	"github.com/osbuild/images/pkg/distro"
 )
+
+const bibPathPrefix = "usr/lib/bootc-image-builder"
 
 type OSRelease struct {
 	PlatformID string
@@ -21,8 +25,9 @@ type OSRelease struct {
 }
 
 type Info struct {
-	OSRelease  OSRelease
-	UEFIVendor string
+	OSRelease          OSRelease
+	UEFIVendor         string
+	ImageCustomization *blueprint.Customizations
 }
 
 func validateOSRelease(osrelease map[string]string) error {
@@ -58,6 +63,26 @@ func uefiVendor(root string) (string, error) {
 	return "", fmt.Errorf("cannot find UEFI vendor in %s", bootupdEfiDir)
 }
 
+func readImageCustomization(root string) (*blueprint.Customizations, error) {
+	prefix := path.Join(root, bibPathPrefix)
+	config, err := buildconfig.LoadConfig(path.Join(prefix, "config.json"))
+	if err != nil && !os.IsNotExist(err) {
+		return nil, err
+	}
+	if config == nil {
+		config, err = buildconfig.LoadConfig(path.Join(prefix, "config.toml"))
+		if err != nil && !os.IsNotExist(err) {
+			return nil, err
+		}
+	}
+	// no config found in either toml/json
+	if config == nil {
+		return nil, nil
+	}
+
+	return config.Customizations, nil
+}
+
 func LoadInfo(root string) (*Info, error) {
 	osrelease, err := distro.ReadOSReleaseFromTree(root)
 	if err != nil {
@@ -71,6 +96,12 @@ func LoadInfo(root string) (*Info, error) {
 	if err != nil {
 		logrus.Debugf("cannot read UEFI vendor: %v, setting it to none", err)
 	}
+
+	customization, err := readImageCustomization(root)
+	if err != nil {
+		return nil, err
+	}
+
 	var idLike []string
 	if osrelease["ID_LIKE"] != "" {
 		idLike = strings.Split(osrelease["ID_LIKE"], " ")
@@ -86,6 +117,7 @@ func LoadInfo(root string) (*Info, error) {
 			IDLike:     idLike,
 		},
 
-		UEFIVendor: vendor,
+		UEFIVendor:         vendor,
+		ImageCustomization: customization,
 	}, nil
 }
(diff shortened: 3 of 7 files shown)


achilleas-k (dismissed):
Thank you, this is great!!

I think this change is important enough to warrant a new test container in CI.
https://github.com/osbuild/bootc-image-builder/blob/main/test/test_manifest.py
Perhaps not a full build, but at least a manifest generation and inspection test.

bib/internal/buildconfig/config.go
```diff
@@ -101,6 +101,15 @@ func loadConfig(path string) (*externalBlueprint.Blueprint, error) {
 	}
 }
 
+func LoadConfig(path string) (*imagesBlueprint.Blueprint, error) {
+	externalBp, err := loadConfig(path)
```
mvo5, line 105: (nitpick) I would prefer if we would follow the idiomatic go style here, i.e.
```go
    externalBp, err := loadConfig(path)
    if err != nil {
         return nil, err
    }
    bp := externalBlueprint.Convert(*externalBp)
    return &bp, nil
}
```
I know its the same but when reading (go) code, I just stumble on anything that does not follow the usual "foo, err := ...; if err != nil {" pattern.

bib/cmd/bootc-image-builder/image.go
```diff
@@ -209,6 +209,18 @@ func genPartitionTable(c *ManifestConfig, customizations *blueprint.Customizatio
 	if err != nil {
 		return nil, fmt.Errorf("error reading disk customizations: %w", err)
 	}
+
+	// Embedded disk customization applies if there was no local customization
+	if fsCust == nil && diskCust == nil && c.SourceInfo != nil && c.SourceInfo.ImageCustomization != nil {
+		imageCustomizations := c.SourceInfo.ImageCustomization
+
+		fsCust = imageCustomizations.GetFilesystems()
```
mvo5, line 217: Having a test for this would be really good, I hesitate to ask you to write as it might be a bit annoying/involved - I am happy to look into it in the morning and provide a diff.

bib/internal/source/source.go
```diff
@@ -58,6 +63,27 @@ func uefiVendor(root string) (string, error) {
 	return "", fmt.Errorf("cannot find UEFI vendor in %s", bootupdEfiDir)
 }
 
+func readImageCustomization(root string) (*blueprint.Customizations, error) {
```
mvo5, line 66: Having a (unit) test for this would be really nice, but I don't want to put this burden on you, happy to look at it (in the morning).

bib/internal/source/source.go
```diff
@@ -58,6 +63,27 @@ func uefiVendor(root string) (string, error) {
...
+	prefix := path.Join(root, bibPathPrefix)
+	config, err := buildconfig.LoadConfig(path.Join(prefix, "config.json"))
+	if err != nil && os.IsNotExist(err) {
```
mvo5, line 69: I wonder if we can slightly tweak this to make it more linear, i.e. to error earlier. Right now one needs to keep the state of err and config in (brain) memory when reading this to reason when what error is returned. Tweaking it slightly like:
```go
func readImageCustomization(root string) (*blueprint.Customizations, error) {
	prefix := path.Join(root, bibPathPrefix)
	config, err := buildconfig.LoadConfig(path.Join(prefix, "config.json"))
	if err != nil && !os.IsNotExist(err) {
		return nil, err
	}
	if config == nil {
		// no json config found, try toml
		config, err = buildconfig.LoadConfig(path.Join(prefix, "config.toml"))
	}
	if err != nil && !os.IsNotExist(err) {
		return nil, err
	}
	// no config found in either toml/json
	if config == nil {
		return nil, nil
	}

	return config.Customizations, nil
}
``` 
I /think/ the above version is equivalent to the original one (and I can write a test that validates it tomorrow).

mvo5 (commented):
Thank you! This is a really nice feature, we should probably add it to the README.md as well that this is now possible. 

I also wonder if we should error (or warn) if there are most customizations embedded than the filesystem/disk ones. Mostly to avoid surprises when people might think adding extra stuff would work when in fact it does not. Obviously followup material.

I also included tiny suggestions for the go-code but and would love some tests but all of those I can do in a followup, I don't want to slow this (excellent) work down.

alexlarsson:
I added some tests

bib/cmd/bootc-image-builder/image.go
alexlarsson: added unit tests

bib/internal/source/source.go
alexlarsson: added unit test

bib/internal/source/source.go
alexlarsson: fixed

alexlarsson:
Note: The tests fail to set `part_type` because we lack this blueprint commit: https://github.com/osbuild/blueprint/commit/bb3eeadabff76c1ee61dd11967faf4e5ff55c594
which adds handling of PartType in the Convert() function.

Once we update the blueprint version we should extend these tests to include this, and the new options.

alexlarsson:
> I also wonder if we should error (or warn) if there are most customizations embedded than the filesystem/disk ones. Mostly to avoid surprises when people might think adding extra stuff would work when in fact it does not. Obviously followup material.

Yeah, that is probably a good idea.

mvo5 (approved):
Thank you, this is very nice!

alexlarsson:
Some test is failing with:
```
org.osbuild.bootc.install-to-filesystem: fab7420133de53329157eca54e58fbe60f081859afc46b22c14d9231befaa71e {
  "kernel-args": [
    "rw",
    "console=tty0",
    "console=ttyS0",
    "systemd.journald.forward_to_console=1"
  ],
  "target-imgref": "quay.io/fedora/fedora-bootc:43"
}
device/disk (org.osbuild.loopback): loop0 acquired (locked: False)
mount/- (org.osbuild.btrfs): mounting /dev/loop0p4 -> /store/tmp/buildroot-tmp-7wl3rl6k/mounts/
mount/boot (org.osbuild.xfs): mounting /dev/loop0p3 -> /store/tmp/buildroot-tmp-7wl3rl6k/mounts/boot
mount/boot-efi (org.osbuild.fat): mounting /dev/loop0p2 -> /store/tmp/buildroot-tmp-7wl3rl6k/mounts/boot/efi
Mount transient overlayfs for /etc/containers
Creating bind mount for run/osbuild/containers
Installing image: docker://quay.io/fedora/fedora-bootc:43
Initializing ostree layout
ERROR Installing to filesystem: Creating ostree deployment: invalid reference format
```

Doesn't seem related to this PR, but I don't understand how the reference format is invalid in the first place?

mvo5:
> Some test is failing with:
[..]
> Doesn't seem related to this PR, but I don't understand how the reference format is invalid in the first place?

It is not related, I can reproduce this locally with:
```console
$ git rev-parse HEAD
66049b592e1e1b8ecbce4d0b06fd6dbb2fead663
$ sudo pytest -s -vv ./test/test_build_disk.py::test_image_boots[container_ref=quay.io/fedora/fedora-bootc:43,disk_config=btrfs,image=raw,rootfs=btrfs,use_librepo=True]
...
```
and get the same error. So I guess we need to investigate  the bootc-fedora:43 container, but we can skip the test for now to unblock you, this fedora:43 issue is outside of our controll.

alexlarsson:
The other failure seems to be:
```
Resolved "registry" as an alias (/etc/containers/registries.conf.d/000-shortnames.conf)
Trying to pull docker.io/library/registry:2...
time="2025-05-21T09:37:53Z" level=warning msg="Failed, retrying in 1s ... (1/3). Error: initializing source docker://registry:2: reading manifest 2 in docker.io/library/registry: toomanyrequests: You have reached your unauthenticated pull rate limit. https://www.docker.com/increase-rate-limit"
time="2025-05-21T09:38:25Z" level=warning msg="Failed, retrying in 1s ... (2/3). Error: initializing source docker://registry:2: reading manifest 2 in docker.io/library/registry: toomanyrequests: You have reached your unauthenticated pull rate limit. https://www.docker.com/increase-rate-limit"
time="2025-05-21T09:38:56Z" level=warning msg="Failed, retrying in 1s ... (3/3). Error: initializing source docker://registry:2: reading manifest 2 in docker.io/library/registry: toomanyrequests: You have reached your unauthenticated pull rate limit. https://www.docker.com/increase-rate-limit"
Error: initializing source docker://registry:2: reading manifest 2 in docker.io/library/registry: toomanyrequests: You have reached your unauthenticated pull rate limit. https://www.docker.com/increase-rate-limit
_ ERROR at setup of test_image_boots[container_ref=quay.io/fedora/fedora-bootc:42,image=qcow2,rootfs=btrfs,use_librepo=True] _

shared_tmpdir = PosixPath('/var/tmp/bib-tests/shared1')
request = <SubRequest 'registry_conf' for <Function test_image_is_generated[container_ref=quay.io/centos-bootc/centos-bootc:stream9,image=qcow2+raw+vmdk+vhd+gce,use_librepo=True,use_terminal=True]>>
```

mvo5 (approved):
Thank you!

achilleas-k (approved):
Thanks for all of this.  It's great to finally have this feature!
LGTM

PR #964, opened 2025-06-17
bib: fix `TestManifestSerialization` tests

mvo5:
With https://github.com/osbuild/images/pull/1571 merged we now potentially have an extra `target` build pipeline that needs container specs.

Sadly a misconfiguration in the branch protection for "main" merged images v0.151 with the now broken unit tests in https://github.com/osbuild/bootc-image-builder/pull/955 (the branch protection is now fixed).

This commit now also fixes the test by adding a mock container spec for the "target" pipeline.

### DIFF
--- a/bib/cmd/bootc-image-builder/main_test.go
+++ b/bib/cmd/bootc-image-builder/main_test.go
@@ -211,6 +211,9 @@ var diskContainers = map[string][]container.Spec{
 	"image": {
 		containerSpec,
 	},
+	"target": {
+		containerSpec,
+	},
 }
 
 // TODO: this tests at this layer is not ideal, it has too much knowledge


supakeen (approved):
Nice find :)

mvo5:
photo_5370609875412975485_y

It looks like the iso install for centos-10 is not happy right now either

supakeen:
> I included a workaround for the installer issue (disabling the test for now), once this is merged I will open a "revert" PR for the test-disable so that we do not forget the issue.

Mmm, I thought these had already made it into CentOS; if we're still seeing issues there's no problem with skipping the test for now.

All the Konflux and Testing Farm tests are still red, is that also expected?

Issue #645, opened 2024-09-11
New selinux denial


mvo5:
During the "bib" build we have a new selinux denial that needs investigation.

So far we know:
```
type=AVC msg=audit(1726075609.598:896): avc:  denied  { nnp_transition nosuid_transition } for  pid=18439 comm="bootc" scontext=system_u:system_r:install_t:s0:c264,c849 tcontext=system_u:system_r:container_runtime_t:s0:c264,c849 tclass=process2 permissive=0
```

cgwalters:
Also looks like potential fallout from https://github.com/containers/bootc/commit/0527ca96202633625f79dfe06277b96cfb522000 (xref https://github.com/osbuild/bootc-image-builder/issues/639#issuecomment-2338273655 )

Although, I'm not seeing this here. I guess the mainly relevant thing becomes the version of the *host system* doing the install - commonly I'm testing with podman-machine which host is f40.

But yes, I think this is one we need to get the selinux maintainers to allow. Can you get the version of the host system selinux policy here and then we can file a bug?

PR #973, opened 2025-07-04
test: fix various testing-farm issue

mvo5:
test: enable `OSBUILD_EXPERIMENTAL=debug-qemu-user`

This commit enables OSBUILD_EXPERIMENTAL=debug-qemu-user
during the tests so that we see what unimplemented syscalls
or ioctls are used.

This should help with the cross-arch failure debugging we
see in tmt right now.

---

plan: add `osbuild-{lvm2,ostree}` to tmt test

This commit adds the missing `osbuild-{lvm2,ostree}` to
integration.fmf so that the manifest lvm tests work.

---

test: skip the test_image_build_without_se_linux_denials

Skip the test_image_build_without_se_linux_denials test until
https://github.com/osbuild/bootc-image-builder/issues/645 is
resolved.

It pains me to do this but the test is failing for some time
and bib itself cannot do anything to resolve this (AIUI it
need an upstream selinux policy change so that install_t
can transition to container_runtime_t because of
https://github.com/bootc-dev/bootc/commit/0527ca96202633625f79dfe06277b96cfb522000

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -28,5 +28,5 @@ VOLUME /var/lib/containers/storage
 LABEL description="This tools allows to build and deploy disk-images from bootc container inputs."
 LABEL io.k8s.description="This tools allows to build and deploy disk-images from bootc container inputs."
 LABEL io.k8s.display-name="Bootc Image Builder"
-LABEL io.openshift.tags="base fedora40"
+LABEL io.openshift.tags="base fedora42"
 LABEL summary="A container to create disk-images from bootc container inputs"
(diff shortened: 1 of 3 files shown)


mvo5:
Note that the "Testing farm go unit tests " will only resolve itself *after* the merge to master because for testing farm we use pull_request_target trigger which gets the workflow file from "main" instead of the branch for safety reasons.

ondrejbudai (dismissed):
Thanks!

ondrejbudai (approved):

Issue #1000, opened 2025-07-29
konflux pipeline down


cgwalters:
> Does anyone know how I can get a new version of quay.io/centos-bootc/bootc-image-builder to be built? It seems there have been no builds in [11 days](https://quay.io/centos-bootc/bootc-image-builder) and I've fixed a few bugs upstream that are thus still affecting users. So far I've advised those users to just build a local container, but since this is a recurring theme (some users have set up their own pipelines to build that container ...) I'd thought I'd ask here.

Reportedly this may be https://issues.redhat.com/browse/KONFLUX-9346

cgwalters:
Related: We should also discuss somewhere the future of the centos-bootc namespace and the production builds.

A tension here is that bootc-image-builder is in theory at least partially OS independent, totally unlike the bootc base images. 

There's really three options:

- Build this container in each OS namespace i.e. introduce quay.io/fedora/bootc-image-builder and quay.io/centos/bootc-image-builder
- Have upstream builds in e.g. ghcr.io/osbuild (or quay.io/osbuild, whatever)
- Both of the above

supakeen:
A related bug that this is blocking is #968. There's also some workaround linked there.

cgwalters:
> Personally I'd much prefer if these things live upstream instead of downstream :)

The centos-bootc one feels more "midstream" to me than upstream in that sense, but yes.

cgwalters:
Yes. That said, Red Hat products will continue to be required to build with Konflux downstream. 
One obvious thing to start with what you have in PR #1001 is it's only building for x86_64. I did some web searching which turns up https://github.com/praetorian-inc/noseyparker/pull/214 which I think is a decent pattern that we could probably try to productize more and share across upstreams.

We may also end up wanting to maintain shared tools to keep Konflux pipelines and GHA in sync. (In general I've been trying to move load-bearing logic outside of shell-script-in-YAML in GHA in that vein)

cgwalters:
Thanks @shi2wei3 !

On the overall topic there's definitely no issues AFAIK with ensuring that the team here has operational control and visibility over the current CentOS Konflux pipelines. Here I think at least ⅔ of the problem is that the release policies being deleted was quite painful to trace down and not obvious (as opposed to straight up build failures).

But if you want I'm also personally fine with just building the containers via some other mechanism and we can ensure that what's in `quay.io/centos-bootc/bootc-image-builder` syncs from there instead of being built via Konflux.

[...]

PR #1001, opened 2025-07-29
github: publish upstream container to ghcr

achilleas-k:
At some point we stopped building and pushing the container to ghcr to focus on having only Konflux builds and a single source of the upstream (though, more like "midstream") container on quay.io.  However, Konflux builds are not entirely reliable and are inaccessible to external users and contributors.  So while the container itself is publicly available under the quay.io/centos-bootc namespace, investigating (and solving) build failures requires contacting people with access to the private job logs.

Re-enabling ghcr builds allows us (and practically anyone) to easily investigate and solve build failures and gives us a reliable upstream container available shortly after every PR is merged, making troubleshooting against the latest version a lot easier.

This reverts commit 62dda279dae875539436f30aafc5edd45b7fb88e.

See also discussion in #1000.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -3,14 +3,19 @@ name: Build containers
 on:
   pull_request:
     branches: [main]
+  workflow_dispatch:
   # for merge queue
   merge_group:
+  push:
+    branches: [main]
 
 env:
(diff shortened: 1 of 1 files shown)


mvo5 (approved):
\o/

PR #1010, opened 2025-08-04
bib: set the boot mode to UEFI only for aarch64

achilleas-k:
Setting it to hybrid creates the BIOS boot partition (1 MiB empty partition at the start of the partition table), which can cause boot issues with certain devices.

### DIFF
--- a/bib/cmd/bootc-image-builder/image.go
+++ b/bib/cmd/bootc-image-builder/image.go
@@ -310,10 +310,16 @@ func genPartitionTableDiskCust(c *ManifestConfig, diskCust *blueprint.DiskCustom
 	if err != nil {
 		return nil, err
 	}
+
+	bootMode := platform.BOOT_HYBRID
+	if c.Architecture == arch.ARCH_AARCH64 {
+		// This prevents the creation of BIOS boot partitions for ARM devices
+		bootMode = platform.BOOT_UEFI
+	}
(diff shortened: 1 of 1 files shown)


bib/cmd/bootc-image-builder/image.go
```diff
@@ -310,10 +310,16 @@ func genPartitionTableDiskCust(c *ManifestConfig, diskCust *blueprint.DiskCustom
 	if err != nil {
 		return nil, err
 	}
+
+	bootMode := platform.BOOT_HYBRID
+	if c.Architecture == arch.ARCH_AARCH64 {
```
mvo5, line 315: We probably want to apply the same for ARCH_RISCV64

mvo5 (commented):
I like this! I think we should start with that (and probably also Simons PR) to unblock people. I would love to explore if we can do a bit more here in the images library though, i.e. https://github.com/osbuild/images/compare/main...mvo5:bootc-uefi-aarch64-2?expand=1 or https://github.com/mvo5/images/pull/new/bootc-uefi-aarch64-3 and making the partition building code a bit more aware of the architecture constraints.  

Thinking about it a bit more I wonder if we should actually introduce a new BOOT_NATIVE or BOOT_DEFAULT that would be set on anything but X86_64 as it seems that this is the only platform that has the weird legacy/uefi/hybrid requirement.

Issue #1029, opened 2025-08-25
Bib build ami  disk type failure


chunfuwen:
1. sudo podman run --rm -it --privileged --pull=newer --security-opt label=type:unconfined_t -v /var/lib/libvirt/images/output:/output -v /var/lib/libvirt/images/config.json:/config.json   -v /var/lib/containers/storage:/var/lib/containers/storage   -v /var/lib/libvirt/images/auth.json:/run/containers/0/auth.json   --env-file=/var/lib/libvirt/images/aws.secrets  quay.io/centos-bootc/bootc-image-builder:latest  --type ami --tls-verify=true  --config /config.json  --aws-ami-name build_upstream_bib-centos-component-bootc-ami_x86_64_291 --aws-bucket bib-component-test --aws-region us-east-1  quay.io/centos-bootc/centos-bootc:stream10  --rootfs xfs
2. Output below error message:
  2025/08/25 09:10:16 error: cannot handle AWS setup: unsupported architecture ""
3. The running host is :x86_64

PR #1017, opened 2025-08-20
bib: update to images v0.177

mvo5:
This PR (or the need for it) indicates once more that we really
need to consolidate the manifest generation into the images
library (c.f. https://github.com/osbuild/images/pull/1736) as
most of the adjustments would have one with the various
API changing changes in images if it was part of "images".

---

go.mod: update images to v0.177

---

image: update blueprint import path

We no longer ship the blueprint package inside the images library
but its now part of its own "blueprint" library.

---

main: update for v0.177 API changes

This includes:
- disk.ParitioningMode->parition.PartitioningMode
- anaconda.AdditionalModules -> anaconda.EnabledModules
- use of new InstallerCustomizations
- aws uploader needs arch.Arch

---

bib: enable anaconda Module{Network,Payload,Runtime}

Previously images [0] would take `AdditionalAnacondaModules`,
i.e. we had a set of defaults in code with
- anaconda.ModuleNetwork
- anaconda.ModulePayloads
- anaconda.ModuleRuntime
enabled by default. With the PR landed there are no longer
default so we need to update our code to include those
in the list of EnabledAnacondaModules.

[0] Before https://github.com/osbuild/images/pull/1737/

PR #1030, opened 2025-08-25
bib: fix AWS upload with empty target-arch

mvo5:
This commit fixes the issue that with an unset --target-arch the code will try to convert an empty string to an arch.Arch which then fails. This is a regression from
https://github.com/osbuild/bootc-image-builder/pull/1017 that was not caught.

Closes: https://github.com/osbuild/bootc-image-builder/issues/1029


supakeen (approved):
Do we want a test?

PR #1059, opened 2025-09-25
[WIP] many: switch to bootc based anaconda (away from rpm) (HMS-9392)

mvo5:
[build on top of https://github.com/osbuild/bootc-image-builder/pull/1058, much improved version of https://github.com/osbuild/bootc-image-builder/pull/1053 with support for "legacy" rpm based image types]

This spike show how we could use bootc based containers with anaconda. So instead of building our iso from rpm packages and just putting the bootc container into the rpm generated ISO this spike builds the ISO from a bootc container.

This requires that the container has the right content of course (i.e. anaconda). We would move some of the responsibility to the user (or to a tool that helps the user and hides the complexity here like bootc-iso-imagectl or something) to ensure that. 

The `Containerfile` to generate an installer looks something like:
```Containerfile
FROM {container_ref}
RUN dnf install -y \
     anaconda \
     anaconda-install-env-deps \
     anaconda-dracut \
     dracut-config-generic \
     dracut-network \
     net-tools \
     squashfs-tools \
     grub2-efi-x64-cdboot \
     python3-mako \
     lorax-templates-* \
     biosdevname \
     prefixdevname \
     && dnf clean all
# shim-x64 is marked installed but the files are not in the expected
# place for https://github.com/osbuild/osbuild/blob/v160/stages/org.osbuild.grub2.iso#L91, see
# workaround via reinstall, we could add a config to the grub2.iso
# stage to allow a different prefix that then would be used by
# anaconda.
# once https://github.com/osbuild/osbuild/pull/2202 is merged we
# can update images/ to set the correct efi_src_dir and this can
# be removed
RUN dnf reinstall -y shim-x64
# lorax wants to create a symlink in /mnt which points to /var/mnt
# on bootc but /var/mnt does not exist on some images.
#
# If https://gitlab.com/fedora/bootc/base-images/-/merge_requests/294
# gets merged this will be no longer needed
RUN mkdir /var/mnt
```
this basic template works for cs9,cs10,f42 (Note that the goal here is to essentially remove all quirks
and hide them in images/, ideally the containerfile would be just a "FROM <base>\ndnf install anaconda...").

This branch also has a "--installer-payload" switch so to use this one would essentially write: `bib --type iso --installer-payload quay.io/centos-bootc/centos-bootc:stream9 quay.io/centos-bootc-anaconda:stream9` (as currently the containerfile changes things like /root which makes it no longer a proper bootc container). We could of course explore making this more user friendly, either by a convention like "suffix -installer" for type iso or by creating a derrived constainer locally ourselfs with the above containerfile or helper or by fixing the issues that require /root to be a real dir (but even with that it would probably be undesired to install a container that contains a full anaconda env so something like installer-payload is probably still be needed).

Note that this is big departure from our current approach and we would need to think about a transition strategy/communication if we go down this route.

Thanks to Ondrej, Colin for their help/inputs in this, I hope it captures what they suggested to explore.

/jira-epic HMS-8839

JIRA: [HMS-9392](https://issues.redhat.com/browse/HMS-9392)


supakeen:
The links in the beginning are broken/wrong :)

test/test_build_iso.py
```diff
@@ -83,3 +88,119 @@ def test_iso_install_img_is_squashfs(tmp_path, image_type):
...
         assert "usr/bin/bootc" in output
+
+
+@pytest.mark.skipif(platform.system() != "Linux", reason="boot test only runs on linux right now")
+@pytest.mark.parametrize("container_ref", [
+    "quay.io/centos-bootc/centos-bootc:stream10",
+    "quay.io/fedora/fedora-bootc:42",
+    "quay.io/centos-bootc/centos-bootc:stream9",
+])
+def test_container_iso_installs(tmp_path, build_container, container_ref):
```
mvo5, line 99: Running this locally:
```
======================== 3 passed in 1340.54s (0:22:20) ========================
```
:)

mvo5:
With https://github.com/osbuild/images/pull/1906 this is no longer needed

PR #1066, opened 2025-10-06
many: drop custom `distrodefs` and use images DistroYAML

mvo5:
[build on top of #1065 - only the last commit is relevant for this]

Draft as this requires some work in "images" too:
1. ensure all distro aliases are supported in images
2. ensure there is an image-type for bootc-rpm-based-iso-installer with the correct config/pkgsets (ideally we would reuse something existing)

Instead of having a separate implementation of the required
packages and the required lorax templates and similar we
now reuse the images `data/distrodefs` YAML so that all
the packages/definitions are maintained in a central place.


mvo5:
Sorry, had to force push because of conflicts.

thozza (approved):

PR #1094, opened 2025-10-29
bib: add support for `--type bootc-installer`

mvo5:
This commit adds support for the new `bootc-installer` image
type that will take a bootc container and create an ISO out
of it. It also adds a new `--installer-payload-ref` option
so that the user can specify a different payload container
to install.

See https://github.com/osbuild/images/pull/1906 for details.

This is the equivalent of
https://github.com/osbuild/image-builder-cli/pull/341
for bootc-image-builder and allows us to build these kinds
of images with bib now too.

/jira-epic [HMS-8839](https://issues.redhat.com/browse/HMS-8839)

### DIFF
--- a/test/test_build_iso.py
+++ b/test/test_build_iso.py
@@ -1,12 +1,17 @@
 import os
+import random
+import json
 import platform
+import string
 import subprocess
+import textwrap
 from contextlib import ExitStack
 
 import pytest
 # local test utils
 import testutil
 from containerbuild import build_container_fixture    # pylint: disable=unused-import
+from containerbuild import make_container
 from testcases import gen_testcases
 from vm import QEMU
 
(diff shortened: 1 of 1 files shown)


supakeen (approved):

README.md
```diff
@@ -172,7 +172,8 @@ The following image types are currently available via the `--type` argument:
 | `ami`                 | [Amazon Machine Image](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/AMIs.html) |
 | `qcow2` **(default)** | [QEMU](https://www.qemu.org/)                                                         |
 | `vmdk`                | [VMDK](https://en.wikipedia.org/wiki/VMDK) usable in vSphere, among others            |
-| `anaconda-iso`        | An unattended Anaconda installer that installs to the first disk found.               |
+| `bootc-installer`     | An installer ISO image based on the specified bootc container image.                  |
+| `anaconda-iso`        | (legacy) an unattended Anaconda installer that installs to the first disk found build from RPMs. |
```
supakeen: I don't think this is legacy yet? Does any project currently publish installer containers?

cgwalters: > @cgwalters since you're here now; are there already issues tracking the building of a 'base installer' container for Fedora and/or CentOS?

Not that I know of, we could definitely add something like this in fedora-bootc.

But it's a bit complex because that'd add *another* variable in all of this. The way I was thinking of this is we just document how to add Anaconda as a layer on *top* of one's image - this means it naturally carries *all* configuration from the base (including critical stuff like kernel + kargs + injected 3rd party drivers in /usr/lib/modules etc.)

If we shipped a pre-built installer container...hmm, well I guess it would work equally well really (maybe sometimes better even?) to have users do a matrix build where they do two builds, one `FROM <golden installer image>` and one that generates their target.

mvo5: Sure thing, I can look into this tomorrow.

supakeen (approved):

test/test_build_iso.py
```diff
@@ -83,3 +88,120 @@ def test_iso_install_img_is_squashfs(tmp_path, image_type):
...
+    # shim-x64 is marked installed but the files are not in the expected
+    # place for https://github.com/osbuild/osbuild/blob/v160/stages/org.osbuild.grub2.iso#L91, see
+    # workaround via reinstall, we could add a config to the grub2.iso
+    # stage to allow a different prefix that then would be used by
+    # anaconda.
+    # once https://github.com/osbuild/osbuild/pull/2202 is merged we
+    # can update images/ to set the correct efi_src_dir and this can
+    # be removed
+    # see also https://bugzilla.redhat.com/show_bug.cgi?id=1750708
+    RUN dnf reinstall -y shim-x64
```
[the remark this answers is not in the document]
mvo5: Let me update the comment, sadly its more complicated, see https://github.com/osbuild/osbuild/pull/2205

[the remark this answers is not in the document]
mvo5: Updated

PR #1102, opened 2025-11-05
test: refactor so that all vm related tests can be reused [HMS-9572]

mvo5:
This commit reshuffles the code a bit so that we a reusable ./vmtest directory that can be used by the images library.

With that we can add a toplevel pyproject.toml file so that we can import vmtest via
```console
$ pip install git+https@github.com/osbuild/bootc-image-builder
```
im other projects.

Note that none of this is ideal, I wish python would be more like go here and make this easier, I dislike that we have to have the pyproject in a toplevel dir. But this is (hopefully) a temporary measure until we find a more permanent home for our vm runner or replace it with something like test.thing or the new osbuild QEMU code.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -157,7 +157,7 @@ jobs:
         # podman needs (parts of) the environment but will break when
         # XDG_RUNTIME_DIR is set.
         # TODO: figure out what exactly podman needs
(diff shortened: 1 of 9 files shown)


supakeen:
There are a bunch of pylint errors here, can those be addressed?

PR #1157, opened 2025-11-19
bib: remove bootc-image-builder binary and build from ibcli

mvo5:
This commit removes bib/cmd/bootc-image-builder and all pieces that it pulled in and replaces it with the new image-builder-cli multicall binary.

It also removes one test which is a bit unfortunate but with the move to ibcli this needs to be re-added there anyway.


supakeen:
Not stale

PR #1166, opened 2025-12-11
README: minor fixes and improvements

achilleas-k:
SSIA.

Thanks to @Serisium for pointing some of these out.


bcl (approved):
Ack. But at some point this needs more details about the differences between bootc-installer and anaconda-iso and why you would use one instead of the other.

PR #1189, opened 2026-01-29
Support root-less containers with --in-vm

alexlarsson:
This updates to the latest image-builder-cli, adding support for rootless container use
(https://github.com/osbuild/image-builder-cli/pull/445).

It also updates the use of related APIs to pass runInVm options where needed.

With this, I was able to run a rootless bc-i-b conversion. There is one problem, which is that if you mount ~/.local/share/containers/storage on the host to /var/lib/containers/storage in the contained, podman will complain with:

Error: database static dir "~/.local/share/containers/storage/libpod" does not match our static dir "/var/lib/containers/storage/libpod": database configuration mismatch

Additionally, if you pass the host "/var/lib/containers/storage" into the rootless container you will get read permission errors.

There are two workarounds for this. Either you can use e.g. skopeo to copy the bootc container to a separate (non-root) container storage directory and mount that, or you can cover the "db.sql" file in the storage directory to make podman not print the error.

Neither of these are super clean, and we should try to figure out a better solution, but for now I was at least able to run a complete image build using:

```
$ touch /tmp/foo
$ podman run --rm --security-opt label=type:unconfined_t -ti --privileged \
  --network=none -v $PWD/output:/output \
  -v ~/.local/share/containers/storage:/var/lib/containers/storage \
  -v /tmp/foo:/var/lib/containers/storage/db.sql \
  localhost/bootc-image-builder  --in-vm \
  --rootfs ext4 --type raw \
  quay.io/fedora/fedora-bootc:43
```

### DIFF
--- a/bib/cmd/bootc-image-builder/main.go
+++ b/bib/cmd/bootc-image-builder/main.go
@@ -268,7 +268,7 @@ func cmdBuild(cmd *cobra.Command, args []string) error {
 	runInVM, _ := cmd.Flags().GetBool("in-vm")
 
 	logrus.Debug("Validating environment")
-	if err := setup.Validate(targetArch); err != nil {
+	if err := setup.Validate(targetArch, runInVM); err != nil {
 		return fmt.Errorf("cannot validate the setup: %w", err)
 	}
 	logrus.Debug("Ensuring environment setup")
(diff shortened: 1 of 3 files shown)


bcl (approved):
Looks good!

Issue #824, opened 2025-02-06
Remove hardcoded kernel args


ProbstDJakob:
Is there a reason to keep the hardcoded kernel args?

https://github.com/osbuild/bootc-image-builder/blob/b35eaa8a3a915c58b70fb4572d6ee42de85a69c1/bib/cmd/bootc-image-builder/image.go#L335-L341

The `rw`, at least for me, is not the problem, but the `console=*` args prevent me from using `plymouth`. As soon as I remove those args on boot via the grub editor plymouth shows up.

cgwalters:
Yes agree we should drop them, though in practice we probably should try to move some of it to the base image, as it will mean that default VMs lose consoles.

LukeShortCloud:
We're running into this same exact problem with Playtron GameOS. For now, we're forking bootc-image-builder and removing the `console` kernel arguments.

https://github.com/playtron-os/bootc-image-builder/commit/183f0663cc072998bbdfb3694fc128b3d2698a76

ProbstDJakob:
I personally think that changing default values is a breaking change. If you or others see that the same way the following could be a possible way to keep the "damage" low (could also be applied if you do not think it is a breaking change):

1. Introduce an option like `--no-default-kargs` to bib and/or `{"customizations": {"kernel": {"no-default-kargs": true}}}` to blueprint
2. Start versioning bib (could also be before 1.)
3. When the time is right, release a breaking change both for bib and for `osbuild/images` which removes the option and the default values

But there are two problems with this approach:
1. I'm unsure if or when bib will get tagged releases (though there is #111)
2. I do not know what `osbuild/images` sees as a version bump indicating a breaking change. In the rust world version bumps below 1.0.0 are considered breaking if the minor version changes, but I think that is not the case here. That means 1.0.0 would be the next breaking change and I do not know if `osbuild/images` aims to release that version in the foreseeable future.

supakeen:
I've opened PR's for `image-builder` and `bootc-image-builder` that introduce the `--bootc-no-default-kernel-args` and `--no-default-kernel-args` command line options respectively.

For now no warnings are printed that this will be the future default; but this should allow users currently having problems with the default args to start omitting them from their builds.

@ProbstDJakob let me know if the above approach suits you, if so we can probably close this issue once the PRs merge?

PR #1194, opened 2026-02-25
cmd: bump images to 0.246.0 and wire up `--no-default-kernel-args`

supakeen:
In https://github.com/osbuild/bootc-image-builder/issues/824 it was pretty clear that our default kernel arguments (sometimes) get in the way of use cases.

After some discussion we likely want to, in the future, not ship any default kernel arguments at all. Since this is a breaking change we've put it behind a command-line flag for now; allowing those users that know what they want to turn off the default args.

In the future we might start a transition period to warn people that we'll stop injecting default kernel arguments; and that they can verify their containers by passing the `--bootc-no-default-kargs` flag and the options they have: bootc dropins for managed state kernel args, or blueprints for unmanaged state.

After that we'll likely drop the flag again *once* the default becomes "no kernel args".

---

Since this PR bumps images to `0.246.0` I've adjusted the inspection for the build container to do basically nothing (as we don't need to know anything).

### DIFF
--- a/plans/integration.fmf
+++ b/plans/integration.fmf
@@ -7,6 +7,7 @@ provision:
       is-supported: true
     disk:
       - size: '>= 120 GB'
+    memory: ">= 8 GB"
 prepare:
   how: install
   package:


supakeen:
The testing farm tests seem to be running out of memory; the compiler is getting killed, seems like we'll have to ask testing farm about it and ignore them in this PR.

```
+ go build -tags 'containers_image_openpgp exclude_graphdriver_btrfs exclude_graphdriver_devicemapper' -o ../bin/bootc-image-builder ./cmd/bootc-image-builder
github.com/aws/aws-sdk-go-v2/service/ec2: /usr/lib/golang/pkg/tool/linux_amd64/compile: signal: killed
Error: building at STEP "RUN ./build.sh": while running runtime: exit status 1
```

I've restarted the ISO test cases as they hit a timeout.

bcl (approved):
Looks good!

supakeen:
> Perhaps this could help: https://github.com/osbuild/bootc-image-builder/pull/1199
> 
> 
> 
> Testing farm tests I have no idea what this even is... :-)

Thanks, there's a commit in this PR that fixes testing farm so after I rebase on yours this should go green too!

PR #1199, opened 2026-03-05
test: bump ISO tests timeout

lzap:
We do the same for images, there are ideas how to make this faster but not right now.


bcl (approved):
I suppose :/

lzap:
I also see tests in ibcli (which are copy of this) so I guess I shall do the same there?

